PR #167, opened 2023-08-31
Make setup async to get rid of setup()

olsaarik:
This PR changes the `Bootstrap` API to make `recv` calls asynchronous (they return a `std::future<void>`) and implements this change in `TcpBootstrap` by moving receives to background threads. For each pair of source rank and tag there is a thread that performs the receives in the order they were issued.

`recv` is marked as `[[nodiscard]]` to avoid ignoring the asyncness of the function. However, I didn't see a warning from that in my build and I'm not sure why.

The `NonblockingFuture` class is now gone in favor of `std::future` everywhere. `BaseSemaphore` has a `std::shared_future` because there are multiple calls to `get` in its current form and `std::future` turns invalid after the first call to `get`. It might make sense to redesign the semaphores API to return semaphores in a `std::future`, so that after initialization the `.get()` indirection is avoided.

Alternatively, we could make everything `std::shared_future` to make the APIs overall easier to use, although that might feel a bit prescriptive. `std::future` can be a bit surprising especially on the Python side where calling `.get()` multiple times doesn't feel like it should be an error.

As an unrelated change this PR renames `getRank` to `rank` and `getNRanks` to `size` to conform to C++ and MPI naming.

Finally, instead of fixing the existing Python examples I moved the bootstrap example into a Jupyter notebook in the docs/ directory. The other one didn't seem like it's worth keeping anymore since we have the Python tests now.

### DIFF
--- a/python/test/_cpp/proxy_test.cpp
+++ b/python/test/_cpp/proxy_test.cpp
@@ -39,7 +39,7 @@ class MyProxyService {
         semaphores_(semaphores),
         proxy_([&](mscclpp::ProxyTrigger triggerRaw) { return handleTrigger(triggerRaw); }, [&]() { bindThread(); }) {
     int cudaDevice;
-    cudaGetDevice(&cudaDevice);
+    MSCCLPP_CUDATHROW(cudaGetDevice(&cudaDevice));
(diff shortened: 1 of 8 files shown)


chhwang:
Implemented by another PR. Thank you!

PR #229, opened 2023-11-29
Fix performance downgrade issue & update doc

Binyang2014:
For push function, we only need to make sure the instruction `st.global`  will be executed after the while loop. Since there is a Write-After-Read  hazard for `trigger.fst` (Check `this->triggers[curFifoHead % size].fst != 0` first then write value to `triggers[curFifoHead % size]`), we can expect the compiler and hardware can handle this situation correctly. Remove the `release.sys` there.

BTW, `st.global.release.sys.v2.u64` will cause perf regression issue. Previous we use `st.global.release.cta.v2.u64`, but seems not necessary.


include/mscclpp/fifo_device.hpp
```diff
@@ -64,9 +64,10 @@ struct FifoDeviceHandle {
 
     ProxyTrigger* triggerPtr = &(this->triggers[curFifoHead % size]);
 
-    // store with memory order release so that the while loop does not go pass this.
+    // There is a Write-After-Read hazard for the triggerPtr->fst. So the st instruction will not be executed
+    // before the loop.
 #if defined(MSCCLPP_DEVICE_CUDA)
-    asm volatile("st.global.release.sys.v2.u64 [%0], {%1,%2};" ::"l"(triggerPtr), "l"(trigger.fst), "l"(trigger.snd));
+    asm volatile("st.volatile.global.v2.u64 [%0], {%1,%2};" ::"l"(triggerPtr), "l"(trigger.fst), "l"(trigger.snd));
```
saeedmaleki, line 70: I think this is dangerous since the polling is with `memoryOrderRelaxed`. So, it is possible that the ptx write happens before the poll is done. So, I suggest we change the memory order for polling to release. It should not matter much given that we should rarely get into like 60 when fifo size is large enough.

saeedmaleki (changes requested):

saeedmaleki:
> For push function, we only need to make sure the instruction `st.global` will be executed after the while loop. Since there is a Write-After-Read hazard for `trigger.fst` (Check `this->triggers[curFifoHead % size].fst != 0` first then write value to `triggers[curFifoHead % size]`), we can expect the compiler and hardware can handle this situation correctly. Remove the `release.sys` there.
> 
> BTW, `st.global.release.sys.v2.u64` will cause perf regression issue. Previous we use `st.global.release.cta.v2.u64`, but seems not necessary.

Can you reason about why the compiler/hardware can handle this correctly?

Binyang2014:
> 
> Can you reason about why the compiler/hardware can handle this correctly?
I think the code has Write-After-Read hazard. If the compiler or hardware try to reorder the instruction, it will cause incorrect result.

The code can be simplified as this:
```cpp
volatile int *addr;
while(true) {
  if (atomic_load(addr, std::memory_order_relaxed) == 0) {
    break;
  }
}
atomic_store(addr, 1, std::memory_order_relaxed);

```
For `addr`, it has write-after-read hazard. According to https://en.cppreference.com/w/cpp/language/as_if
`Accesses (reads and writes) to volatile objects occur strictly according to the semantics of the expressions in which they occur. In particular, they are not reordered with respect to other volatile accesses on the same thread.`
, for a single thread, compiler should not reorder the store before load, which will change the outcome of the program.
If we are talking about the different threads, since they will have different `curFifoHead`, then the polling addrs are different, that's should not be a problem.

include/mscclpp/fifo_device.hpp
Binyang2014: After synced offline, I think we agree using `atomic_store` with `memory_order_relax` is safe in this case. Feel free to add comments if have other concerns

saeedmaleki (approved):

liangyuRain:
Hi all, removing the `.release` qualifier may also have the similar problem as in #497. PTX documentation [link](https://docs.nvidia.com/cuda/parallel-thread-execution/#volatile-operation) says

> Volatile instructions are not re-ordered around other volatile instructions, but the memory operations of those instructions may be re-ordered around each other.

So from the perspective of another cuda thread or host, the memory order may not be as expected? Especially, how does fifo guarantee the memory operations before pushing the triggers are actually ordered before `fifo.push` from the perspective of host?

PR #237, opened 2023-12-19
Add unit test with high-res timer for DeviceSyncer

BKP:
The unit test has fixture executions for MI210 hardware. On other GCN architectures, tests may stall out in the polling loop with the DeviceSyncer due to block not being scheduled on SM.

### DIFF
--- a/test/unit/sync_tests.hpp
+++ b/test/unit/sync_tests.hpp
@@ -1,0 +1,128 @@
+/******************************************************************************
+ * Copyright (c) 2024 Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to
+ * deal in the Software without restriction, including without limitation the
+ * rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+ * sell copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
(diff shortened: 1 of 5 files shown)


test/unit/sync_tests.hpp
```diff
@@ -0,0 +1,130 @@
...
+ protected:
+  uint64_t memrealtime_freq_mhz() {
+    cudaDeviceProp deviceProp{};
+    MSCCLPP_CUDATHROW(cudaGetDeviceProperties(&deviceProp, 0));
+#if defined(__HIP_PLATFORM_AMD__) && (__HIP_PLATFORM_AMD__ == 1)
+    switch (deviceProp.gcnArch) {
+      case 900: return 27;
+      case 906: return 25;
+      case 908: return 25;
+      case 910: return 25;
```
chhwang, line 106: @BKP Thank you for contribution! Could you advise me where these numbers are from? 25MHz looks too small as GPU clock frequency.

BKP: The frequency comes from a constant frequency timer.

Information from the instruction can be found in the ISA manual for either MI100 or MI200. (The "27" for GFX900 is not well documented as far as I know, but the "25" from the newer architectures is clearly specified in newer ISA manuals.)

https://gpuopen.com/amd-isa-documentation/

https://www.amd.com/content/dam/amd/en/documents/instinct-tech-docs/instruction-set-architectures/instinct-mi200-cdna2-instruction-set-architecture.pdf

See "s_memrealtime".

PR #250, opened 2024-01-22
NVLS support.

saeedmaleki:


python/mscclpp_benchmark/mscclpp_op.py
```diff
@@ -418,3 +418,82 @@ def auto_tune(self):
             for block_size in block_size_to_try:
                 self.set_params(nblocks, block_size)
                 yield nblocks, block_size
+
+
+class MscclppAllReduce6:
```
saeedmaleki, line 423: @aashaka this is where one can setup a nvls connection and use it for allreduce.

python/mscclpp_benchmark/allreduce.cu
```diff
@@ -775,3 +776,67 @@ extern "C" __global__ void __launch_bounds__(1024, 1)
...
+  asm volatile("multimem.st.global.v4.f32 [%0], {%1,%2,%3,%4};" ::"l"(ptr), "r"(val.x), "r"(val.y), "r"(val.z), \
+               "r"(val.w)                                                                                       \
+               : "memory");
+// specific PTX for fp16 reduction. bf16 would be multimem.ld_reduce.global.add.v4.bf16x2 etc
+#define MULTIMEM_LD(val, ptr)                                     \
+  asm("multimem.ld_reduce.global.add.v4.f32 {%0,%1,%2,%3}, [%4];" \
+      : "=r"(val.x), "=r"(val.y), "=r"(val.z), "=r"(val.w)        \
+      : "l"(ptr)                                                  \
+      : "memory");
+
```
Binyang2014, line 796: Do we need to move this to mscclpp source, and provide version for different types

src/communicator.cc
```diff
@@ -105,6 +105,45 @@ MSCCLPP_API_CPP NonblockingFuture<std::shared_ptr<Connection>> Communicator::con
...
+      if (nvlsRank != rank) bootstrap->send(serialized, nvlsRank, 0);
+    }
+  } else {
+    std::vector<char> data;
+    bootstrap->recv(data, rootRank, 0);
+    conn = std::make_shared<NvlsConnection>(data);
+  }
+
+  // Now let's synchronize all ranks
+  bootstrap->groupBarrier(allRanks);
```
chhwang, line 138: Any reason why we newly define `groupBarrier()` instead of reusing existing `barrier()`?

include/mscclpp/core.hpp
```diff
@@ -445,26 +451,70 @@ class Connection {
...
+  std::shared_ptr<Impl> pimpl_;
+};
+
 /// Used to configure an endpoint.
 struct EndpointConfig {
   static const int DefaultMaxCqSize = 1024;
   static const int DefaultMaxCqPollNum = 1;
   static const int DefaultMaxSendWr = 8192;
   static const int DefaultMaxWrPerSend = 64;
+  static const int DefaultNvlsBufferSize = (1 << 29);
```
chhwang, line 496: What is the intuition behind this number?

include/mscclpp/gpu_utils.hpp
```diff
@@ -118,6 +171,25 @@ Memory safeAlloc(size_t nelem) {
   return Memory(ptr, Deleter());
 }
 
+template <class T, T*(alloc)(size_t, size_t), class Deleter, class Memory>
+Memory safeAlloc(size_t nelem, size_t gran) {
+  if ((nelem * sizeof(T)) % gran) {
+    throw Error("The request allocation size is not divisible by the required granularity:" +
```
chhwang, line 177: Granularity or alignment? I think `cuMemAddressReserve` doesn't require the size to be multiple of the alignment.

Binyang2014: According to: https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__VA.html#group__CUDA__VA_1g899d69a862bba36449789c64b430dc7c, The size of this allocation must be a multiple of the the value given via [cuMemGetAllocationGranularity](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__VA.html#group__CUDA__VA_1g30ee906c2cf66a0347b3dfec3d7eb31a) with the [CU_MEM_ALLOC_GRANULARITY_MINIMUM](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__TYPES.html#group__CUDA__TYPES_1gg3a202e4d32ae296db1af7efe75ce365dc74872d07341bb1ac24ccc4a1c9c2f56) flag

src/communicator.cc
Binyang2014: Mayme allow a part of ranks using nvls?

python/mscclpp_benchmark/allreduce_bench.py
```diff
@@ -147,12 +154,16 @@ def run_benchmark(
     if MPI.COMM_WORLD.size // N_GPUS_PER_NODE == 1:
         if memory.nbytes < 2**20:
             mscclpp_call = MscclppAllReduce2(mscclpp_group, memory, memory_out)
-        elif memory.nbytes < 2**29:
+        elif memory.nbytes < 2**21 if is_nvls_supported() else memory.nbytes < 2**29:
```
saeedmaleki, line 157: This isn't a good way to handle this. We should be able to test all algorithms and pick the best. This looks very manual.

include/mscclpp/gpu_utils.hpp
saeedmaleki: not entirely sure the difference between alignment and granularity.

include/mscclpp/core.hpp
saeedmaleki: this is the recommended size. I would suggest we get rid of this and only get the recommended size from cuMemGetAllocationGranularity. Your thoughts?

src/communicator.cc
saeedmaleki: Exactly.

python/test/nvls_test.cu
```diff
@@ -0,0 +1,77 @@
+// Copyright (c) Microsoft Corporation.
```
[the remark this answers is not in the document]
saeedmaleki: did these disappear entirely?!

include/mscclpp/core.hpp
saeedmaleki: actually I take it back. I notice whatever we choose here, it will occupy that much memory on the GPUs. We might wanna allow user to allocate less than 512MB

src/communicator.cc
chhwang: But we don't have use cases. When do we need that?

python/mscclpp_benchmark/allreduce_bench.py
chhwang: I agree on this.

include/mscclpp/gpu_utils.hpp
chhwang: Alignment is about the starting address, e.g. range 0x08~0x0c granularity is less than 4 bytes but aligned to 8 bytes. But I now understand we need granularity here, thanks!

python/test/nvls_test.cu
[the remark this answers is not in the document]
Binyang2014: You mean the nvls test code? moved to `test/nvls_test.cu`

python/mscclpp_benchmark/allreduce_bench.py
Binyang2014: updated, add `find_best_algo` function. Due to some algos are designed for small message size, exclude these algos from the candidates for large message size

src/communicator.cc
chhwang: Let's just keep it for future usages.

python/mscclpp_benchmark/allreduce_bench.py
chhwang: Looks great!

chhwang (approved):

chhwang:
> @saeedmaleki please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

Hi @saeedmaleki, according to the instruction, could you please reply **"@microsoft-github-policy-service agree"** to the bot comment? Thank you!

PR #255, opened 2024-02-01
Allow MSCCL++ CommGroup to take PyTorch tensors in args

aashaka:
Obtain data_ptr and tensor_size accordingly for torch.Tensor

### DIFF
--- a/python/mscclpp/comm.py
+++ b/python/mscclpp/comm.py
@@ -22,6 +22,8 @@ from ._mscclpp import (
 import mpi4py
 import numpy as np
 
+from mscclpp.utils import is_torch_tensor
+
 
 class CommGroup:
     def __init__(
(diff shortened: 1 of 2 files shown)


python/mscclpp/comm.py
```diff
@@ -21,6 +21,7 @@
 )
 import mpi4py
 import numpy as np
+import torch
```
Binyang2014, line 24: Hmm... then our packages will depend on torch. Could we provide some utils such as `ptr_from_torch` to get the data_ptr?

chhwang: I agree, perhaps we can do something like this and use `TorchTensor` instead of `torch.Tensor`:
```python
try:
    import torch
    TorchTensor = torch.Tensor
except:
    class TorchTensor:
        pass
```

Binyang2014 (approved):
LGTM

chhwang (approved):

PR #264, opened 2024-02-08
Allow semaphores and memory to be registered separately in ProxyService

aashaka:
This is needed in use cases where SimpleProxyChannel does not suffice. For example, when a single semaphore is to be used for multiple tensors or when multiple semaphores should be associated with a tensor.


chhwang (commented):
I guess the method names are not the right ones, because e.g., `register_semaphore_with_proxy()` doesn't take any semaphores as arguments. Can you please rename them?

PR #290, opened 2024-04-18
Allow binding allocated memory to NVLS multicast pointer

roshandathathri:
And change NVLS multimem instructions to static functions


chhwang (approved):
LGTM

PR #293, opened 2024-04-23
Refactoring NVLS interfaces

chhwang:
Move NVLS details from the core to a separate interface


Binyang2014 (approved):
Separate NVLS related code to another class looks good to me. Agree with NVLS is not a kind of connection.

PR #318, opened 2024-06-21
Update allreduce_bench.py

angelica-moreira:
Replacing hardcoded network interface name for generic discovery strategy.


chhwang (approved):
LGTM

PR #321, opened 2024-07-02
Separate NPKit CPU timestamp access from different blocks for AMD platform

yzygitzh:
Reference: https://github.com/ROCm/rccl/pull/1229


Binyang2014 (approved):
LGTM

PR #322, opened 2024-07-03
AllReduce Kernel for Small Messages

caiomcbr:
Adding allreduce kernel code for message sizes smaller than 32 bytes, when the number of elements are smaller than the number of ranks.


chhwang (commented):
LGTM, need to fix scratch buffer sharing issues

PR #326, opened 2024-07-11
Multi-stream CUDA IPC

chhwang:
Let different `CudaIpcConnection`s use different streams on AMD platforms to fully utilize xGMI


src/include/context.hpp
```diff
@@ -16,6 +16,7 @@ namespace mscclpp {
 struct Context::Impl {
   std::vector<std::shared_ptr<Connection>> connections_;
   std::unordered_map<Transport, std::unique_ptr<IbCtx>> ibContexts_;
+  std::vector<std::shared_ptr<CudaStreamWithFlags>> ipcStreams_;
   std::shared_ptr<CudaStreamWithFlags> ipcStream_;
```
chhwang: We need to remove `ipcStream_`.

src/context.cc
```diff
@@ -43,7 +43,14 @@ MSCCLPP_API_CPP std::shared_ptr<Connection> Context::connect(Endpoint localEndpo
...
       throw mscclpp::Error("Local transport is CudaIpc but remote is not", ErrorCode::InvalidUsage);
     }
-    conn = std::make_shared<CudaIpcConnection>(localEndpoint, remoteEndpoint, pimpl_->ipcStream_);
+#if defined(__HIP_PLATFORM_AMD__)
+    pimpl_->ipcStreams_.emplace_back(std::make_shared<CudaStreamWithFlags>(cudaStreamNonBlocking));
+#else
+    if (pimpl_->ipcStreams_.empty()) {
+      pimpl_->ipcStreams_.emplace_back(std::make_shared<CudaStreamWithFlags>(cudaStreamNonBlocking));
+    }
+#endif
```
chhwang: The streams should be lazy-created when they are actually used: [example](https://github.com/microsoft/mscclpp/blob/main/src/connection.cc#L99). So we should not set the flag here; just do `std::make_shared<CudaStreamWithFlags>()`.

src/include/context.hpp
Binyang2014: Fixed

src/context.cc
Binyang2014: Fixed

caiomcbr (approved):
LGTM

PR #341, opened 2024-08-23
Fix bug for construct sempaphore

Binyang2014:
Current semaphore construction requires two-way communication, e.g., to construct a semaphore signaling from rank 0 to rank 1, both rank 0 and rank 1 need to send a message to each other. This PR fixes an executor bug that fails to conduct two-way communication for constructing such one-way semaphores, and instead hangs during the semaphore construction. In the future, we may need to change the implementation to construct semaphore via one-way communication.


src/include/execution_plan.hpp
```diff
@@ -33,6 +33,13 @@ struct hash<mscclpp::ChannelKey> {
            std::hash<int>()(static_cast<int>(key.dstBufferType)) ^ std::hash<int>()(static_cast<int>(key.channelType));
   }
 };
+
+template <>
+struct hash<std::pair<int, mscclpp::ChannelType>> {
+  std::size_t operator()(const std::pair<int, mscclpp::ChannelType>& key) const {
+    return std::hash<int>()(key.first) ^ std::hash<int>()(static_cast<int>(key.second));
```
chhwang, line 40: @Binyang2014 This looks unsafe as `(1, ChannelType::PROXY)` pair will have the same hash value as `(2, ChannelType::SM)`.

chhwang (changes requested):

src/include/execution_plan.hpp
Binyang2014: Thanks, fixed

chhwang (approved):
LGTM

PR #371, opened 2024-10-22
Add separate option for double scratch buffer

yzygitzh:
This commit separates two things: whether uses packet and whether uses double buffer.

### DIFF
--- a/src/executor/executor.cc
+++ b/src/executor/executor.cc
@@ -66,6 +66,7 @@ struct ExecutionContext {
   size_t scratchBufferSize;
   std::shared_ptr<char> deviceExecutionPlansBuffer;
   int nthreadsPerBlock;
+  bool isUsingDoubleScratchBuffer;
 };
 
 struct Executor::Impl {
@@ -106,6 +107,7 @@ struct Executor::Impl {
     context.scratchBufferSize = scratchBufferSize;
     context.proxyService = std::make_shared<ProxyService>();
     context.nthreadsPerBlock = plan.impl_->getNThreadsPerBlock();
+    context.isUsingDoubleScratchBuffer = plan.impl_->getIsUsingDoubleScratchBuffer();
     this->setupConnections(context, rank, plan);
     this->setupRegisteredMemories(context, sendbuff, recvbuff, sendBufferSize, recvBufferSize, rank, plan);
     this->setupChannels(context, sendbuff, recvbuff, sendBufferSize, recvBufferSize, rank, plan);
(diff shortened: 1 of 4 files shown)


src/executor/executor.cc
```diff
@@ -343,17 +354,18 @@ Executor::Executor(std::shared_ptr<Communicator> comm) : impl_(std::make_unique<
...
   size_t offsetIn = (char*)sendbuff - (char*)sendBasePtr;
   size_t offsetOut = (char*)recvbuff - (char*)recvBasePtr;
-
   ExecutionContext context =
       this->impl_->setupExecutionContext(rank, (void*)sendBasePtr, (void*)recvBasePtr, sendBuffSize, recvBuffSize,
-                                         offsetIn, offsetOut, sendBytes, recvBytes, plan);
-  this->impl_->launchKernel(context, rank, sendbuff, recvbuff, dataType, stream, packetType);
+                                         offsetIn, offsetOut, sendBytes, recvBytes, plan, flag);
+  this->impl_->launchKernel(context, rank, sendbuff, recvbuff, dataType, stream, packetType, flag);
+  flag++;
```
chhwang, line 368: What happens if the communication group changes during runtime? Will `flag` be still correct?

chhwang (approved):

src/executor/executor.cc
yzygitzh: As long as they run in the same stream, it should be OK.

src/executor/executor.cc
```diff
@@ -82,11 +82,13 @@ struct Executor::Impl {
...
+                                         int flag) {
     ExecutionContextKey key = {sendbuff, recvbuff, sendBufferSize, recvBufferSize, plan.impl_->name};
     if (this->contexts.find(key) != this->contexts.end()) {
       plan.impl_->operationsReset();
-      plan.impl_->lightLoadExecutionPlan(inputMessageSize, outputMessageSize, contsSrcOffset, constDstOffset);
+      plan.impl_->lightLoadExecutionPlan(inputMessageSize, outputMessageSize, contsSrcOffset, constDstOffset, rank,
+                                         sendBufferSize, recvBufferSize, flag);
       this->setupDeviceExecutionPlan(this->contexts[key], rank, plan);
       this->contexts[key].deviceExecutionPlansBuffer =
           allocExtSharedCuda<char>(this->contexts[key].deviceExecutionPlans.size() * sizeof(DeviceExecutionPlan));
```
Binyang2014, line 94: If find the key, seems it will overwrite the previous plan. It may cause problem when using cuda graph. Only last plan will be copied to GPU memory. Maybe we need to distinguish the plan with `inputMessageSize, outputMessageSize, flag`. And let kernel load the plan with correct device_ptr.

chhwang:
@Binyang2014 Any updates in this PR?

Binyang2014:
> @Binyang2014 Any updates in this PR?

I think there not blocker for this PR. Just need to resolve the conflict. Let me check with Ziyue

src/executor/executor.cc
```diff
@@ -152,7 +153,8 @@ struct Executor::Impl {
         return this->contexts[key];
       }
       plan.impl_->operationsReset();
-      plan.impl_->lightLoadExecutionPlan(inputMessageSize, outputMessageSize, constSrcOffset, constDstOffset);
+      plan.impl_->lightLoadExecutionPlan(inputMessageSize, outputMessageSize, constSrcOffset, constDstOffset, rank,
```
Binyang2014, line 156: Seems we need to add flag % 2 to `DeviceExecutionPlanKey`, otherwise we will not recalculate the scratch buffer offset, even flag is changed

PR #378, opened 2024-11-06
Add kernel-based verification for executor_test

yzygitzh:
Add kernels to fill and test data for correctness test in executor_test.py.

### DIFF
--- a/python/test/executor_test.py
+++ b/python/test/executor_test.py
@@ -10,47 +10,12 @@ from mscclpp import (
     npkit,
 )
 import mscclpp.comm as mscclpp_comm
+from mscclpp.utils import KernelBuilder, pack
 import os
+import struct
 
 import cupy as cp
 from mpi4py import MPI
-
-
-def bench_time(niters: int, ngraphIters: int, func):
-    # capture cuda graph for niters of the kernel launch
-    stream = cp.cuda.Stream(non_blocking=True)
-    with stream:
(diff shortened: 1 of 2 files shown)


python/test/executor_test.py
```diff
@@ -33,12 +48,58 @@ def bench_time(niters: int, ngraphIters: int, func):
...
+        file="executor_test_verifier.cu", kernel_name=test_data_kernel_name, file_dir=file_dir
+    ).get_compiled_kernel()
+    nblocks = 64
+    nthreads = 1024
+
+    stream = cp.cuda.Stream(non_blocking=True)
+    with stream:
+        stream.begin_capture()
+        for i in range(n_iters):
+            fill_data_params = pack(input_buf) + struct.pack("Q", input_buf.nbytes // type_size) + pack(rank, i)
```
Binyang2014, line 93: A bit confused, why we don't use function from cupy to init data and verify the correctness but write the kernel. Because of the limitation of cuda graph?

yzygitzh: cupy calls need extra CUDA synchronization to know whether the results are correct at CPU side. I want to put the check inside the kernel to minimize such synchronization, to better make sure there is no barrier issue or data corruption issue in asynchronous kernel execution conditions. It’s actually more like running a real GPU application.

Binyang2014 (approved):

PR #379, opened 2024-11-07
Fix light load bug

Binyang2014:
Fix lightLoadExecutionPlan issue.
An execution context many have multi device execution plans. These plans share the channel connections which are constructed before.
A deviceExecutionPlanKey is introduced to identify these plans. We can get the current device execution plan key via:
`contexts.currentDevicePlan`

### DIFF
--- a/src/executor/executor.cc
+++ b/src/executor/executor.cc
@@ -22,21 +22,62 @@ struct ExecutionContextKey {
            recvBuffSize == other.recvBuffSize && plan == other.plan;
   }
 };
+
+struct ExecutionDevicePlanKey {
+  size_t inputMessageSize;
+  size_t outputMessageSize;
+  size_t contsSrcOffset;
+  size_t constDstOffset;
+
+  bool operator==(const ExecutionDevicePlanKey& other) const {
+    return inputMessageSize == other.inputMessageSize && outputMessageSize == other.outputMessageSize &&
+           contsSrcOffset == other.contsSrcOffset && constDstOffset == other.constDstOffset;
+  }
+};
+
 }  // namespace mscclpp
 
 namespace std {
+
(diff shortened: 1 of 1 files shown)


src/executor/executor.cc
```diff
@@ -22,21 +22,62 @@ struct ExecutionContextKey {
            recvBuffSize == other.recvBuffSize && plan == other.plan;
   }
 };
+
+struct DeviceExecutionPlanKey {
+  size_t inputMessageSize;
+  size_t outputMessageSize;
+  size_t contsSrcOffset;
```
yzygitzh, line 29: `contsSrcOffset` -> `constSrcOffset`?

src/executor/executor.cc
```diff
@@ -84,15 +126,24 @@ struct Executor::Impl {
                                          size_t outputMessageSize, size_t contsSrcOffset, size_t constDstOffset,
                                          size_t sendBufferSize, size_t recvBufferSize, const ExecutionPlan& plan) {
     ExecutionContextKey key = {sendbuff, recvbuff, sendBufferSize, recvBufferSize, plan.impl_->name};
+    DeviceExecutionPlanKey devicePlanKey = {inputMessageSize, outputMessageSize, contsSrcOffset, constDstOffset};
```
yzygitzh, line 129: Same as above

yzygitzh (changes requested):

src/executor/executor.cc
Binyang2014: Fixed

src/executor/executor.cc
Binyang2014: Fixed

yzygitzh (approved):

PR #380, opened 2024-11-08
Fixing Bug Const Offset in Execution Plan

caiomcbr:
The offset was not differentiating between the buffer types, causing the offset to be incorrect when the buffer type was not `SCRATCH`.


src/executor/execution_plan.cc
```diff
@@ -292,6 +292,19 @@ void ExecutionPlan::Impl::setupChannels(const json& gpus) {
 }
 
 void ExecutionPlan::Impl::setupOperations(const json& gpus, size_t contsSrcOffset, size_t constDstOffset) {
+  auto getConstOffset = [&](BufferType type) -> size_t {
+    switch (type) {
+      case BufferType::INPUT:
+        return contsSrcOffset;
```
chhwang, line 298: Please fix the typo

chhwang (approved):

PR #383, opened 2024-11-13
Add cross threadblock barrier

Binyang2014:


src/executor/execution_plan.cc
```diff
@@ -17,6 +17,8 @@ std::vector<T> filter(const std::vector<T>& vec, Predicate pred) {
 
 auto getOpType = [](const std::string& str) {
   if (str == "nop") {
+    return mscclpp::OperationType::THREADBLOCK_BARRIER;
```
chhwang, line 20: `"nop"` returns `THREADBLOCK_BARRIER`?

chhwang (approved):

src/executor/execution_plan.cc
Binyang2014: Fixed. Return NOP instead of THREADBLOCK_BARRIER.

PR #390, opened 2024-11-22
Providing reduce-scatter test support

caiomcbr:


python/test/executor_test_verifier.cu
```diff
@@ -36,7 +36,7 @@ FILL_DATA(int32, int)
 
 #define TEST_DATA_ALL_GATHER(FuncNameType, DataType)                                                       \
   extern "C" __global__ void __launch_bounds__(1024, 1) test_data_all_gather_##FuncNameType(               \
-      DataType* result_buf, DataType* test_buf, size_t num_elems, int num_ranks, int seq) {                \
+      DataType* result_buf, DataType* test_buf, size_t num_elems, int num_ranks, int my_rank, int seq) {   \
```
Binyang2014, line 39: Will the compiler complain that `my_rank` is not used

Binyang2014 (approved):

python/test/executor_test_verifier.cu
caiomcbr: I included this parameter because in executor_test.py, the function is used generically. Therefore, I applied the same parameters to TEST_DATA_ALL_GATHER, TEST_DATA_ALL_REDUCE, and TEST_DATA_REDUCE_SCATTER. Although this parameter is only utilized in TEST_DATA_REDUCE_SCATTER. Do you think it would be better to call this function individually in executor_test.py?

PR #396, opened 2024-12-01
Select algo according to json config

Binyang2014:
The way to run nccl-test over mscclpp:
mpirun -np 8 --bind-to numa --allow-run-as-root -x LD_PRELOAD=$(pwd)/build/apps/nccl/libmscclpp_nccl.so -x NCCL_DEBUG=WARN  -x MSCCLPP_EXECUTION_PLAN_DIR=/execution-files /root/nccl-tests/build/all_reduce_perf   -b 1K -e 1G -f 2 -d half -G 20 -w 10 -n 20

### DIFF
--- a/apps/nccl/src/nccl.cu
+++ b/apps/nccl/src/nccl.cu
@@ -400,8 +400,8 @@ NCCL_API ncclResult_t ncclCommInitRank(ncclComm_t* comm, int nranks, ncclUniqueI
       setupRemoteMemories(commPtr->comm, rank, commPtr->scratchBuff.get(), SCRATCH_SIZE, mscclpp::Transport::CudaIpc);
   commPtr->executor = std::make_shared<mscclpp::Executor>(mscclppComm);
 
-  if (getenv("COMMUNICATION_COLLECTIVE_DIR")) {
-    std::string collectiveDir = getenv("COMMUNICATION_COLLECTIVE_DIR");
+  if (getenv("MSCCLPP_EXECUTION_PLAN_DIR")) {
+    std::string collectiveDir = getenv("MSCCLPP_EXECUTION_PLAN_DIR");
     for (const auto& entry : std::filesystem::directory_iterator(collectiveDir)) {
       if (entry.is_regular_file()) {
         std::string filename = entry.path().filename().string();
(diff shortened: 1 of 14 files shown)


src/include/execution_plan.hpp
```diff
@@ -63,6 +63,7 @@ struct ChannelInfo {
 struct ExecutionPlan::Impl {
  public:
   Impl(const std::string name, const std::string planPath);
```
chhwang, line 65: Looks like we should remove this line

apps/nccl/src/nccl.cu
```diff
@@ -183,6 +192,13 @@ static std::vector<mscclpp::SmChannel> setupSmChannels(ncclComm_t comm,
   return channels;
 }
 
+static std::pair<std::string, executionPlanInstance> loadExecutionPlan(const char* filename) {
```
chhwang, line 195: no need to make it `const char *`, better to keep it `const std::string&`

apps/nccl/src/nccl.cu
```diff
@@ -383,52 +399,16 @@ NCCL_API ncclResult_t ncclCommInitRank(ncclComm_t* comm, int nranks, ncclUniqueI
...
-  else
-    commPtr->allGatherLargeMessageSizeBoundary = 1 << 20;
-
-  if (commPtr->allReduceSmallMessageSizeBoundary > commPtr->allReduceLargeMessageSizeBoundary)
-    return ncclInvalidArgument;
-  if (commPtr->allGatherSmallMessageSizeBoundary > commPtr->allGatherLargeMessageSizeBoundary)
-    return ncclInvalidArgument;
+  if (getenv("MSCCLPP_EXECUTION_PLAN_DIR")) {
+    std::string collectiveDir = getenv("MSCCLPP_EXECUTION_PLAN_DIR");
+    for (const auto& entry : std::filesystem::directory_iterator(collectiveDir)) {
```
chhwang, line 404: Need an exception handler in case `collectiveDir` is not a valid directory

chhwang (approved):
LGTM

apps/nccl/src/nccl.cu
Binyang2014: Thanks, check with `std::filesystem::is_directory(collectiveDir)` first and return ncclInvalidArgument if dir is invalid

apps/nccl/src/nccl.cu
```diff
@@ -383,52 +399,19 @@ NCCL_API ncclResult_t ncclCommInitRank(ncclComm_t* comm, int nranks, ncclUniqueI
...
-  if (commPtr->allGatherSmallMessageSizeBoundary > commPtr->allGatherLargeMessageSizeBoundary)
-    return ncclInvalidArgument;
+  if (getenv("MSCCLPP_EXECUTION_PLAN_DIR")) {
+    std::string collectiveDir = getenv("MSCCLPP_EXECUTION_PLAN_DIR");
+    if (!std::filesystem::is_directory(collectiveDir)) {
+      return ncclInvalidArgument;
+    }
+    for (const auto& entry : std::filesystem::directory_iterator(collectiveDir)) {
+      if (entry.is_regular_file()) {
+        std::string filename = entry.path().filename().string();
```
caiomcbr, line 409: Are you using this variable?

Binyang2014: removed it. Thanks

PR #400, opened 2024-12-05
Revised ProxyChannel interfaces

chhwang:
* Renamed `ProxyChannel` -> `BaseProxyChannel` and `SimpleProxyChannel` -> `ProxyChannel`. It makes the interface more consistent by defining channels to be associated with a certain src/dst memory region: `ProxyChannel` as "sema + src/dst + fifo" and `SmChannel` as "sema + src/dst". BaseProxyChannel is not associated with any memory regions, as "sema + fifo".
* `ProxyChannelDeviceHandle` now inherits from `BaseProxyChannelDeviceHandle`, instead of having one as a member.

### DIFF
--- a/src/semaphore.cc
+++ b/src/semaphore.cc
@@ -94,7 +94,7 @@ MSCCLPP_API_CPP SmDevice2DeviceSemaphore::SmDevice2DeviceSemaphore(Communicator&
         setupInboundSemaphoreId(communicator, connection.get(), localInboundSemaphore_.get());
     isRemoteInboundSemaphoreIdSet_ = true;
   } else if (AllIBTransports.has(connection->transport())) {
-    // We don't need to really with any of the IB transports, since the values will be local
+    // Should we throw an error here?
     isRemoteInboundSemaphoreIdSet_ = false;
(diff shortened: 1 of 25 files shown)


Binyang2014 (approved):
LGTM

src/semaphore.cc
```diff
@@ -94,7 +94,7 @@ MSCCLPP_API_CPP SmDevice2DeviceSemaphore::SmDevice2DeviceSemaphore(Communicator&
         setupInboundSemaphoreId(communicator, connection.get(), localInboundSemaphore_.get());
     isRemoteInboundSemaphoreIdSet_ = true;
   } else if (AllIBTransports.has(connection->transport())) {
-    // We don't need to really with any of the IB transports, since the values will be local
+    // Should we throw an error here?
```
chhwang, line 97: @Binyang2014 How do you think about this?

Binyang2014: Since the class name is `SmDevice2DeviceSemaphore` I think it makes sense to throw error here. SM channel only supports the mmap way

PR #401, opened 2024-12-05
Setup pipeline for mscclpp over nccl

Binyang2014:
Setup pipeline for mscclpp over nccl
Run `all_reduce_perf` via nccl API


python/requirements_cuda12.txt
```diff
@@ -5,3 +5,4 @@ netifaces
 pytest
 numpy
 matplotlib
+ninja
```
chhwang, line 8: What is `ninja` used for here?

chhwang (approved):

python/requirements_cuda12.txt
Binyang2014: Pls check this output: https://msazure.visualstudio.com/One/_build/results?buildId=109627341&view=logs&j=a940a0bc-e16f-5378-628a-dcb452f8b6f6&t=4bccd649-850f-5156-886b-2dc4575d8cd7 seems related with python build tool

Binyang2014: Maybe due to base image upgrade from ubuntu 20.04 to ubuntu 22.04

chhwang: It looks like because `scikit_build_core` is upgraded and the later version do not support cmake.minimum-version anymore. Could you please fix our pyproject.toml?

chhwang: We should use cmake.version instead, then probably it will not try to use ninja.

Binyang2014: Sure, let me take a look

Binyang2014: Not working...  `pip install .` works in venv.

PR #403, opened 2024-12-07
Reduce memory usage for scratch buffer

Binyang2014:
In the executor, we allocate the scratch buffer based on `sendMemRange`. However, for certain execution plans, this allocation may be unsuitable, as the plan does not support messages of this size.

To avoid allocating to much data and cause OOM error, set scratch buffer size to `min(scratchBufferSize(maxMessageSizeSupportedForPlan), scratchBufferSize(sendMemRange))`


src/executor/execution_plan.cc
```diff
@@ -165,20 +165,35 @@ std::vector<BufferType> ExecutionPlan::Impl::getConnectedBufferTypes(int rank) c
...
+  size_t inputChunks = this->inputChunks.at(rank);
+  if (inputChunks != 0)
+    sizePerChunk = (inputSize + inputChunks - 1) / this->inputChunks.at(rank);
   else
-    throw mscclpp::Error("Output or Input chunks must be greater than 0", mscclpp::ErrorCode::ExecutorError);
+    throw mscclpp::Error("Input chunks must be greater than 0", mscclpp::ErrorCode::ExecutorError);
 
   if (this->isUsingPacket) {
-    return sizePerRank * this->scratchChunks.at(rank) * 2 /* data + flag*/ * 2 /*double buffer*/;
+    return sizePerChunk * this->scratchChunks.at(rank) * 2 /* data + flag*/ * 2 /*double buffer*/;
```
caiomcbr, line 178: In the JSON, we have a field to decide whether to use double buffering. Perhaps we could allocate that extra space only if that field is set to true. I'm not sure if it is worth economizing this space, but if there is an advantage, we could consider double buffering only when necessary.

caiomcbr (approved):
LGTM

src/executor/execution_plan.cc
```diff
@@ -165,20 +165,35 @@ std::vector<BufferType> ExecutionPlan::Impl::getConnectedBufferTypes(int rank) c
...
-    sizePerRank = outputSize / this->outputChunks.at(rank);
+
+size_t ExecutionPlan::Impl::getScratchBufferSize(int rank, size_t inputSize) const {
+  size_t sizePerChunk = 0;
+  size_t inputChunks = this->inputChunks.at(rank);
+  if (inputChunks != 0)
+    sizePerChunk = (inputSize + inputChunks - 1) / this->inputChunks.at(rank);
   else
-    throw mscclpp::Error("Output or Input chunks must be greater than 0", mscclpp::ErrorCode::ExecutorError);
+    throw mscclpp::Error("Input chunks must be greater than 0", mscclpp::ErrorCode::ExecutorError);
```
caiomcbr, line 175: It seems that there might be instances where the input size is zero. For example, in the broadcast, ranks other than the root will not have an input buffer, only an output buffer.

Binyang2014: Makes sense, let me modify the code

Binyang2014: Fixed

caiomcbr (approved):
LGTM

PR #404, opened 2024-12-09
Support nccl-test with mscclpp-nccl on H100 GPUs

seagater:
Add implementation for  ncclMemAlloc and ncclMemFree with mscclpp::allocSharedPhysicalCuda.  Leverage a global map to store associations between raw pointer and shared pointer.
Add placeholder APIs ncclCommRegister() and ncclCommDeregister().
Build pass with docker image ghcr.io/microsoft/mscclpp/mscclpp:base-cuda12.4 on H100 node.


Binyang2014:
Close as move to another branch

PR #407, opened 2024-12-10
Fix synchronization in allreduce8 kernel

dsidler:
Running kernel allreduce8 across 64 vGPUs (in CPX mode) revealed a synchronization bug. The PR addresses it by ensuring that signals are only issued after all threads in the block have issued their writes to guarantee correct ordering between data writes and signal writes.


dsidler:
@microsoft-github-policy-service agree [company="Advanced Micro Devices, Inc."]

dsidler:
@microsoft-github-policy-service agree company="Advanced Micro Devices, Inc."

chhwang (approved):

PR #411, opened 2024-12-16
Disable CuMemMap check for ROCm

Binyang2014:


src/registered_memory.cc
```diff
@@ -56,6 +56,9 @@ CUmemAllocationHandleType getNvlsCompatibleMemHandleType() {
 
 // Check if ptr is allocaed by cuMemMap
 bool isCuMemMapAllocated(void* ptr) {
```
dsidler, line 58: nit: produces compiler warning when compiling for rocm since `ptr` is not used

chhwang (approved):

PR #415, opened 2024-12-17
Flushing Proxy Channels at CPU side upon reaching the Inflight Request Limit

caiomcbr:


include/mscclpp/proxy_channel.hpp
```diff
@@ -11,6 +11,8 @@
 
 namespace mscclpp {
 
+constexpr int MAX_INFLIGHT_REQUEST = 500;
```
Binyang2014, line 14: What's the default CQ size. It's a constant number or can be configured by users?

caiomcbr: The default value is 1024, depending on the hardware the value can be bigger, for example the Mellanox HCA can support up to 2048. So the user can configure a different value, but must be small than the hardware specification.

chhwang: > The default value is 1024, depending on the hardware the value can be bigger, for example the Mellanox HCA can support up to 2048. So the user can configure a different value, but must be small than the hardware specification.

Can we make this value configurable?

caiomcbr: Currently we are using the default value every time that we create IB connection at the executor:
image

Maybe we could use some env variable to allow the user change it.

src/proxy_channel.cc
```diff
@@ -85,6 +93,7 @@ ProxyHandlerResult ProxyService::handleTrigger(ProxyTrigger triggerRaw) {
   if (trigger->fields.type & TriggerSync) {
```
yzygitzh, line 93: How about we just say `trigger->fields.type & TriggerSync || inflightRequests > MAX_INFLIGHT_REQUEST` here to merge the same block above

yzygitzh: Current way might have performance issue. When we call `flush`, it does busy polling until all inflight requests appear in CQ. This might waste some CPU cycles which instead could be used for submitting new requests. A better logic could be:
```
while (proxy_thread_running) {
  if (there_are_new_requests_to_submit) submit_requests();
  inflight += num_submitted_requests;
  if (there_are_new_requests_completed) complete_requests();
  inflight -= num_completed_requests;
}
```
But current `connection::flush` api doesn't do this. Perhaps we need to expose something like https://github.com/microsoft/mscclpp/blob/main/src/ib.cc#L288 to proxy thread.

caiomcbr: Interesting. I agree with you, when we flush the connection we need to wait all the requests, and maybe we can only clean some of the requests, we don't need to clean all of them. Although, I'm not sure where we can expose it, maybe the best place could be inside the ib connection class, but I'm not sure about this.

include/mscclpp/proxy_channel.hpp
chhwang: @caiomcbr `EndpointConfig` is a configurable struct that can be modified by users. We only set default values during construction, and users can change it. I meant if you can add a new inflight request field in `EndpointConfig`.

src/proxy_channel.cc
chhwang: > Current way might have performance issue. When we call `flush`, it does busy polling until all inflight requests appear in CQ. This might waste some CPU cycles which instead could be used for submitting new requests. A better logic could be:
> 
> ```
> while (proxy_thread_running) {
>   if (there_are_new_requests_to_submit) submit_requests();
>   inflight += num_submitted_requests;
>   if (there_are_new_requests_completed) complete_requests();
>   inflight -= num_completed_requests;
> }
> ```
> 
> But current `connection::flush` api doesn't do this. Perhaps we need to expose something like https://github.com/microsoft/mscclpp/blob/main/src/ib.cc#L288 to proxy thread.

This looks not straightforward to achieve, because other transports like CudaIpc cannot achieve this in an efficient way. It is possible, but will require event recording for every `put`, which can cause some GPU driver overhead. Anyway, this overhead is amortized as we can have a number of inflight requests, and flusing every IB requests one-by-one looks like an uncommon scenario. We can come back to this if we actually find a need of optimization here. cc @mahdiehghazim @SreevatsaAnantharamu

chhwang: @caiomcbr Could you please tackle @yzygitzh 's first comment? Additionally, you should set the default max inflight request into a very large number (like the max integer) if the connection type is CudaIpc.

caiomcbr: Okay, this make sense.

include/mscclpp/core.hpp
```diff
@@ -388,6 +388,11 @@ class Endpoint {
   /// @return The transport used.
   Transport transport();
 
+  /// Get the max inflight requests.
+  ///
+  /// @return max inflight requests.
+  int maxInflightRequests();
```
chhwang, line 394: The name is not very accurate

include/mscclpp/core.hpp
```diff
@@ -472,14 +484,13 @@ struct EndpointConfig {
...
-
-  /// Default constructor. Sets transport to Transport::Unknown.
-  EndpointConfig() : transport(Transport::Unknown) {}
+  int maxInflightRequests;
 
   /// Constructor that takes a transport and sets the other fields to their default values.
   ///
   /// @param transport The transport to use.
-  EndpointConfig(Transport transport) : transport(transport) {}
+  EndpointConfig(Transport transport = Transport::Unknown, int maxInflightRequests = -1)
```
chhwang, line 492: Please add all other members as a constructor variable as well

src/connection.cc
```diff
@@ -36,10 +36,13 @@ std::string Connection::getTransportName() {
          TransportNames[static_cast<int>(this->remoteTransport())];
 }
 
+int Connection::getMaxInflightRequest() { return maxInflightRequests; }
+
 // CudaIpcConnection
 
 CudaIpcConnection::CudaIpcConnection(Endpoint localEndpoint, Endpoint remoteEndpoint, cudaStream_t stream)
-    : stream_(stream) {
+    : Connection(localEndpoint.maxInflightRequests() != -1 ? localEndpoint.maxInflightRequests() : INT_MAX),
```
chhwang, line 44: Perhaps we can just use `-1` instead of `INT_MAX`, regarding -1 as infinite

chhwang (changes requested):

src/connection.cc
caiomcbr: Do you want the same for EthernerConnection?

chhwang: Yes

chhwang (approved):

PR #433, opened 2025-01-01
Update version number

Binyang2014:


chhwang (changes requested):
You need to update versions in the code as well: https://github.com/microsoft/mscclpp/blob/main/include/mscclpp/core.hpp#L7-L9

We may need an automation to prevent future mistakes.

Binyang2014:
> You need to update versions in the code as well: https://github.com/microsoft/mscclpp/blob/main/include/mscclpp/core.hpp#L7-L9
> 
> We may need an automation to prevent future mistakes.

Add VERSION file, we only need to modify this file to update the version, github action will do following things

chhwang (approved):
LGTM

PR #441, opened 2025-01-06
Adding Read Put Packet operation at Executor

caiomcbr:


yzygitzh:
This looks good to me, but I remember @Binyang2014 suggested a more general solution instead of hacking the code path.

src/include/execution_kernel.hpp
```diff
@@ -568,8 +598,13 @@ __global__ void executionKernel([[maybe_unused]] int rank /*for debug*/, T* inpu
                                op.inputChannelIndexes, op.outputOffsets, op.inputOffsets, op.nOutputs, op.nInputs,
                                op.size, false);
     } else if (op.type == OperationType::PUT_PACKET) {
-      handlePutPacket<PacketType>(scratchSize, smChannels, proxyChannels, op.outputChannelIndexes, op.outputOffsets,
-                                  op.inputOffsets, op.nOutputs, op.size, op.channelType, flag);
+      if (op.inputBufferType == BufferType::SCRATCH && op.channelType == ChannelType::SM) {
```
Binyang2014, line 601: Do we need to introduce new OP to handle this. Current implementation depends on channel type and buffer type, seems not a good choice.

src/include/execution_kernel.hpp
```diff
@@ -330,6 +330,36 @@ MSCCLPP_DEVICE_INLINE void handlePutPacket(size_t scratchSize, DeviceHandle<SmCh
...
+    if (ch_idx >= nDstChannels) {
+      return;
+    }
+    // For proxy channel, we assume src and dst are in packet format
+    uint32_t dstOffset = scratchBaseOffset + dstOffsets[ch_idx] * 2;
+    uint32_t srcOffset = scratchBaseOffset + srcOffsets[ch_idx] * 2;
+    proxyChannels[dstChannelIndexes[ch_idx]].put(dstOffset, srcOffset, size * 2);
+  }
+}
+
```
Binyang2014, line 362: For port channel, seems we need to read the data to somewhere then put it to other rank. Need read with flag to make sure the data is ready first

src/include/execution_kernel.hpp
```diff
@@ -334,6 +334,44 @@ MSCCLPP_DEVICE_INLINE void handlePutPacket(size_t scratchSize, DeviceHandle<Memo
...
+    if (ch_idx >= nDstChannels) {
+      return;
+    }
+
+    // Ensuring Data Is Ready
+    size_t nPackets = size * 2 / sizeof(PacketType);
+    for (size_t pkt_idx = threadIdx.x; pkt_idx < nPackets; pkt_idx += blockDim.x) {
+      PacketType* pkts = (PacketType*)((char*)scratch + scratchBaseOffset + srcOffsets[ch_idx] * 2);
+      PacketPayload<PacketType> data = pkts[pkt_idx].read(flag);
+    }
```
Binyang2014, line 366: Can you give an example for this case? It's for sending packet via intra node channel then trans via inter node channel? If so, we may need more threads for this, not just  `#tid < nDstChannels`

src/include/execution_kernel.hpp
```diff
@@ -334,6 +334,44 @@ MSCCLPP_DEVICE_INLINE void handlePutPacket(size_t scratchSize, DeviceHandle<Memo
...
+    int ch_idx = threadIdx.x;
+    if (ch_idx >= nDstChannels) {
+      return;
+    }
+
+    // Ensuring Data Is Ready
+    size_t nPackets = size * 2 / sizeof(PacketType);
+    for (size_t pkt_idx = threadIdx.x; pkt_idx < nPackets; pkt_idx += blockDim.x) {
+      PacketType* pkts = (PacketType*)((char*)scratch + scratchBaseOffset + srcOffsets[ch_idx] * 2);
+      PacketPayload<PacketType> data = pkts[pkt_idx].read(flag);
```
Binyang2014, line 365: We may need to set data to back to the scratch buffer, not sure if compiler will optimize the instruction if we only read the data

Binyang2014 (approved):
LGTM, can we add some examples for the new instruction?

PR #444, opened 2025-01-08
Fix Python binding of exceptions

chhwang:
* Fixed errors to be catchable from Python code
* Skip IB tests in Python unit tests when IB ports are down


python/mscclpp/error_py.cpp
```diff
@@ -9,7 +9,19 @@
 namespace nb = nanobind;
 using namespace mscclpp;
 
-void register_error(nb::module_& m) {
+#define REGISTER_EXCEPTION_TRANSLATOR(name_)                                                                         \
+  nb::register_exception_translator(                                                                                 \
+      [](const std::exception_ptr &p, void *payload) {                                                               \
+        try {                                                                                                        \
+          std::rethrow_exception(p);                                                                                 \
```
Binyang2014, line 16: Just curious, why need to rethrow here

Binyang2014 (approved):

python/mscclpp/error_py.cpp
chhwang: To my understanding, this is just for checking the type of exception `p` by try & catch. It is suggested in the nanobind documentation. https://nanobind.readthedocs.io/en/latest/exceptions.html#handling-custom-exceptions

PR #446, opened 2025-01-08
Add support for CPX mode on MI300X

nusislam:
This PR adds support for CPX mode on MI300X. It also adds an Allreduce kernel for large messages that provides better performance (than the default allreduce8 kernel) in CPX mode when the processes are in a single OAM.


nusislam:
@microsoft-github-policy-service agree company="AMD"

PR #451, opened 2025-01-10
Resolve cuMemMap error

Binyang2014:
* Updated `RegisteredMemory::Impl::Impl(const std::vector<char>& serialization)` to use both minimum and recommended granularities for memory address reservation and mapping. This will resolve the cuMemMap error

### DIFF
--- a/src/registered_memory.cc
+++ b/src/registered_memory.cc
@@ -24,22 +24,6 @@
   } while (false)
 
 namespace {
-size_t getMinGranularity(size_t size) {
-#if (CUDA_NVLS_SUPPORTED)
-  return mscclpp::detail::getMulticastGranularity(size, CU_MULTICAST_GRANULARITY_MINIMUM);
-#else
-  throw mscclpp::Error("Only support GPU with NVLS support", mscclpp::ErrorCode::InvalidUsage);
-#endif
-}
-
-size_t getRecommendedGranularity(size_t size) {
-#if (CUDA_NVLS_SUPPORTED)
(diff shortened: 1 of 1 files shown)


src/registered_memory.cc
```diff
@@ -241,10 +235,12 @@ RegisteredMemory::Impl::Impl(const std::vector<char>& serialization) {
     if (this->isCuMemMapAlloc) {
       CUmemGenericAllocationHandle handle;
       MSCCLPP_CUTHROW(cuMemImportFromShareableHandle(&handle, entry.shareableHandle, getNvlsCompatibleMemHandleType()));
-      size_t gran = getRecommendedGranularity();
-      MSCCLPP_CUTHROW(cuMemAddressReserve((CUdeviceptr*)&base, this->size, gran, 0, 0));
-      MSCCLPP_CUTHROW(cuMemMap((CUdeviceptr)base, this->size, 0, handle, 0));
-      detail::setReadWriteMemoryAccess(base, this->size);
+      size_t minGran = getMinGranularity(this->size);
+      size_t recommendedGran = getRecommendedGranularity(this->size);
```
chhwang, line 239: Seems we don't need new functions but can use `detail::getMulticastGranularity` directly

chhwang (approved):

src/registered_memory.cc
Binyang2014: Thanks, fixed

chhwang (approved):

PR #461, opened 2025-02-08
remove unnecessary sync

Binyang2014:
`nop` instruction is only for synchronization within the same threadblock. Cross threadblock synchronization is handled by `barrier` instruction. So insert `nop` only if the dependency is within the same threadblock.


Binyang2014:
Removed unnecessary `nop` instruction.  MSCCLPPLang test fail is expected

SreevatsaAnantharamu (approved):

PR #464, opened 2025-02-19
Exchange recvbuff at the beginning of the CUDA kernel call.

SreevatsaAnantharamu:
Goal here is to support zero-copy when buffers in only a few GPUs are changed.


SreevatsaAnantharamu:
Currently, I have just tested that the recvbuffs are exchanged correctly. Have implemented it in broadcast fallback where the memory channel created based on scratch buff is passed to the kernel. Need to still fully integrate it.

PR #466, opened 2025-02-21
Adjusting AllGather Collective in MSCCLLang

caiomcbr:


python/mscclpp/language/program.py
```diff
@@ -155,8 +155,7 @@ def lower(self):
         )
         for gpu in program.gpus:
             gpu.input_chunks = len(self.buffers[gpu.rank][Buffer.input]) * self.instances
-            if not self.collective.inplace:
-                gpu.output_chunks = len(self.buffers[gpu.rank][Buffer.output]) * self.instances
+            gpu.output_chunks = len(self.buffers[gpu.rank][Buffer.output]) * self.instances
```
Binyang2014, line 158: Maybe we can revert this change? For in-place algo, doesn't need to expose output chunks

Binyang2014 (approved):

python/mscclpp/language/program.py
caiomcbr: We have a problem for allgather. I agree with you that for the in-place algorithm, we shouldn't expose output chunks. The issue is that specifically for allgather, the in-place scenario only uses output chunks, causing some divergencies.
image
So maybe we can change this line to: "if self.collective.name != 'allgather' and 'not self.collective.inplace:'"

Binyang2014: Can we put this function into collective class? I mean call like this: `gpu.output_chunks = self.collective.get_output_chunk_count(len(self.buffers[gpu.rank][Buffer.output]), self.instances)`. There is a basic implementation at base class and allgather class has its' own implementation

Binyang2014 (approved):
LGTM

PR #469, opened 2025-02-25
nccl/rccl integration

seagater:
Use dlopen to load nccl/rccl apis from shared library to replace Fallback code for Allgather, Allreduce, Broadcast, ReduceScatter.

Add two related environment variables
-x MSCCLPP_ENABLE_SHARED_LIB=TRUE -x MSCCLPP_NCCL_LIB_PATH=path_to_libnccl.so/librccl.so


SreevatsaAnantharamu:
image

Additional symbols that need to be loaded from libnccl.so:
- ncclGetUniqueId
- ncclCommInitRank

nccl_ops_t->ncclGetUniqueId(

Here, before returning, you have to call nccl_ops_t->ncclCommInitRank and create a new real NCCL's communicator. Inside MSCCL++'s ncclComm_t, you can have a void * or ncclComm_t nccl_comm.

nccl_ops_t->ncclCommInitRank(&commPtr->nccl_comm,

seagater:
Add two related environment variables:
-x MSCCLPP_ENABLE_SHARED_LIB=TRUE -x MSCCLPP_NCCL_LIB_PATH=path_to_libnccl.so/librccl.so

Support dlopen for following nccl apis:
ncclCommInitRank
ncclGetUniqueId
ncclCommDestroy
ncclCommUserRank
ncclAllGather
ncclAllReduce
ncclBroadcast

Pass following tests 
**nccl-test:**
Allgather, Allreduce, Broadcast

**rccl-test:**
Allgather, Allreduce, Broadcast

apps/nccl/src/nccl.cu
```diff
@@ -94,6 +171,8 @@ struct ncclComm {
 
   uint32_t numScratchBuff;
   uint32_t buffFlag;
+
+  ncclComm_t nccl_comm;
```
Binyang2014, line 175: Change the variable name to camelCase style? Same as others

apps/nccl/src/nccl.cu
```diff
@@ -35,6 +37,81 @@
 
 #define NUM_CHANNELS_PER_CONNECTION 64
 
+typedef enum dlopen_err {
+  DLOPEN_SUCCESS = 0,
+  DLOPEN_ERROR = 1,
+} dlopen_err_t;
```
Binyang2014, line 43: Change to camelCase style

apps/nccl/src/nccl.cu
```diff
@@ -35,6 +37,81 @@
...
+  ncclResult_t (*AllGather)(const void* sendbuff, void* recvbuff, size_t sendcount, ncclDataType_t datatype,
+                            ncclComm_t comm, cudaStream_t stream);
+  ncclResult_t (*Broadcast)(const void* sendbuff, void* recvbuff, size_t count, ncclDataType_t datatype, int root,
+                            ncclComm_t comm, cudaStream_t stream);
+  ncclResult_t (*ReduceScatter)(const void* sendbuff, void* recvbuff, size_t recvcount, ncclDataType_t datatype,
+                                ncclRedOp_t op, ncclComm_t comm, cudaStream_t stream);
+} nccl_ops_t;
+
+nccl_ops_t nccl_ops;
+void* nccl_dl_handle = NULL;
```
Binyang2014, line 61: Ditto

apps/nccl/src/nccl.cu
```diff
@@ -396,6 +475,21 @@ NCCL_API ncclResult_t ncclGetUniqueId(ncclUniqueId* uniqueId) {
     WARN("uniqueId is nullptr");
     return ncclInvalidArgument;
   }
+
+  const bool mscclppEnableSharedLib = mscclpp::env()->enableSharedLib;
```
Binyang2014, line 479: better name for this? Seems too general for this env name

apps/nccl/src/nccl.cu
```diff
@@ -717,6 +845,10 @@ NCCL_API ncclResult_t ncclAllReduce(const void* sendbuff, void* recvbuff, size_t
     return ncclInvalidArgument;
   }
 
+  if (mscclppOpenSharedLib == true) {
```
Binyang2014, line 848: @SreevatsaAnantharamu @seagater, Do we you think we need to add flag for each communication collective? Such as `NCCL_FALLBACK="allreduce,allgather...."`, And user can use `NCCL_FALLBACK="all"` for all operations. If not specific, we will use ncclFallback for all unsupported collectives.

Binyang2014 (commented):
Thanks for this PR! A few comments

apps/nccl/src/nccl.cu
SreevatsaAnantharamu: This sounds like a good idea @Binyang2014. As you have mentioned above, users can just prescribe a comma-separated list of the collective for which we use the NCCL/RCCL fallback, or "all" to force.

These boolean variables can be part of our implementation of ncclComm.

apps/nccl/src/nccl.cu
SreevatsaAnantharamu: Also, I would recommend to have the prefix, mscclppNccl in these names.

apps/nccl/src/nccl.cu
SreevatsaAnantharamu: nccl_ops_t is very close to being an actual datatype name in applications. Can we call this as mscclppNcclOpsT or something like that?

apps/nccl/src/nccl.cu
```diff
@@ -396,6 +475,21 @@ NCCL_API ncclResult_t ncclGetUniqueId(ncclUniqueId* uniqueId) {
...
+  if (mscclppEnableSharedLib == true) {
+    int dlopen_status = nccl_dlopen_init();
+    if (dlopen_status == DLOPEN_SUCCESS) {
+      mscclppOpenSharedLib = true;
+    } else {
+      return ncclInternalError;
+    }
+  }
+  if (mscclppOpenSharedLib == true) {
+    nccl_ops.GetUniqueId(uniqueId);
```
SreevatsaAnantharamu, line 489: Do we need this? Wouldn't mscclpp's unqiue id be the same as nccl's unique id?

apps/nccl/src/nccl.cu
SreevatsaAnantharamu, line 175: Maybe we make ncclComm as void * just for clarity. This is a pointer to a different NCCL struct. It is not the encompassing ncclComm_t struct.

SreevatsaAnantharamu (commented):
I have left comments. Please take a look.

The most important is that we have to create both NCCL's communicator and our MSCCLPP communicator.

apps/nccl/src/nccl.cu
```diff
@@ -417,12 +511,29 @@ NCCL_API ncclResult_t ncclCommInitRank(ncclComm_t* comm, int nranks, ncclUniqueI
     WARN("nranks is %d, rank is %d", nranks, rank);
     return ncclInvalidArgument;
   }
+
+  const bool mscclppEnableSharedLib = mscclpp::env()->enableSharedLib;
+  if (mscclppEnableSharedLib == true && rank != 0) {
```
SreevatsaAnantharamu, line 516: It is not given that rank = 0 is always the root rank.

Maybe we avoid call to nccl_dlopen_init in ncclGetUniqueId.

Also, creating multiple communicators will lead to multiple dlopens which is not needed. 

We could make mscclppNcclOps as a global ptr initialized to NULL. Then, we can only do dlopen if it is NULL and calloc it.

seagater:
Update the environment variable names:
    -x MSCCLPP_ENABLE_NCCL_FALLBACK=TRUE 
    -x MSCCLPP_NCCL_LIB_PATH=$NCCL_LIB
    -x MSCCLPP_FORCE_NCCL_FALLBACK_OPERATION="allreduce,allgather,broadcast,reducescatter" or "all"

apps/nccl/src/nccl.cu
seagater: Resolved.

apps/nccl/src/nccl.cu
seagater: Resolved.

apps/nccl/src/nccl.cu
seagater: Resolved.

apps/nccl/src/nccl.cu
seagater: Resolved. Changed to const bool mscclppEnableNcclFallback = mscclpp::env()->enableNcclFallback;
The environment variable is changed to -x MSCCLPP_ENABLE_NCCL_FALLBACK=TRUE

apps/nccl/src/nccl.cu
seagater: Resolved. Now we can use 
-x MSCCLPP_FORCE_NCCL_FALLBACK_OPERATION="allreduce,allgather,broadcast,reducescatter" or "all"

apps/nccl/src/nccl.cu
```diff
@@ -417,12 +511,29 @@ NCCL_API ncclResult_t ncclCommInitRank(ncclComm_t* comm, int nranks, ncclUniqueI
...
+    } else {
+      return ncclInternalError;
+    }
+  }
+
+  ncclComm* commPtr = new ncclComm();
+  if (mscclppOpenSharedLib == true) {
+    nccl_ops.CommInitRank(&commPtr->nccl_comm, nranks, commId, rank);
+    *comm = commPtr->nccl_comm;
+    return ncclSuccess;
```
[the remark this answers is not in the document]
seagater: Resolved.

apps/nccl/src/nccl.cu
seagater: Resoled. Now we keep both mscclpp::UniqueId and internal ncclUniqueId.
Create a internal ncclUniqueId array and use mscclpp::TcpBootstrap::allgather() to get the ncclUniqueId created on rank:0 so that each rank can get the same ncclUniqueId. This one will be used to create nccl communicator.

apps/nccl/src/nccl.cu
seagater: Resolved. We call dlopen init in ncclCommInitRank and create a ncclUniqueId and send it to all ranks by mscclpp::TcpBootStrap::allgather().

apps/nccl/src/nccl.cu
seagater: Resolved.

apps/nccl/src/nccl.cu
```diff
@@ -479,6 +578,34 @@ NCCL_API ncclResult_t ncclCommInitRank(ncclComm_t* comm, int nranks, ncclUniqueI
...
+
+  if (mscclppNcclDlopenSharedLib == true) {
+    ncclUniqueId mscclppNcclUniqueId[nranks];
+    if (rank == 0) {
+      mscclppNcclOps.GetUniqueId(&mscclppNcclUniqueId[rank]);
+    }
+    // After allGather, mscclppNcclUniqueId[0] on each rank has the same ncclUniqueId
+    bootstrap->allGather(mscclppNcclUniqueId, sizeof(ncclUniqueId));
+
+    commPtr->mscclppNcclComm = malloc(sizeof(ncclComm_t));
```
Binyang2014, line 600: Using new/delete?

apps/nccl/src/nccl.cu
```diff
@@ -505,6 +632,13 @@ NCCL_API ncclResult_t ncclCommDestroy(ncclComm_t comm) {
     NpKit::Shutdown();
   }
 #endif
+
+  if (mscclppNcclDlopenSharedLib == true) {
+    mscclppNcclOps.CommDestroy(*reinterpret_cast<ncclComm_t*>(comm->mscclppNcclComm));
+    mscclppNcclDlopenFinalize();
+    free(comm->mscclppNcclComm);
```
Binyang2014, line 639: Change to new/delete pair

apps/nccl/src/nccl.cu
```diff
@@ -479,6 +578,34 @@ NCCL_API ncclResult_t ncclCommInitRank(ncclComm_t* comm, int nranks, ncclUniqueI
...
+    }
+  }
+
+  if (mscclppNcclDlopenSharedLib == true) {
+    ncclUniqueId mscclppNcclUniqueId[nranks];
+    if (rank == 0) {
+      mscclppNcclOps.GetUniqueId(&mscclppNcclUniqueId[rank]);
+    }
+    // After allGather, mscclppNcclUniqueId[0] on each rank has the same ncclUniqueId
+    bootstrap->allGather(mscclppNcclUniqueId, sizeof(ncclUniqueId));
```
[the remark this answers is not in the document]
seagater: Resolved. Add TcpBootstrap::Impl::broadcast and use broadcast for mscclppNcclUniqueId.

apps/nccl/src/nccl.cu
seagater: Resolved.

apps/nccl/src/nccl.cu
seagater: Resolved.

apps/nccl/src/nccl.cu
```diff
@@ -798,6 +951,13 @@ NCCL_API ncclResult_t ncclReduceScatter(const void* sendbuff, void* recvbuff, si
     return ncclInvalidArgument;
   }
 
+  const char* fallbackList = mscclpp::env()->forceNcclFallbackOperation.c_str();
+  const char* collOps = "reducescatter";
+  if (mscclppNcclDlopenSharedLib == true && mscclppNcclInFallbackList(collOps, fallbackList)) {
```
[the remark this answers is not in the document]
seagater: Resolved. Also update for "allreduce", "allgather", "broadcast".

Binyang2014 (approved):
LGTM

PR #476, opened 2025-03-04
Update allgather fallback algo

Binyang2014:
Enhancements to all-gather operation, a temporary solution to fix the memory overhead when integrating msccl++ with pytorch.
This solution will not register input/output buffer to msccl++, so the temp output buffer for allgather could be reused by torch automatically.

* Introduced a new `allgather8` kernel function in `apps/nccl/src/allgather.hpp` to handle larger data sizes more efficiently. This includes double buffering to hide synchronization overhead and support for both in-place and out-of-place operations.
* Modified the `allgather` function to decide between `allgather6` and `allgather8` based on data size and platform, improving performance for large data sizes.

Configuration and environment improvements:

* Added a new environment variable `MSCCLPP_DISABLE_CHANNEL_CACHE` to control whether the channel cache is disabled, enhancing configurability. This variable is now part of the `Env` class and is logged during environment initialization.
* Removed the redundant global variable `mscclppDisableChannelCache` from `src/debug.cc` and updated its usage to refer to the new environment variable.

### DIFF
--- a/apps/nccl/src/allgather.hpp
+++ b/apps/nccl/src/allgather.hpp
@@ -212,7 +212,7 @@ cudaError_t allgather(T* buff, T* scratch, T* resultBuff, mscclpp::DeviceHandle<
                       size_t channelInOffset, size_t channelOutOffset, int rank, int nRanksPerNode, int worldSize,
                       size_t nelems, cudaStream_t stream) {
   int nBlocks = 28;
-  if (nelems * sizeof(T) <= 64 * (1 << 20)) {
+  if (nelems * sizeof(T) <= 32 * (1 << 20)) {
     if (nelems <= 4096) {
       nBlocks = 7;
     } else if (nelems <= 32768) {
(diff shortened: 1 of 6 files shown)


apps/nccl/src/allgather.hpp
```diff
@@ -111,20 +111,130 @@ __global__ void __launch_bounds__(1024, 1)
...
+  if (lid < nPeer) {
+    channels[lid] = memoryChans[lid];
+  }
+  __syncwarp();
+  const int tid = threadIdx.x + blockIdx.x * blockDim.x;
+  // we can use double buffering to hide synchronization overhead
+  for (size_t itr = 0; itr < nItrs; itr++) {
+    if (threadIdx.x < static_cast<uint32_t>(nPeer)) {
+      channels[threadIdx.x].signal();
+      channels[threadIdx.x].wait();
```
caiomcbr, line 146: Should we use tid instead of threadIdx.x for signal and wait? For example, in BlockIdx.x 0 and BlockIdx.x 1, it has threadIdx.x 0, so in this case, it will send two signals on channel 0. So, the correct approach should be something like this, right?
channels[tid ].signal();
channels[tid ].wait();

Binyang2014: Oh, for different thread block, we use different channel, code is here: `auto memoryChans = memoryChannels + chanOffset`. It will use different channel

caiomcbr: You are right! I missed that, sorry.

caiomcbr (approved):
LGTM

PR #483, opened 2025-03-18
Fix correctness issue when mscclppDisableChannelCache set to true

Binyang2014:
If `mscclppDisableChannelCache` set to true, we need to keep every channel information avoid the channel info in GPU side be released.


seagater (approved):
Looks good to me. Seems like another kind of cache to save the channelInfo?

PR #497, opened 2025-04-07
Memory ordering

liangyuRain:
There appear to be two locations where the memory ordering is not guaranteed.
- In `DeviceSyncer`, the atomic load and store are in relaxed order. There seems to be no memory ordering guarantee across threadblocks. For example:
```
TB0                     | TB1
write global memory     |
DeviceSyncer.sync(...)  | DeviceSyncer.sync(...)
                        | read global memory
```
There appears to be no guarantee that TB1 sees the TB0's modification at the time of read. There is a fence before the spin loop in DeviceSyncer.sync to establish a release pattern, but no fence after the wait to establish an acquire pattern. The fence_acq_rel_gpu() may also be replaced by stronger __threadfence(), or possibly by weaker release fence plus acquire fence.
- In `FifoDeviceHandle`, the memory ordering of writing the trigger should be release? to match the acquire in
https://github.com/microsoft/mscclpp/blob/adc9ee56840a520289781fe8ac22673fe14fad19/src/fifo.cc#L46

### DIFF
--- a/include/mscclpp/concurrency_device.hpp
+++ b/include/mscclpp/concurrency_device.hpp
@@ -19,6 +19,14 @@ struct DeviceSyncer {
   ~DeviceSyncer() = default;
 
 #if defined(MSCCLPP_DEVICE_COMPILE)
+  MSCCLPP_DEVICE_INLINE void fence_acq_rel_gpu() {
+#if defined(__HIP_PLATFORM_AMD__)
+    __builtin_amdgcn_fence(mscclpp::memoryOrderAcqRel, "agent");
+#else
+    asm volatile("fence.acq_rel.gpu;":: : "memory");
+#endif
+  }
+
   /// Synchronize all threads inside a kernel. Guarantee that all previous work of all threads in cooperating blocks is
   /// finished.
   /// @param blockNum The number of blocks that will synchronize.
@@ -28,8 +36,8 @@ struct DeviceSyncer {
     __syncthreads();
     if (blockNum == 1) return;
     if (threadIdx.x == 0) {
-      // Need a `__threadfence()` before to flip `flag`.
-      __threadfence();
+      // Fence to establish release pattern
+      fence_acq_rel_gpu();
       unsigned int tmp = preFlag_ ^ 1;
       if (atomicInc(&count_, maxOldCnt) == maxOldCnt) {
         atomicStore(&flag_, tmp, memoryOrderRelaxed);
@@ -37,6 +45,8 @@ struct DeviceSyncer {
         POLL_MAYBE_JAILBREAK((atomicLoad(&flag_, memoryOrderRelaxed) != tmp), maxSpinCount);
       }
       preFlag_ = tmp;
+      // Fence to establish acquire pattern
+      fence_acq_rel_gpu();
     }
     // We need sync here because only a single thread is checking whether
     // the flag is flipped.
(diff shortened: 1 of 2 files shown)


Binyang2014:
Do we really need this? According to the doc: https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#memory-fence-functions. 
```
threadfence() ensures that no writes to all memory made by the calling thread after the call to __threadfence() are observed by any thread in the device as occurring before any write to all memory made by the calling thread before the call to __threadfence().
```

So if other blocks see the flag, they will observe the data write before threadfence()

Binyang2014:
For `DeviceSyncer.sync(...)`, the `__threadfence` will be executed in each threadblock, so for your case, it will be
```
TB0                     | TB1                   
Thread0                 | Thread0               | Thread1
write global memory     |                       |
__threadfence           | __threadfence         |
write to flag           | read from flag        |
                        | __syncthreads         | __syncthreads
                        |                       | read global memory
```

__threadfence and __syncthreads guarantee Tb1 thread 1 will read the correct data. You think it's not good enough?

liangyuRain:
The `__threadfence` of TB1 thread0 here is irrelevant as TB1 can be faster than TB0 such that `__threadfence` is even earlier than `write global memory` of TB0 thread0 in wall time.

The "Litmus Test: Message Passing" [here](https://docs.nvidia.com/cuda/parallel-thread-execution/#causality-axiom) in PTX documentation is exactly the scenario of DeviceSyncer. However, this minimal example also includes a memory fence after reading the flag. The `__syncthreads` in TB1 thread0 is a memory fence, but its scope is limited to block, not satisfying [memory synchronization](https://docs.nvidia.com/cuda/parallel-thread-execution/#memory-synchronization) item 4's [morally strong](https://docs.nvidia.com/cuda/parallel-thread-execution/#morally-strong-operations) requirement.

Binyang2014:
For fifo, please refer this PR: https://github.com/microsoft/mscclpp/pull/229. We don't need to consider the memory visibility for this. For `DeviceSyncer.sync(...)` I got your point, even we got the flag from TB1, since there is not fence we can not guarantee the order. 
Could you help to measure the performance? Is there any performance downgrade?

Binyang2014:
@liangyuRain I think you are correct about the memory ordering. @chhwang Can you help review this PR?

include/mscclpp/concurrency_device.hpp
```diff
@@ -28,15 +36,17 @@ struct DeviceSyncer {
     __syncthreads();
     if (blockNum == 1) return;
     if (threadIdx.x == 0) {
-      // Need a `__threadfence()` before to flip `flag`.
-      __threadfence();
+      // Fence to establish release pattern
+      fence_acq_rel_gpu();
```
Binyang2014, line 40: Can we move the release to  `atomicStore`?

include/mscclpp/concurrency_device.hpp
```diff
@@ -28,15 +36,17 @@ struct DeviceSyncer {
...
       unsigned int tmp = preFlag_ ^ 1;
       if (atomicInc(&count_, maxOldCnt) == maxOldCnt) {
         atomicStore(&flag_, tmp, memoryOrderRelaxed);
       } else {
         POLL_MAYBE_JAILBREAK((atomicLoad(&flag_, memoryOrderRelaxed) != tmp), maxSpinCount);
       }
       preFlag_ = tmp;
+      // Fence to establish acquire pattern
+      fence_acq_rel_gpu();
```
Binyang2014, line 49: Maybe we can move the acquire to atomicLoad function

Binyang2014: Thanks, let me try on my side

Binyang2014:
Hi @liangyuRain pls try this PR: #499, I think the reason of poor performance for acquire atomicLoad comes from we put too many instructions in while loop. I tested with this solution, and perf number looks good.

PR #498, opened 2025-04-09
Order DAG Implementation

caiomcbr:
Due to issues with the order in which operations are passed to MSCCLLang, bugs were occurring. This pre-order DAG structure aims to ensure that operations are executed in the correct order according to their dependencies.

### DIFF
--- a/python/mscclpp/language/program.py
+++ b/python/mscclpp/language/program.py
@@ -8,6 +8,8 @@ from mscclpp.language.types import DataFormat, ChannelType, ChunkRef, Replicatio
 from mscclpp.language.ir import *
 from mscclpp.language.dag import DagOptimizer, DagLower, InstructionDAG
 from mscclpp.language.rank import Rank
+from queue import Queue
+from typing import Dict, Tuple
 
 _current_program = None
 
@@ -54,6 +56,7 @@ class MSCCLPPProgram:
         # Initialize the input buffers
         self.buffers = collective.init_buffers()
         self.instr_dag = InstructionDAG(self.num_ranks, self.buffers)
+        self.order_dag = OrderDAG()
         self.ranks = []
         for r in range(self.num_ranks):
             self.ranks.append(Rank(r))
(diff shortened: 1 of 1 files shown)


python/mscclpp/language/program.py
```diff
@@ -430,7 +489,125 @@ def print_chunk_info(self, index=0):
         print(self._get_chunk(index + self.index))
 
 
+class OrderDAG:
```
Binyang2014, line 492: Better to move to another file

python/mscclpp/language/program.py
```diff
@@ -430,7 +489,125 @@ def print_chunk_info(self, index=0):
...
+            ].empty():
+                if (op.src.rank, op.dst.rank, tb) not in self.waiting:
+                    self.waiting[(op.src.rank, op.dst.rank, tb)] = Queue()
+                self.waiting[(op.src.rank, op.dst.rank, tb)].put(node)
+            else:
+                signalling_node = self.signalling[(op.src.rank, op.dst.rank, tb)].get()
+                signalling_node.next_nodes.append(node)
+                node.input += 1
+
+    def execute_operations(self):
```
Binyang2014, line 554: I think you try to implement a `topological-sorting` right? Maybe we can refer to some existing python implementation for this. I found a library for this: https://docs.python.org/3.10/library/graphlib.html. Maybe you can find smarter version

python/mscclpp/language/topo_sort.py
```diff
@@ -0,0 +1,167 @@
...
+        """
+        queue = Queue()
+        for node in self.start_nodes:
+            queue.put(node)
+
+        while not queue.empty():
+            node = queue.get()
+            op = node.operation
+
+            if op.inst == Instruction.put:
```
Binyang2014, line 89: I am thinking if we can have an abstract class, then we could call `op.exec(chunk, ...)`, instead of have many if-else blocks.

Binyang2014 (approved):

python/mscclpp/language/topo_sort.py
Binyang2014: Yes, but I am not sure if it difficult. But abstract operation class will make code clean

Issue #496, opened 2025-04-06
[Bug] IB Error after removing requirement for `CU_DEVICE_ATTRIBUTE_HANDLE_TYPE_FABRIC_SUPPORTED`


liangyuRain:
Hi, the following code is broken right after a3d8d6807b877f7b281bc750ffab7feefc63a17f
```python
import cupy as cp
from mscclpp import TcpBootstrap, Communicator, Transport
from mscclpp._mscclpp import RawGpuBuffer
bootstrap = TcpBootstrap.create(0, 1)
comm = Communicator(bootstrap)

cp.cuda.Device(0).use()
memory = RawGpuBuffer(1024 * 1024)
data_ptr = memory.data()
comm.register_memory(data_ptr, 1024 * 1024, Transport.IB0)
```
```bash
Traceback (most recent call last):                                                                             
  File "/root/bug.py", line 10, in <module>                                                                    
    comm.register_memory(data_ptr, 1024 * 1024, Transport.IB0)                                                 
mscclpp._mscclpp.IbError: (14, 'ibv_reg_mr failed (errno 14) (Ib failure: Bad address)')    
```
Environment:
```bash
(base) root@xxx:~# nvcc --version
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2024 NVIDIA Corporation
Built on Thu_Mar_28_02:18:24_PDT_2024
Cuda compilation tools, release 12.4, V12.4.131
Build cuda_12.4.r12.4/compiler.34097967_0
(base) root@xxx:~# nvidia-smi
Sun Apr  6 15:37:56 2025       
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 535.216.03             Driver Version: 535.216.03   CUDA Version: 12.4     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  NVIDIA H100 80GB HBM3          On  | 00000000:19:00.0 Off |                    0 |
| N/A   29C    P0              71W / 700W |      0MiB / 81559MiB |      0%      Default |
|                                         |                      |             Disabled |
+-----------------------------------------+----------------------+----------------------+
...8xH100
```

[...]

PR #499, opened 2025-04-11
Improve signal/wait performance and fix barrier issue

Binyang2014:
Remove __assert_fail for release build. This will reduce the number of PTX instructions inside the loop. Also Trying to resolve this issue reported in #497. Reduce the number of PTX instructions from 8 to 6. 
8 ranks signal/wait will reduce from 3.2us->2.8us on NDv5
Also NDEBUG flag is confused here, sometime it will not be set. Use customized flag for debug build.

Here is current PTX:
```
      ld.u64  %rd12, [%rd2+-24];
      mov.u64         %rd13, %rd12;
      mov.u64         %rd11, %rd13;
      ld.acquire.sys.b64 %rd10,[%rd11];
      setp.lt.u64     %p1, %rd10, %rd3;
      @%p1 bra        $L__BB2_1;
```

If we change to `asm volatile("ld.global.acquire.sys.b64 %0, [%1];" : "=l"(flag) : "l"(flag_addr));` will reduce to 4 instructions. We can get 2.1 us for 8 ranks signal/wait
```
        ld.u64  %rd9, [%rd1+-24];
        ld.global.acquire.sys.b64 %rd8, [%rd9];
        setp.lt.u64     %p1, %rd8, %rd2;
        @%p1 bra        $L__BB2_1;
```

### DIFF
--- a/include/mscclpp/concurrency_device.hpp
+++ b/include/mscclpp/concurrency_device.hpp
@@ -7,6 +7,8 @@
 #include "atomic_device.hpp"
 #include "poll_device.hpp"
 
+#include <stdio.h>
+
 namespace mscclpp {
 
 /// A device-wide barrier.
@@ -24,16 +26,15 @@ struct DeviceSyncer {
   /// @param blockNum The number of blocks that will synchronize.
   /// @param maxSpinCount The maximum number of spin counts before asserting. Never assert if negative.
   MSCCLPP_DEVICE_INLINE void sync(int blockNum, int64_t maxSpinCount = 100000000) {
-    unsigned int maxOldCnt = blockNum - 1;
+    int maxOldCnt = blockNum;
     __syncthreads();
     if (blockNum == 1) return;
     if (threadIdx.x == 0) {
       unsigned int tmp = preFlag_ ^ 1;
-      if (atomicInc(&count_, maxOldCnt) == maxOldCnt) {
-        atomicStore<unsigned int, scopeDevice>(&flag_, tmp, memoryOrderRelease);
-      } else {
-        POLL_MAYBE_JAILBREAK((atomicLoad<unsigned int, scopeDevice>(&flag_, memoryOrderAcquire) != tmp), maxSpinCount);
-      }
+      int val = (tmp << 1) - 1;
+      maxOldCnt = val == 1 ? maxOldCnt : 0;
+      atomicFetchAdd(&count_, val, memoryOrderRelease);
+      POLL_MAYBE_JAILBREAK((atomicLoad<int, scopeDevice>(&count_, memoryOrderAcquire) != maxOldCnt), maxSpinCount);
       preFlag_ = tmp;
     }
     // We need sync here because only a single thread is checking whether
@@ -43,10 +44,8 @@ struct DeviceSyncer {
 #endif  // !defined(MSCCLPP_DEVICE_COMPILE)
 
  private:
-  /// The flag to indicate whether the barrier is reached by the latest thread.
-  unsigned int flag_;
   /// The counter of synchronized blocks.
-  unsigned int count_;
+  int count_;
   /// The flag to indicate whether to increase or decrease @ref flag_.
   unsigned int preFlag_;
 };
--- a/include/mscclpp/concurrency_device.hpp
+++ b/include/mscclpp/concurrency_device.hpp
@@ -31,7 +31,7 @@ struct DeviceSyncer {
       unsigned int tmp = preFlag_ ^ 1;
       int val = (tmp << 1) - 1;
       targetCnt = val == 1 ? targetCnt : 0;
-      atomicFetchAdd(&count_, val, memoryOrderRelease);
+      atomicFetchAdd<int, scopeDevice>(&count_, val, memoryOrderRelease);
       POLL_MAYBE_JAILBREAK((atomicLoad<int, scopeDevice>(&count_, memoryOrderAcquire) != targetCnt), maxSpinCount);
       preFlag_ = tmp;
     }
--- a/include/mscclpp/concurrency_device.hpp
+++ b/include/mscclpp/concurrency_device.hpp
@@ -28,13 +28,11 @@ struct DeviceSyncer {
     __syncthreads();
     if (blockNum == 1) return;
     if (threadIdx.x == 0) {
-      // Need a `__threadfence()` before to flip `flag`.
-      __threadfence();
       unsigned int tmp = preFlag_ ^ 1;
       if (atomicInc(&count_, maxOldCnt) == maxOldCnt) {
-        atomicStore(&flag_, tmp, memoryOrderRelaxed);
+        atomicStore<unsigned int, scopeDevice>(&flag_, tmp, memoryOrderRelease);
       } else {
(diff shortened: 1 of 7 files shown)


liangyuRain:
Hi @Binyang2014 the pull request looks good to me. However, I noticed changing `NDEBUG` to `DEBUG_BUILD` causes compile error for any assert in my program. It looks like when building mscclpp with pip install, there are macros still using NDEBUG. I am cherry picking the commits on top of 0f21ed4 given #496.

I am unable to benchmark the code. I am curious if memoryOrderRelaxed atomicLoad is faster than memoryOrderAcquire one. If so, maybe it is more efficient to use memoryOrderRelaxed atomicLoad in the `POLL_MAYBE_JAILBREAK` and have an acquire fence only once after the loop.
https://github.com/microsoft/mscclpp/blob/43b4441bcb28075a64382790b30fd055988226bf/include/mscclpp/concurrency_device.hpp#L35

include/mscclpp/concurrency_device.hpp
```diff
@@ -28,13 +28,11 @@ struct DeviceSyncer {
     __syncthreads();
     if (blockNum == 1) return;
     if (threadIdx.x == 0) {
-      // Need a `__threadfence()` before to flip `flag`.
-      __threadfence();
       unsigned int tmp = preFlag_ ^ 1;
       if (atomicInc(&count_, maxOldCnt) == maxOldCnt) {
-        atomicStore(&flag_, tmp, memoryOrderRelaxed);
+        atomicStore<unsigned int, scopeDevice>(&flag_, tmp, memoryOrderRelease);
```
chhwang, line 33: This means now we have a fence only for a single block, which is incorrect

test/unit/compile_tests.cu
```diff
@@ -3,7 +3,7 @@
 
 #include <gtest/gtest.h>
 
-#undef NDEBUG
+#undef DEBUG_BUILD
```
chhwang, line 6: This test is for checking whether `assert.h` and `poll_device.hpp` can be included at the same time, so we should do:
```cpp
#undef NDEBUG
#ifndef DEBUG_BUILD
#define DEBUG_BUILD
#endif  // DEBUG_BUILD
```

chhwang (changes requested):

include/mscclpp/concurrency_device.hpp
Binyang2014: According to https://docs.nvidia.com/cuda/cuda-c-programming-guide/#memory-fence-functions.  It's for device level does not thread block level. Since we just sync thread blocks inside one device, `cuda::memory_order_seq_cst, cuda::thread_scope_device` should be enough.

void __threadfence(); is equivalent to [cuda::atomic_thread_fence(cuda::memory_order_seq_cst, cuda::thread_scope_device)](https://nvidia.github.io/libcudacxx/extended_api/synchronization_primitives/atomic/atomic_thread_fence.html) and ensures that no writes to all memory made by the calling thread after the call to __threadfence() are observed by any thread in the device as occurring before any write to all memory made by the calling thread before the call to __threadfence().

Binyang2014:
> Hi @Binyang2014 the pull request looks good to me. However, I noticed changing `NDEBUG` to `DEBUG_BUILD` causes compile error for any assert in my program. It looks like when building mscclpp with pip install, there are macros still using NDEBUG. I am cherry picking the commits on top of [0f21ed4](https://github.com/microsoft/mscclpp/commit/0f21ed44b8366620492a3c0e76baf6b390be8450) given #496.
> 
> I am unable to benchmark the code. I am curious if memoryOrderRelaxed atomicLoad is faster than memoryOrderAcquire one. If so, maybe it is more efficient to use memoryOrderRelaxed atomicLoad in the `POLL_MAYBE_JAILBREAK` and have an acquire fence only once after the loop.
> 
> https://github.com/microsoft/mscclpp/blob/43b4441bcb28075a64382790b30fd055988226bf/include/mscclpp/concurrency_device.hpp#L35

Yes, it's faster. But I am not sure the correctness for this case. If we get the flag via memoryOrderRelaxed atomicLoad, can system guarantee the following  atomicLoad memoryOrderAcquire will see same value? I don't find any document for this

include/mscclpp/concurrency_device.hpp
liangyuRain: I think @chhwang is correct. If we remove the __threadfence, then all blocks except the last block doing the atomicStore will not have any device scope release memory fence, as the atomicInc is by default memoryOrderRelaxed.

chhwang:
> I am suggesting something similar to #497 that we keep the memoryOrderRelaxed atomicStore and atomicLoad, but add fence before and after. For documentation from cuda, please refer to [#497 (comment)](https://github.com/microsoft/mscclpp/pull/497#issuecomment-2787540404), especially item 3 of [8.8](https://docs.nvidia.com/cuda/parallel-thread-execution/#release-and-acquire-patterns). Basically, if one thread writes a value that is read by another thread and there are fences before the write and after the read, then the two threads have release-acquire relation even if the read and write are relaxed. We should establish a release-acquire relation between _**any two threadblocks**_ participating in the DeviceSyncer. I think we should probably also add fence2 to #497 like the following:
> 
> ```c++
> fence_acq_rel_gpu(); // fence1
> unsigned int tmp = preFlag_ ^ 1;
> if (atomicInc(&count_, maxOldCnt) == maxOldCnt) {
>   fence_acq_rel_gpu(); // fence2
>   atomicStore(&flag_, tmp, memoryOrderRelaxed);
> } else {
>   POLL_MAYBE_JAILBREAK((atomicLoad(&flag_, memoryOrderRelaxed) != tmp), maxSpinCount);
> }
> preFlag_ = tmp;
> fence_acq_rel_gpu(); // fence3
> ```
> 
> Suppose the threadblocks reach the atomicInc in order 1, ..., n. Then from any block x to block y, the release-acquire relation has two cases:
> 
> * x<y:
>   
>   * block x `fence1 + atomicInc` release -> block y `atomicInc + fence3` acquire
> * x>y:
>   
>   * block x `fence1 + atomicInc` release -> block n `atomicInc + fence2` acquire
>   * block n `fence2 + atomicStore` release -> block y `atomicLoad + fence3` acquire
> 
> And of course, this only establishes the release-acquire between thread0 of all blocks. The __syncthreads before and after all of these will transition the release-acquire to between any two threads of all blocks.

`atomicInc` is only for picking the latest block. I don't think we need release-acquire relation there.

include/mscclpp/concurrency_device.hpp
Binyang2014: @liangyuRain @chhwang , pls help to review again, use `atomicFetchAdd(&count_, val, memoryOrderRelease);` and `atomicStore(remoteInboundSemaphoreId, semaphoreGetLocal(), memoryOrderRelease);` for barrier

liangyuRain:
Looks good to me. The release atomicFetchAdd and acquire atomicLoad, both on the same memory location, can establish mutual release-acquire between threadblocks. Please try replacing the memoryOrderAcquire atomicLoad in the poll loop with a memoryOrderRelaxed one and add an acquire fence after poll loop ends, if this brings performance improvement. The correctness is guaranteed by item 3 of [8.8](https://docs.nvidia.com/cuda/parallel-thread-execution/#release-and-acquire-patterns) acquire pattern.

chhwang (approved):

PR #501, opened 2025-04-15
Fix the virtual address mapping issue of cuMemMap in fallback code

seagater:
Fix the virtual address mapping issue of cuMemMap by using the size of device memory allocation instead of user buffer size


src/registered_memory.cc
```diff
@@ -57,6 +57,7 @@ RegisteredMemory::Impl::Impl(void* data, size_t size, TransportFlags transports,
     void* baseDataPtr;
     size_t baseDataSize;  // dummy
```
Binyang2014, line 58: Remove the `dummy` in comments

Binyang2014 (approved):

src/registered_memory.cc
seagater: Removed.

Binyang2014 (approved):

PR #502, opened 2025-04-15
apps/nccl: fix a bug in allreduce kernels for graph mode

nusislam:
`allreduce7` and `allreduceAllpairs` kernels were updating the LL protocol flag on the host side. So, it was not properly captured in graph mode. This PR fixes the issue by updating the flag in the kernels.

### DIFF
--- a/apps/nccl/src/allreduce.hpp
+++ b/apps/nccl/src/allreduce.hpp
@@ -168,11 +168,17 @@ template <Op OpType, typename T>
 __global__ void allreduceAllPairs(T* buff, T* scratch, T* resultBuff,
                                   mscclpp::DeviceHandle<mscclpp::MemoryChannel>* memoryChannels,
                                   size_t channelDataOffset, size_t channelScratchOffset, int rank, int nRanksPerNode,
-                                  int worldSize, size_t nelems, uint32_t flag) {
+                                  int worldSize, size_t nelems, uint32_t* deviceFlag, uint32_t numScratchBuff) {
   // This version of allreduce only works for single nodes
   if (worldSize != nRanksPerNode) return;
   if (sizeof(T) == 2) nelems = (nelems * sizeof(T) + sizeof(T)) / sizeof(int);
   const int nPeers = nRanksPerNode - 1;
+
+  uint32_t flag = deviceFlag[blockIdx.x];
+
+  size_t scratchBaseOffset = (flag % numScratchBuff) ? SCRATCH_SIZE/numScratchBuff : 0;
+  channelScratchOffset = scratchBaseOffset;
+
   const int nBlocksPerPeer = gridDim.x / nPeers;
   const int localBlockIdx = blockIdx.x % nBlocksPerPeer;
   const int tid = threadIdx.x + localBlockIdx * blockDim.x;
@@ -198,13 +204,17 @@ __global__ void allreduceAllPairs(T* buff, T* scratch, T* resultBuff,
     }
     dst[idx] = data;
   }
+  __syncthreads();
+  if (threadIdx.x == 0) {
+  	deviceFlag[blockIdx.x] = deviceFlag[blockIdx.x] + 1;
+  }
 }
 
 template <Op OpType, typename T>
 __global__ void __launch_bounds__(1024, 1)
     allreduce7(T* buff, T* scratch, T* resultBuff, mscclpp::DeviceHandle<mscclpp::MemoryChannel>* memoryChannels,
                size_t channelDataOffset, size_t channelScratchOffset, int rank, int nRanksPerNode, int worldSize,
-               size_t nelems, uint32_t flag
+               size_t nelems, uint32_t* deviceFlag, uint32_t numScratchBuff
 #if defined(ENABLE_NPKIT)
                ,
                NpKitEventCollectContext* npKitEventCollectContexts, uint64_t* cpuTimestamp) {
@@ -246,6 +256,11 @@ __global__ void __launch_bounds__(1024, 1)
 
   const int nPeers = nRanksPerNode - 1;
   const size_t nPkts = nelems / 2;
+
+  uint32_t flag = (uint32_t) deviceFlag[blockIdx.x];
+
+  size_t scratchBaseOffset = (flag % numScratchBuff) ? SCRATCH_SIZE/numScratchBuff : 0;
+  channelScratchOffset = scratchBaseOffset;
 
   int nelemsPerRank = nelems / worldSize;
   if ((nelemsPerRank % 2)) nelemsPerRank = (nelemsPerRank * sizeof(T) + sizeof(T)) / sizeof(T);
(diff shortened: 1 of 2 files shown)


apps/nccl/src/allreduce.hpp
```diff
@@ -478,13 +485,20 @@ __global__ void __launch_bounds__(32, 1)
     }
     dst[idx] = data;
   }
+  __syncthreads();
+
+  deviceSyncer->sync(gridDim.x);
```
Binyang2014, line 490: Do we need deviceSyncer here? If we run allreduce sequentially, it will be launched in different kernel with same stream. And `Kernels launched into the same stream are guaranteed to execute consecutively`

apps/nccl/src/nccl.cu
```diff
@@ -524,6 +529,13 @@ static void ncclCommInitRankFallbackSingleNode(ncclComm* commPtr, std::shared_pt
   commPtr->scratchBuff = mscclpp::GpuBuffer<char>(SCRATCH_SIZE).memory();
   commPtr->remoteScratchRegMemories =
       setupRemoteMemories(commPtr->comm, rank, commPtr->scratchBuff.get(), SCRATCH_SIZE, mscclpp::Transport::CudaIpc);
+
+  hipMalloc((void**)&(commPtr->syncer), sizeof(mscclpp::DeviceSyncer));
+  hipMemset((void*)(commPtr->syncer), 0, sizeof(mscclpp::DeviceSyncer));
+
+  uint64_t initFlag = 1;
+  hipMalloc((void**)&(commPtr->deviceFlag), sizeof(uint64_t));
+  hipMemcpy((void*)(commPtr->deviceFlag), &initFlag, sizeof(uint64_t), hipMemcpyHostToDevice);
```
Binyang2014, line 538: Maybe use cudaMalloc here, we already map cudaMalloc, cudaFree... to hipMalloc, hipFree

nusislam: will do.

apps/nccl/src/allreduce.hpp
Binyang2014: I think kernel launch works like a barrier. Before we launch another kernel, it will make sure the previous barrier finished its' work. Here is the doc for this: https://docs.nvidia.com/cuda/cuda-c-programming-guide/#managing-data-visibility-and-concurrent-cpu-gpu-access-with-streams `Kernels launched into the same stream are guaranteed to execute consecutively,`

nusislam: I am not worried about subsequent kernel launch. What if `block-0` is faster than the other blocks in the same kernel launch?

Binyang2014: Okay, got you point. If block-0 is faster than the other blocks, it will update the deviceFlag. But since other threadblock still use cached flag, seems there is no issue? Other threadblock will use the latest deviceFlag in next kernel launch

apps/nccl/src/nccl.cu
```diff
@@ -524,6 +531,16 @@ static void ncclCommInitRankFallbackSingleNode(ncclComm* commPtr, std::shared_pt
   commPtr->scratchBuff = mscclpp::GpuBuffer<char>(SCRATCH_SIZE).memory();
   commPtr->remoteScratchRegMemories =
       setupRemoteMemories(commPtr->comm, rank, commPtr->scratchBuff.get(), SCRATCH_SIZE, mscclpp::Transport::CudaIpc);
+
+  uint32_t initFlag[56];
+  for (int i = 0; i < 56; i++) {
+	initFlag[i] = 1;
+  }
+  cudaMalloc((void**)&(commPtr->deviceFlag28), sizeof(uint32_t) * 28);
```
Binyang2014, line 539: Maybe we can use ` mscclpp::detail::gpuCallocShared` to allocate memory, it will return shardPtr, then we don't need to care about release buffer

chhwang: +1

apps/nccl/src/nccl.cu
```diff
@@ -194,6 +194,9 @@ struct ncclComm {
   uint32_t numScratchBuff;
   uint32_t buffFlag;
 
+  uint32_t* deviceFlag28;
+  uint32_t* deviceFlag56;
```
chhwang, line 198: Please use shared pointers as Binyang's comment.

apps/nccl/src/allreduce.hpp
```diff
@@ -441,11 +441,17 @@ template <typename T>
...
+                      Op op, size_t nelems, uint32_t* deviceFlag) {
   // This version of allreduce only works for single nodes
   if (worldSize != nRanksPerNode) return;
   if (sizeof(T) == 2) nelems = (nelems * sizeof(T) + sizeof(T)) / sizeof(int);
   const int nPeers = nRanksPerNode - 1;
+
+  uint32_t flag = deviceFlag[blockIdx.x];
+
+  size_t scratchBaseOffset = (flag % 2) ? SCRATCH_SIZE/2 : 0;
+  channelScratchOffset = scratchBaseOffset;
```
chhwang, line 453: Please use `comm->numScratchBuff` instead of the magic number `2`.

apps/nccl/src/allreduce.hpp
```diff
@@ -478,13 +484,18 @@ __global__ void __launch_bounds__(32, 1)
     }
     dst[idx] = data;
   }
+  __syncthreads();
```
chhwang, line 487: Do we need this `__syncthreads()`?

apps/nccl/src/allreduce.hpp
```diff
@@ -527,6 +538,11 @@ __global__ void __launch_bounds__(1024, 1)
   const int nPeers = nRanksPerNode - 1;
   const size_t nPkts = nelems / 2;
 
+  uint32_t flag = (uint32_t) deviceFlag[blockIdx.x];
+
+  size_t scratchBaseOffset = (flag % 2) ? SCRATCH_SIZE/2 : 0;
```
chhwang, line 543: Please use `comm->numScratchBuff` instead of the magic number `2`.

apps/nccl/src/allreduce.hpp
```diff
@@ -589,6 +606,7 @@ __global__ void __launch_bounds__(1024, 1)
     result[idx].x = data.x;
     result[idx].y = data.y;
   }
+  __syncthreads();
```
chhwang, line 609: Do we need this `__syncthreads()`?

apps/nccl/src/allreduce.hpp
```diff
@@ -580,6 +596,7 @@ __global__ void __launch_bounds__(1024, 1)
       channels[index].write(offset, packet);
     }
   }
+  __syncthreads();
```
chhwang, line 599: Do we need this `__syncthreads()`?

apps/nccl/src/allreduce.hpp
nusislam: We need `__syncthreads()`. Otherwise, `threadIdx.x==0` may be out of sync and update the flag before other threads sees the previous value.

Binyang2014 (approved):
LGTM

apps/nccl/src/allreduce.hpp
nusislam: Done.

chhwang (approved):
LGTM

PR #505, opened 2025-04-18
Fix performance issue introduced in PR: 499

Binyang2014:
1. use `fence+relaxed` to replace `release` for fifo. `fence+relax` is more efficient on A100
2. Update the deviceSyncer. Previous one cannot handle threadBlock number change correctly. Use three counters to solve this issue. Reset previous counter before sync on current counter.
3. Introduce relaxedWait which can be used with relaxedSignal for case doesn't need guarantee the memory visibility

### DIFF
--- a/include/mscclpp/concurrency_device.hpp
+++ b/include/mscclpp/concurrency_device.hpp
@@ -26,7 +26,7 @@ struct DeviceSyncer {
   /// @param blockNum The number of blocks that will synchronize.
   /// @param maxSpinCount The maximum number of spin counts before asserting. Never assert if negative.
   MSCCLPP_DEVICE_INLINE void sync(int blockNum, int64_t maxSpinCount = 100000000) {
-    int targetCnt = blockNum;
+    unsigned int targetCnt = blockNum;
     __syncthreads();
     if (blockNum == 1) return;
     if (threadIdx.x == 0) {
--- a/include/mscclpp/concurrency_device.hpp
+++ b/include/mscclpp/concurrency_device.hpp
@@ -28,11 +28,13 @@ struct DeviceSyncer {
     __syncthreads();
     if (blockNum == 1) return;
     if (threadIdx.x == 0) {
-      unsigned int tmp = preFlag_ ^ 1;
-      int val = (tmp << 1) - 1;
-      targetCnt = val == 1 ? targetCnt : 0;
-      atomicFetchAdd<int, scopeDevice>(&count_, val, memoryOrderRelease);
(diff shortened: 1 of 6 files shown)


Binyang2014:
@liangyuRain pls help to review

include/mscclpp/fifo_device.hpp
liangyuRain, line 1: Surprised to learn this, given threadfence is offering the stronger sequential consistency than release. Maybe also consider using `asm volatile("fence.acq_rel.gpu;":: : "memory");`? PTX documentation [link](https://docs.nvidia.com/cuda/parallel-thread-execution/#parallel-synchronization-and-communication-instructions-membar-fence) says
> `fence.acq_rel` is a light-weight fence that is sufficient for memory synchronization in most programs.
> ...
> `fence.sc` is a slower fence that can restore sequential consistency when used in sufficient places, at the cost of performance.

liangyuRain:
Left a comment. Otherwise looks good to me.

include/mscclpp/fifo_device.hpp
Binyang2014: I think it's a bug on A100. H100 doesn't have such issue, they provide similar performance.

chhwang: I would prefer making the current version specific to older CUDA architectures and keep the `release` version as the default.

chhwang (approved):

include/mscclpp/concurrency_device.hpp
```diff
@@ -28,11 +30,13 @@ struct DeviceSyncer {
...
-      int val = (tmp << 1) - 1;
-      targetCnt = val == 1 ? targetCnt : 0;
-      atomicFetchAdd<int, scopeDevice>(&count_, val, memoryOrderRelease);
-      POLL_MAYBE_JAILBREAK((atomicLoad<int, scopeDevice>(&count_, memoryOrderAcquire) != targetCnt), maxSpinCount);
+      unsigned int tmp = (preFlag_ + 1) % NUM_DEVICE_SYNCER_COUNTER;
+      unsigned int next = (tmp + 1) % NUM_DEVICE_SYNCER_COUNTER;
+      unsigned int* count = &count_[tmp];
+      count_[next] = 0;
+      atomicFetchAdd<unsigned int, scopeDevice>(count, 1U, memoryOrderRelease);
+      POLL_MAYBE_JAILBREAK((atomicLoad<unsigned int, scopeDevice>(count, memoryOrderAcquire) != targetCnt),
```
chhwang, line 38: Please change `targetCnt` data type to `unsigned int`

chhwang (approved):

PR #506, opened 2025-04-21
Optimized allreduce fallback for ~10KB sizes

chhwang:
* Pass the op type as a template parameter
* Use the all-pairs algorithm for ~10KB
* Don't write channel handles on the shared memory for small sizes
* A reduction bug fix & cleanup


Binyang2014 (approved):
LGTM

PR #508, opened 2025-04-22
Revised MemoryChannel interfaces

chhwang:
* Moved the `MemoryChannel::copy()` method out of the `MemoryChannel` as a standalone function.
* Renamed `mscclpp::putPackets()` and `mscclpp::getPackets()` to `mscclpp::copyToPackets()` and `mscclpp::copyFromPackets()` respectively for consistency.
* Renamed `MemoryChannel::getPackets()` to `MemoryChannel::unpackPackets()` for clarity. Renamed `getPacketBuffer` to `packetBuffer`.
* Added the `MemoryChannel::unpackPacket()` method that unpacks one packet in the buffer.
* Added the `BaseMemoryChannel` class that only contains a semaphore without memory addresses.
* Removed the `MemoryDevice2DeviceSemaphoreDeviceHandle::signalPacket()` method that is lacking use cases.


Binyang2014 (approved):
LGTM, One question is how to use BaseMemoryChannel.
Seems we could use like this right?
```cpp
void kernel(mscclpp::DeviceHandle<BaseMemoryChannel> ch, src /*local src*/, dst /*remote dst*/) {
  ch[peer_id].signal;
  ch[peer_id].wait;
  mscclpp::copy(dst, src, size);
}
```

chhwang:
> LGTM, One question is how to use BaseMemoryChannel. Seems we could use like this right?
> 
> ```c++
> void kernel(mscclpp::DeviceHandle<BaseMemoryChannel> ch, src /*local src*/, dst /*remote dst*/) {
>   ch[peer_id].signal;
>   ch[peer_id].wait;
>   mscclpp::copy(dst, src, size);
> }
> ```

Yes this is how I intend

Binyang2014 (approved):

PR #509, opened 2025-04-23
Use a stream pool for `gpuCalloc*()`

chhwang:
Previous `gpuCalloc*()` creates a new stream for each allocation, which messes the timeline up in profiler traces. Now `GpuStreamPool` allows reusing the temporal streams.

### DIFF
--- a/src/gpu_utils.cc
+++ b/src/gpu_utils.cc
@@ -28,6 +28,29 @@ void CudaStreamWithFlags::set(unsigned int flags) {
 
 bool CudaStreamWithFlags::empty() const { return stream_ == nullptr; }
 
+GpuStream::GpuStream(std::shared_ptr<GpuStreamPool> pool, std::shared_ptr<CudaStreamWithFlags> stream)
+    : pool_(pool), stream_(stream) {}
+
+GpuStream::~GpuStream() { pool_->streams_.push_back(stream_); }
+
+GpuStreamPool::GpuStreamPool() {}
+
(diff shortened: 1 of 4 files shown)


src/gpu_utils.cc
```diff
@@ -28,6 +28,29 @@ void CudaStreamWithFlags::set(unsigned int flags) {
...
 bool CudaStreamWithFlags::empty() const { return stream_ == nullptr; }
 
+GpuStream::GpuStream(std::shared_ptr<GpuStreamPool> pool, std::shared_ptr<CudaStreamWithFlags> stream)
+    : pool_(pool), stream_(stream) {}
+
+GpuStream::~GpuStream() { pool_->streams_.push_back(stream_); }
+
+GpuStreamPool::GpuStreamPool() {}
+
+GpuStream GpuStreamPool::getStream(unsigned int flags) {
```
Binyang2014, line 38: When we try to get a stream from a stream pool, do we need to find the stream with same flag? For example:
```cpp
auto stream = gpuStreamPool()->getStream(cudaStreamDefault);
stream = nullptr
auto stream = gpuStreamPool()->getStream(cudaStreamNonBlocking);
```
Current logic will reuse the stream with `cudaStreamDefault`  flag

chhwang: Fixed to provide only non-blocking streams

Binyang2014 (approved):

PR #514, opened 2025-04-25
Asynchronous setup

chhwang:
Cherry-picked a part of features from #167: now `Communicator::setup()` is unneeded. `Communicator::sendMemory()` conducts the task inline, and `Communicator::recvMemory()` and `Communicator::connect()` conducts the task asynchronously without explicit setup.


Binyang2014:
Try to understand this PR correctly. It only addresses the connection phase and does not resolve the issue of supporting on-side registered memory. In the send/recv case, both ranks still need to send their registered memory information to the peer, even though the receiving side does not require the peer's registered memory right?

chhwang:
> Try to understand this PR correctly. It only addresses the connection phase and does not resolve the issue of supporting on-side registered memory. In the send/recv case, both ranks still need to send their registered memory information to the peer, even though the receiving side does not require the peer's registered memory right?

This is not what this PR addresses. What you are talking about is the channel's requirement. This PR is only about simplifying the initialization phase.

Binyang2014 (approved):

PR #519, opened 2025-05-05
[Draft] DSL design for next version

Binyang2014:


docs/design/mscclpp-dsl-next.md
```diff
@@ -0,0 +1,153 @@
+# MSCCL++ DSL Concepts
+### Channel
+All cross ranks related op is done through channels. A channel is a communication medium between ranks. It can be a network socket, shared memory, or any other form of IPC. The channel is responsible for sending and receiving messages between ranks.
+
+Exmpales:
```
caiomcbr, line 5: Typo: Exmpales -> Examples

docs/design/mscclpp-dsl-next.md
```diff
@@ -0,0 +1,153 @@
...
+```python
+channel = Channel(dst_rank, src_rank, channel_type, tag)
+channel.put(dst_chunk, src_chunk, tb=0)
```
caiomcbr, line 8: For the put operation, should we include the size?

docs/design/mscclpp-dsl-next.md
```diff
@@ -0,0 +1,153 @@
...
+```
+
+Here is the example for two ranks synchonization with each others.
+```python
+nranks = 2
+for i in range(nranks):
+    src_rank = i
+    dst_rank = (i + 1) % nranks
+    channel = Channel(dst_rank, src_rank, channel_type=Channel.memory, tag=0)
+    channel.relaxSignal(tb=0, sync=None)
```
caiomcbr, line 20: What means this sync parameter?

docs/design/mscclpp-dsl-next.md
```diff
@@ -0,0 +1,153 @@
...
+    channel.relaxWait(tb=0, sync="after")
+```
+
+#### For nvls based channel
+NVLS channel need to bind with a group of buffers from a set of ranks. Each operation on the channel will be performed on the buffers in the group.
+
+```python
+nvls_chan = SwitchChannel(rank_list=[], buffer=Buffer.input, tag=0) # this interface may need to refine
```
caiomcbr, line 28: Our current SwitchChannel operations only support the same index and buffer parameters for all the chunks. In the future, do we plan to support this operation using different indexes and buffers? I mean, if we will support this in the future, maybe we can use chunk_list instead of rank_list.

docs/design/mscclpp-dsl-next.md
```diff
@@ -0,0 +1,153 @@
...
+
+    # copy data back to output buffer
+    chan.signal(tb=0, sync="before")
+    chan.wait(tb=0, sync="after")
+    dst_chunk = Chunk(src_rank, Buffer.output, chunk_index, 1)
+    src_chunk = Chunk(src_rank, Buffer.scratch, chunk_index, 1)
+    rank.copy(dst_chunk, src_chunk, tb=0)
+```
+
+### Chunk/Tensor
```
caiomcbr, line 66: It might be helpful to explain this section before the allreduce example, as we use the chunk definition in the example.

docs/design/mscclpp-dsl-next.md
Binyang2014: We already have size info in chunk, maybe do no need here

docs/design/mscclpp-dsl-next.md
Binyang2014: It means if it requires threadblock synchronization before/after the signal operation

docs/design/mscclpp-dsl-next.md
```diff
@@ -0,0 +1,153 @@
...
+
+```python
+dst = Chunk(rank, index, size)
+scr = Chunk(rank, index + 1, size)
+Rank.copy(dst_chunk, src_chunk, tb=0)
+Rank.reduce(dst_chunk, src_chunk, op="sum", tb=0)
+```
+
+### Synchronization
+We provide some synchronization primitives to sync threadblocks inside a rank. The synchronization is done through a barrier or semaphore. The barrier is used to synchronize all threadblocks in a rank, while the semaphore is used to synchronize a specific threadblock.
```
caiomcbr, line 77: I think this part is a bit confusing. The barrier does not synchronize all the thread blocks in the rank, and the semaphore can asynchronously signal and wait between the thread blocks. Perhaps we could improve this to something like: The barrier ensures synchronization of a set of thread blocks within the rank, while the semaphore synchronizes threads within a thread block and allows asynchronous signaling and waiting between thread blocks.

docs/design/mscclpp-dsl-next.md
```diff
@@ -0,0 +1,153 @@
...
+```python
+rank.barrier(tb_list=[])
+sem = Rank.Semaphore(rank=0, size=1, tag=0)
+sem.acquire(tb=0)
+sem.release(tb=0)
+```
+
+The synchronization inside the thread-block can be inferred by MSCCL++ DSL auotmaticly. Which mean if we have data dependence between two operations, we will insert a synchronization point between them.
```
caiomcbr, line 85: Typo: auotmaticly -> automatically

docs/design/mscclpp-dsl-next.md
```diff
@@ -0,0 +1,153 @@
...
+
+
+## For Pipeline Loop
+For some cases, we need to pipeline the kernel to overlap some operations. For example, the first stage is copy data from input buffer to scratch buffer, the second stage is transfer data from scratch buffer to other peers. We could use `Rank.semphore` to synchronize the two stages. 
+```python
+sem = Rank.Semaphore(rank=0, size=1, tag=0)
+Rank.copy(dst_chunk, src_chunk, tb=0)
+sem.release(tb=0)
+channel = Channel(dst_rank, src_rank, channel_type, tag)
+sme.acquire(tb=1)
```
caiomcbr, line 100: Typo: sme.acquire(tb=1) -> sem.acquire(tb=1)

docs/design/mscclpp-dsl-next.md
```diff
@@ -0,0 +1,153 @@
...
+channel.put(dst_chunk, src_chunk, tb=1)
+```
+
+Also we could provide some gramar sugar to make the pipeline more readable. For example, we could use `Loop` to construct the pipeline. 
+```python
+sem = Rank.Semaphore(rank=rank, size=1, tag=0)
+with Loop.iteration(unit=2**20, num_chunks=1) as iter:
+    # the dst_chunk and src_chunk size but same as loop context
```
caiomcbr, line 108: Do you mean the dst_chunk and src_chunk sizes must match the loop context size, meaning they should have the same value as the num_chunks?

docs/design/mscclpp-dsl-next.md
```diff
@@ -0,0 +1,153 @@
...
+        sem0.release(tb=0)
+
+        # do allreduce in scratch buffer
+        sem0.acquire(tb=1, sync="after")
+        nvls_chan.group_load_reduce(chunk_index, 1, op="sum", tb=0, iter_context=iter)
+        nvls_chan.group_store(chunk_index, 1, tb=0, iter_context=iter)
+        chan1.signal(tb=1, sync="before")
+        sem1.release(tb=1)
+
+        # copy data back to output buffer
```
caiomcbr, line 147: Could you tell me where the section is that copies the reduced data from the remote rank to the local output buffer?

docs/design/mscclpp-dsl-next.md
Binyang2014: We can discuss this. For current NVLS implementation, the shared address space needs to bind with ranks at the creation time. Also the buffer is bound the nvls channel, the channel will not be reused by other chunks

docs/design/mscclpp-dsl-next.md
Binyang2014: Oh, it's done via `nvls_chan.group_store`. This will broadcast reduced data to all other ranks

docs/design/mscclpp-dsl-next.md
```diff
@@ -0,0 +1,154 @@
+# MSCCL++ DSL Concepts
+### Chunk/Tensor
+Chunk is a data structure that holds the data to be sent or received. For some local operations, such as copy/reduce we provide some functions to manipulate the chunk.
+
+```python
+dst = Chunk(rank, index, size)
+scr = Chunk(rank, index + 1, size)
+Rank.copy(dst_chunk, src_chunk, tb=0)
+Rank.reduce(dst_chunk, src_chunk, op="sum", tb=0)
```
caiomcbr, line 9: dst_chunk -> dst
src_chunk -> src

docs/design/mscclpp-dsl-next.md
```diff
@@ -0,0 +1,175 @@
+# MSCCL++ DSL Concepts
+### Chunk/Tensor
+Chunk is a data structure that holds the data to be sent or received. For some local operations, such as copy/reduce we provide some functions to manipulate the chunk.
+
+```python
+rank = Rank(rank_id)
+input_buffer = rank.get_input_buffer()
```
caiomcbr, line 7: How would be the interface to retrieve the scratch buffer?

docs/design/mscclpp-dsl-next.md
```diff
@@ -0,0 +1,175 @@
...
+dst_chunk = input_buffer[0:1]
+scr_chunk = intput_buffer[1:2]
+rank.copy(dst_chunk, src_chunk, tb=0)
+rank.reduce(dst_chunk, src_chunk, op="sum", tb=0)
+```
+
+### Channel
+All cross ranks related op is done through channels. A channel is a communication medium between ranks. It can be a network socket, shared memory, or any other form of IPC. The channel is responsible for sending and receiving messages between ranks.
+
+**Note:** Each time call channel will created a new one. If user want to reuse the channel, please keep the channel object.
```
caiomcbr, line 17: Why did you decide to change to this approach instead of using the tag to determine if we will create more channels or not? I mean, It seems easier to code using the tag for general purposes, rather than worrying about maintaining the object.

docs/design/mscclpp-dsl-next.md
```diff
@@ -0,0 +1,175 @@
...
+
+## For Pipeline Loop
+For some cases, we need to pipeline the kernel to overlap some operations. For example, the first stage is copy data from input buffer to scratch buffer, the second stage is transfer data from scratch buffer to other peers. We could use `Rank.semphore` to synchronize the two stages. 
+```python
+sem = Rank.Semaphore(rank=0, size=1)
+rank = Rank(0)
+rank.copy(dst_chunk, src_chunk, tb=0)
+sem.release(tb=0)
+channel = Channel(dst_rank, src_rank, channel_type)
+sme.acquire(tb=1)
```
caiomcbr, line 114: Typo: sme -> sem

docs/design/mscclpp-dsl-next.md
```diff
@@ -0,0 +1,175 @@
...
+channel.put(dst_chunk, src_chunk, tb=1)
+```
+
+Also we could provide some gramar sugar to make the pipeline more readable. For example, we could use `Loop` to construct the pipeline. 
+```python
+sem = Rank.Semaphore(rank=rank, size=1)
+rank = Rank(src_rank)
+with Loop.iteration(unit=2**20, num_chunks=1) as iter:
+    # the dst_chunk and src_chunk size but same as loop context
```
caiomcbr, line 123: I understand your point, but perhaps we could refine this comment to: The dst_chunk and src_chunk sizes should match the num_chunks parameter in the loop context.

docs/design/mscclpp-dsl-next.md
```diff
@@ -0,0 +1,175 @@
...
+## For Pipeline Loop
+For some cases, we need to pipeline the kernel to overlap some operations. For example, the first stage is copy data from input buffer to scratch buffer, the second stage is transfer data from scratch buffer to other peers. We could use `Rank.semphore` to synchronize the two stages. 
+```python
+sem = Rank.Semaphore(rank=0, size=1)
+rank = Rank(0)
+rank.copy(dst_chunk, src_chunk, tb=0)
+sem.release(tb=0)
+channel = Channel(dst_rank, src_rank, channel_type)
+sme.acquire(tb=1)
+channel.put(dst_chunk, src_chunk, tb=1)
```
caiomcbr, line 115: You mentioned that this algorithm should copy data from the input buffer to the scratch buffer locally, meaning the src_chunk is the local input buffer and the dst_chunk is the local scratch buffer. The second stage involves copying from the scratch buffer to another peer. Therefore, should it be something like this: channel.put(other_peer_chunk, dst_chunk, tb=1), no?

docs/design/mscclpp-dsl-next.md
```diff
@@ -0,0 +1,175 @@
...
+Also we could provide some gramar sugar to make the pipeline more readable. For example, we could use `Loop` to construct the pipeline. 
+```python
+sem = Rank.Semaphore(rank=rank, size=1)
+rank = Rank(src_rank)
+with Loop.iteration(unit=2**20, num_chunks=1) as iter:
+    # the dst_chunk and src_chunk size but same as loop context
+    rank.copy(dst_chunk, src_chunk, tb=0, iter_context=iter)
+    sem.release(tb=0)
+    channel = Channel(dst_rank, src_rank, channel_type)
+    sme.acquire(tb=1)
```
caiomcbr, line 127: Typo: sme -> sem

docs/design/mscclpp-dsl-next.md
```diff
@@ -0,0 +1,175 @@
...
+    channel.put(dst_chunk, src_chunk, tb=1, iter_context=iter)
```
caiomcbr, line 128: Same thing here, regarding the dst_chunk being the local scratch buffer and the src_chunk being the local input buffer. I'm not sure if your intention was to create something more like pseudocode and refer to the first parameter as the destination chunk and the second as the source chunk; it's understandable.

docs/design/mscclpp-dsl-next.md
```diff
@@ -0,0 +1,175 @@
...
+            src_chunk = input_buffer[offset:offset+1]
+            rank.copy(dst_chunk, src_chunk, tb=0, iter_context=iter)
+        chan.signal(tb=0, sync="before")
+        chan.wait(tb=0, sync="after")
+        sem0.release(tb=0)
+
+        # do allreduce in scratch buffer
+        sem0.acquire(tb=1, sync="after")
+        nvls_chan.group_load_reduce(offset, size=1, op="sum", tb=0, iter_context=iter)
+        nvls_chan.group_store(offset, size=1, tb=0, iter_context=iter)
```
caiomcbr, line 161: Shouldn't this two NVLs operation be using tb=1?

docs/design/mscclpp-dsl-next.md
Binyang2014: You are right

docs/design/mscclpp-dsl-next.md
Binyang2014: Fixed

docs/design/mscclpp-dsl-json-file.md
```diff
@@ -0,0 +1,478 @@
...
+  {
+    "name": "put",
+    "o_buff": [
+      {
+      "srcbuff": "o",
+      "srcoff": 0,
+      "dstbuff": "o",
+      "dstoff": 0,
+      "cnt": 1
+      }
```
Binyang2014, line 241: How to make this buffer with buffer collections

docs/design/mscclpp-dsl-json-file.md
```diff
@@ -0,0 +1,478 @@
...
+              "name": "put",
+              "o_buff": [
+                {
+                "srcbuff": "o",
+                "srcoff": 1,
+                "dstbuff": "o",
+                "dstoff": 1,
+                "cnt": 1
+                }
+              ],
```
Binyang2014, line 419: Same here. We can get the channel via channel id, but how to find the associate buffer. Need to refer buffer collection in the put operation?

docs/design/mscclpp-dsl-json-file.md
```diff
@@ -0,0 +1,478 @@
...
+```
+
+## GPU
+The gpus field is the core of the JSON file, containing the detailed configuration of the collective algorithm for each GPU. It is defined as a list, where each element describes the setup for a specific GPU. Each GPU entry includes the following fields:
+
+- ```id```: The identifier of the GPU.
+- ```inputChunks```: The number of chunks in the input buffer.
+- ```outputChunks```: The number of chunks in the output buffer.
+- ```scratchChunks```: The number of chunks in the scratch buffer.
+- ```chunkGroups```: The number of chunk groups used in the algorithm.
```
Binyang2014, line 41: We may need to rethink about `chunkGroups` field

docs/design/mscclpp-dsl-json-file.md
```diff
@@ -0,0 +1,478 @@
...
+        ]
+      }
+    ]
+  }
+]
+```
+
+### Buffer Collection
+The buffer collection field will contain the description of all the remote buffers that his GPU need to access, it will contain the following fields:
+
```
Binyang2014, line 125: Maybe change to a better name? It's for remote buffer, buffer collection is a bit confused

docs/design/mscclpp-dsl-json-file.md
```diff
@@ -0,0 +1,478 @@
...
+-
+
+Example
+
+```json
+"bufferCollection": [
+  {
+    "rank": 1,
+    "buff": "i"
+  }
```
Binyang2014, line 137: Do we need some info for this buffer is for intra node or inter node. Or the buffer can be bond to memory channel or proxy channel. For proxy channel we need to call `ProxyService::addMemory(RegisteredMemory memory)` to register the buffer. for memory channel we need to keep buffer remote addr, it's different

docs/design/mscclpp-dsl-json-file.md
```diff
@@ -0,0 +1,478 @@
...
+    ]
+  }
+]
+```
+
+The most important field here is the ops fields, where there is the description for all the operations, each operation has specific fields, let's given an overview in all the possible fields and after this go throught each of them:
+
+- ```name```: Operation name, it include: signal, wait, put, get.
+- ```ctype```: Channel type, it include: memory, port, switch.
+- ```i_cids```: The channel id of the channels used in the receive operations.
```
Binyang2014, line 177: Since we decouple channel and buffer, maybe we can just use cids for both i_cids and o_cids

docs/design/mscclpp-dsl-json-file.md
```diff
@@ -0,0 +1,478 @@
...
+              "i_cids": [
+                {
+                  "id": 0
+                }
+              ],
+              "ctype": "memory",
+            },
+            {
+              "name": "put",
+              "o_buff": [
```
Binyang2014, line 316: Can we make it more clear. For example put operation has two buffers, input buffer and output buffer.
it could be 
```json
{
  "name": "put",
   "src_buff": {
      "refer_id": -1, // -1 for local
      "offset": 0,
      "type": "i"
    },
   "dst_buff": {
     "refer_id": 1,
      "offset": 0
    },
   "cids": [{"id": 0}],
   "ctype": "memory"
}
```

caiomcbr: Is there any situation where the src_buffer refer_id field isn't -1? I mean, maybe we could remove the refer_id for the src_buff.

docs/design/mscclpp-dsl-json-file.md
```diff
@@ -0,0 +1,561 @@
...
+```
+
+### Put Operation
+The put operation is composed by the field ```name```, ```src_buff```, ```dst_buff```, ```cids```, ```ctype```.
+
+```json
+"ops": [
+  {
+    "name": "put",
+    "src_buff": [
```
Binyang2014, line 232: As you mentioned before, how to deal with readReduceCopySend. It has remote_src_buff, local_dst_buff, remote_dst_buff. Also may need to move cid into buffer block, since we need to use this channel to send data. How about this:
```json
 {
              "name": "put",
              "local_src_buff": [
                {
                  "offset": 0,
                  "type": "i",
                }
              ],
              "remote_dst_buff": [
                {
                  "buff_id": 0,
                  "offset": 0,
                  "cid": 0,
                }
              ],
             "size": 1,
}
```

docs/design/mscclpp-dsl-next.md
```diff
@@ -0,0 +1,175 @@
...
+        sem1.acquire(tb=2)
+        chan1.wait(tb=2, sync="after")
+        for index in range(nranks):
+            dst_chunk = output_buffer[index:index+1]
+            src_chunk = scratch_buffer[index:index+1]
+            rank.copy(dst_chunk, src_chunk, tb=2, iter_context=iter)
+```
+
+## All2All support
+For now, DSL only support static all2all algorithm. For all2allv support, we need to get the send/recv size at the runtime. It may require some placeholder at the Json execution plan and relace to the real size at the runtime. If we could make chunk size be variable, we could use the same way to support all2allv.
```
mahdiehghazim, line 175: --> "Currently, the DSL only supports the static all2all algorithm. To support all2allv, we need to obtain the send/receive sizes at runtime. This may require using placeholders in the JSON execution plan, which would be replaced with the actual sizes during execution. If we can make the chunk size variable, the same approach could be used to support all2allv."

docs/design/mscclpp-dsl-json-file.md
```diff
@@ -0,0 +1,561 @@
...
+- ```num_threads_per_block```: Specifies the number of threads per thread block. Typical values include 256, 512, 768, or 1024, depending on performance requirements.
+- ```use_double_scratch_buffer```: A boolean flag that enables to double the scratch buffer. When set to true, the scratch buffer double the size and is logically divided into two halves. This allows alternating executions of the same algorithm to use different halves of the buffer, for instance, odd-numbered executions use the first half, and even-numbered executions use the second. This strategy helps avoid memory conflicts when the same algorithm is invoked multiple times.
+- ```min_message_size```: The minimum message size supported by this algorithm. If the message is smaller than this value, the algorithm will not be selected for execution.
+- ```max_message_size```: The maximum message size supported by this algorithm. If the message exceeds this value, the algorithm will not be selected for execution.
+
+Example:
+
+```json
+{
+  "name": "reducescatter",
```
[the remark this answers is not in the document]
caiomcbr: Yes, the main function of the name is to have a different setup for each algorithm that uses the same collective. The algorithm can perform the same task even with the same concept, but if they have different names, they will have distinct setups (channels, buffers, etc.). Anyway, it's better to change as you mentioned to avoid confusion between the algorithm name and the collective.

docs/design/mscclpp-dsl-json-file.md
```diff
@@ -0,0 +1,561 @@
...
+    "cids": [
+      0
+    ],
+    "ctype": "memory",
+  }
+]
+```
+
+### Wait Operation
+The signal operation is composed by the field ```name```, ```cids```, ```ctype```.
```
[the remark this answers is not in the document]
caiomcbr: Yes, you are right, it should be wait operation.

docs/design/mscclpp-dsl-json-file.md
```diff
@@ -0,0 +1,561 @@
...
+    ],
+    "cids": [
+      0
+    ],
+    "ctype": "memory"
+  }
+]
+```
+
+### The json field for device semaphore
```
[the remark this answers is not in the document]
caiomcbr: Yes, Binyang added this part, I'll added some description.

docs/design/mscclpp-dsl-next.md
```diff
@@ -0,0 +1,175 @@
...
+
+Here is the example for two ranks synchonization with each others.
+```python
+nranks = 2
+for i in range(nranks):
+    src_rank = i
+    dst_rank = (i + 1) % nranks
+    channel = Channel(dst_rank, src_rank, channel_type=Channel.memory)
+    channel.relaxedSignal(tb=0, sync=None)
+    channel.relaxedWait(tb=0, sync="after")
```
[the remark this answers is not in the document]
caiomcbr: The relaxed signal/wait is a light fence that does not involve data synchronization. In the example, I use this light fence because I only want to check if the other peers are ready to start exchanging data, and not to ensure that some data has arrived or something similar.

docs/design/mscclpp-dsl.md
```diff
@@ -1,114 +1,179 @@
...
-- `duplicated`: Each chunk is split into smaller parts based on the number of instances, duplicating the same instructions for all parts. For example, ChunkA is split into ChunkA0 and ChunkA1, while ChunkB is split into ChunkB0 and ChunkB1. Both ChunkA0 and ChunkA1 belong to Instance 0, and both ChunkB0 and ChunkB1 belong to Instance 1.
-- `interleaved`: Assign chunks to instances in an interleaved manner. For example, ChunkA and ChunkB are split into to ChunkA0, ChunkA1, ChunkB0, and ChunkB1. ChunkA0 and ChunkB0 belong to Instance 0, while ChunkA1 and ChunkB1 belong to Instance 1.
+This example demonstrates a pipelined loop that copies data from an input buffer to a scratch buffer, then transfers it to other peers. The `Semaphore` is used to synchronize the two stages. `unit` specifies the size of each chunk, and `num_chunks` indicates how many chunks will be processed in the loop.
+
+```python
+sem = Semaphore(rank=rank, initial_value=1)
```
caiocbr: Shouldn't the initial value be set to 0 here?

docs/design/mscclpp-dsl.md
```diff
@@ -1,114 +1,179 @@
...
+    rank.copy(dst_chunk, src_chunk, tb=0)
+    sem.release(tb=0)
+    sem.acquire(tb=1)
+    channel.put(other_peer_chunk, dst_chunk, tb=1)
+``` 
+
+
+Here is the example for two ranks allreduce. Which achieve non-zero copy and use nvls. We use 3 thread-blocks to do the allreduce.
+The first thread-block is used to copy data from input buffer to scratch buffer, the second thread-block is used to do allreduce in scratch buffer, and the third thread-block is used to copy data from scratch buffer to output buffer.  The thread-blocks are synchronized by semaphores.
+```python
```
caiomcbr, line 110: Update example. Consider using this https://github.com/microsoft/mscclpp/blob/feature/dsl/python/mscclpp/language/tests/single_node/allreduce_pipeline.py as a reference.

docs/design/mscclpp-dsl.md
```diff
@@ -1,114 +1,179 @@
...
+Example for two ranks allreduce via switch channel.
 ```python
-c = chunk(rank, Buffer.input, index, size)
+# Creating Channels
+switch_chan = SwitchChannel(rank_list=[gpu for gpu in range(gpu_size)], buffer_type=BufferType.input)
+for gpu in range(gpu_size):
+    buffer_offset = gpu
+    rank = Rank(gpu)
+    input_buffer = rank.get_input_buffer()
+    switch_chan.at_rank(gpu).group_load_reduce(buffer_offset, size=1, dst_chunk=input_buffer[gpu : gpu + 1], tb=0)
```
caiocbr: Update group_load_reduce -> reduce

docs/design/mscclpp-dsl.md
```diff
@@ -1,114 +1,179 @@
...
+    switch_chan.at_rank(gpu).group_store(input_buffer[gpu : gpu + 1], buffer_offset, size=1, tb=0)
```
caiocbr: Update group_store -> broadcast

caiomcbr (approved):
LGTM

PR #521, opened 2025-05-10
Fix #514

chhwang:
* In cases when the same `tag` is used for receiving data from the same remote rank, #514 changed the behavior of `Communicator::connect` and `Communicator::recvMemory` to receive data in the order of `std::shared_future::get()` is called, instead of the original behvaior that receive data in the order of the method calls. Since the original behavior is more intuitive, we get that back. Now when `get()` is called on a future, the async function will first call `wait()` on the latest previously returned future. In a recursive manner, this will call `wait()` on all previous futures that are not yet ready.
* Removed all deprecated API calls and replaced into the new ones.

### DIFF
--- a/src/communicator.cc
+++ b/src/communicator.cc
@@ -35,6 +35,9 @@ MSCCLPP_API_CPP void Communicator::sendMemory(RegisteredMemory memory, int remot
 }
 
 MSCCLPP_API_CPP std::shared_future<RegisteredMemory> Communicator::recvMemory(int remoteRank, int tag) {
+  if (pimpl_->lastRecvItem_ && pimpl_->lastRecvItem_->isReady()) {
+    pimpl_->lastRecvItem_.reset();
+  }
   auto future = std::async(std::launch::deferred, [this, remoteRank, tag, lastRecvItem = pimpl_->lastRecvItem_]() {
     if (lastRecvItem) {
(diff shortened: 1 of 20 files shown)


src/communicator.cc
```diff
@@ -31,30 +31,50 @@ MSCCLPP_API_CPP RegisteredMemory Communicator::registerMemory(void* ptr, size_t
 }
 
 MSCCLPP_API_CPP void Communicator::sendMemory(RegisteredMemory memory, int remoteRank, int tag) {
-  pimpl_->bootstrap_->send(memory.serialize(), remoteRank, tag);
+  bootstrap()->send(memory.serialize(), remoteRank, tag);
 }
 
 MSCCLPP_API_CPP std::shared_future<RegisteredMemory> Communicator::recvMemory(int remoteRank, int tag) {
-  return std::async(std::launch::deferred, [this, remoteRank, tag]() {
+  if (pimpl_->lastRecvItem_ && pimpl_->lastRecvItem_->isReady()) {
```
Binyang2014, line 38: With this change, the recv order will be same as the send order. But for the message comes from different remoteRank, we can recv the message out of the send order. Maybe we can use different lastRecvItem per rank?

chhwang: Thanks for pointing out

PR #529, opened 2025-05-21
Rename `ChannelTrigger` fields and check field values in debug builds

chhwang:


src/port_channel.cc
```diff
@@ -67,7 +67,7 @@ MSCCLPP_API_CPP void ProxyService::bindThread() {
 
 ProxyHandlerResult ProxyService::handleTrigger(ProxyTrigger triggerRaw) {
   ChannelTrigger* trigger = reinterpret_cast<ChannelTrigger*>(&triggerRaw);
-  std::shared_ptr<Host2DeviceSemaphore> semaphore = semaphores_[trigger->fields.chanId];
+  std::shared_ptr<Host2DeviceSemaphore> semaphore = semaphores_[trigger->fields.semaphoreId];
```
Binyang2014, line 70: A bit confused about this field. Here if we want to put data, we need get semaphore object, then use the connection from semaphore to send data.
Maybe use channel is easier to understand. Get use the connection from channel to send data.

PR #532, opened 2025-05-24
Add `Connection::atomicAdd` interface

chhwang:


src/connection_kernels.cu
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the MIT license.
+
+#include <mscclpp/atomic_device.hpp>
+
+#include "connection_kernels.hpp"
+
+namespace mscclpp {
+
+__global__ void connectionAtomicAddKernel(uint64_t* dst, uint64_t value) {
+  atomicFetchAdd(dst, value, memoryOrderRelease);
```
Binyang2014: Why do we need memoryOrderRelease here. BTW, can we use this mechanism for flag?

chhwang:
Moved to #784

PR #534, opened 2025-05-25
Fix memory leak bug

### LINKED ISSUE #533: [Bug] RegisteredMemory not properly destroyed
Hi, the following code causes GPU OOM on hopper with nvls enabled. I am using the latest main branch.
```python
from mscclpp import Transport, TcpBootstrap, Communicator
from mscclpp._mscclpp import Context, RawGpuBuffer
import cupy as cp
cp.cuda.Device(0).use()
bootstrap = TcpBootstrap.create(0, 1)
bootstrap.initialize(bootstrap.create_unique_id(), 60)
comm = Communicator(bootstrap)
for i in range(100):
    if i % 10 == 0:
        print(f"{i=}", flush=True)
    mem = RawGpuBuffer(2 ** 30)
    reg = comm.register_memory(mem.data(), mem.bytes(), Transport.CudaIpc)
    del reg, mem
```
Output:
```bash
i=0
i=10
i=20
i=30
i=40
i=50
i=60
i=70
Traceback (most recent call last):
  File "<stdin>", line 4, in <module>
mscclpp._mscclpp.CuError: (2, 'Call to result failed./.../mscclpp/src/gpu_utils.cc:128 (Cu failure: out of memory)')
```
The code is fine if memory is not registered. Could you please check if it can be reproduced on your side?

Binyang2014:
Fix #533

### DIFF
--- a/src/port_channel.cc
+++ b/src/port_channel.cc
@@ -20,7 +20,8 @@ MSCCLPP_API_CPP PortChannel::PortChannel(SemaphoreId semaphoreId, std::shared_pt
 
 MSCCLPP_API_CPP ProxyService::ProxyService(size_t fifoSize)
     : proxy_(std::make_shared<Proxy>([&](ProxyTrigger triggerRaw) { return handleTrigger(triggerRaw); },
-                                     [&]() { bindThread(); }, fifoSize)) {
+                                     [&]() { bindThread(); }, fifoSize)),
+      lock_(false) {
   int cudaDevice;
   MSCCLPP_CUDATHROW(cudaGetDevice(&cudaDevice));
   deviceNumaNode_ = getDeviceNumaNode(cudaDevice);


liangyuRain:
Also, I wonder if the MR registered in IB should also be properly unregistered in RegisteredMemory's destructor?
https://github.com/microsoft/mscclpp/blob/7263bb7131bc3b39b79b3bfd908d27f58a7c7a28/src/registered_memory.cc#L90

Binyang2014:
> Also, I wonder if the MR registered in IB should also be properly unregistered in RegisteredMemory's destructor?
> 
> https://github.com/microsoft/mscclpp/blob/7263bb7131bc3b39b79b3bfd908d27f58a7c7a28/src/registered_memory.cc#L90

For MR and QP, thinking if we could wrap it into `shared_ptr`.

src/port_channel.cc
```diff
@@ -38,10 +38,26 @@ MSCCLPP_API_CPP SemaphoreId ProxyService::addSemaphore(std::shared_ptr<Host2Devi
...
   memories_.push_back(memory);
   return memories_.size() - 1;
 }
 
+MSCCLPP_API_CPP void ProxyService::removeMemory(MemoryId memoryId) {
+  if (reusableMemoryIds_.find(memoryId) != reusableMemoryIds_.end() || memoryId >= memories_.size()) {
+    WARN("Attempted to remove a memory that is not registered or already removed: %u", memoryId);
+    return;
+  }
+  memories_[memoryId] = RegisteredMemory();
```
chhwang, line 62: We need a resource lock so that the removal doesn't happen while there exists unflushed triggers on the flight in the proxy. Also, we need a mechanism that prevents proxies on remote ranks from processing requests on this RegisteredMemory.

Binyang2014: I think users are responsible for safely releasing the buffer. Before removing it, ensure that no peers or device-side operations are still accessing this memory. Adding spinlock here to prevent race conditions inside the proxy_channel.

PR #544, opened 2025-06-04
Do not use tail replica by default

chhwang:


src/fifo.cc
```diff
@@ -41,27 +37,36 @@ MSCCLPP_API_CPP Fifo::~Fifo() = default;
...
   trigger.snd = ptr->snd;
   return trigger;
 }
 
 MSCCLPP_API_CPP void Fifo::pop() {
-  atomicStore(&(pimpl->triggers.get()[pimpl->hostTail % pimpl->size].fst), uint64_t{0}, memoryOrderRelease);
-  (pimpl->hostTail)++;
+  uint64_t curTail = *(pimpl->hostTail);
+  atomicStore(&(pimpl->triggers.get()[curTail % pimpl->size].fst), uint64_t{0}, memoryOrderRelease);
+  *(pimpl->hostTail) = curTail + 1;
```
Binyang2014, line 50: Do we need make `*(pimpl->hostTail) = curTail + 1;` atomic?

Binyang2014 (approved):

PR #547, opened 2025-06-06
Support Fuse Operation on MSCCPP DSL

caiomcbr:
Fuse Operations

Rank:

- Sync + Sync = Sync

Memory Channel:

- Signal + Signal = Signal (Operations should use different channels)
- Wait + Wait = Wait (Operations should use different channels)
- Relax Signal + Relax Signal = Relax Signal (Operations should use different channels)
- Relax Wait + Relax Wait = Relax Wait (Operations should use different channels)
- Get + Get = Get (Operations size should match)
- Put + Put = Put (Operations size should match)
- Put Packet + Put Packet = Put Packet (Operations size should match)
- Reduce + Reduce = Reduce (Operations should have the same local_src_buff, local_dst_buff and reduce_operation)
- Reduce Packet + Reduce Packet = Reduce Packet (Operations should have the same local_src_buff, local_dst_buff and reduce_operation)
- Read Reduce + Read Reduce = Read Reduce (Operations should have the same local_src_buff, local_dst_buff and reduce_operation)

- Reduce + Put = Reduce Send (Operations should match the reduce dst_buff and the put src_buff)
- Reduce Send + Put = Reduce Send (Operations should match the reduce dst_buff and the put src_buff)
- Read Reduce + Put = Read Reduce Send (Operations should match the reduce dst_buff and the put src_buff)
- Read Reduce Send + Put = Read Reduce Send (Operations should match the reduce dst_buff and the put src_buff)
- Reduce Packet + Put Packet = Reduce Send Packet (Operations should match the reduce dst_buff and the put src_buff)
- Reduce Send Packet + Put Packet = Reduce Send Packet (Operations should match the reduce dst_buff and the put src_buff)

Port Channel:

- Signal + Signal = Signal (the operations should use different channels)
- Wait + Wait = Wait (the operations should use different channels)
- Flush + Flush = Flush
- Put + Put = Put (Operations size should match)
- Put With Signal + Put With Signal = Put With Signal (Operations size should match)
- Put With Signal and Flush + Put With Signal and Flush = Put With Signal and Flush (Operations size should match)


python/mscclpp/language/general.py
```diff
@@ -2,4 +2,5 @@
 
 
 def JSON():
+    get_program().optimize_operations()
```
Binyang2014: I remember we have an option to disable op fusion. Do we keep this option?

python/mscclpp/language/internal/operations.py
```diff
@@ -77,12 +87,30 @@ def __init__(self, channels_ids: List[int], channel_type: ChannelType, relaxed:
...
+            and self.channel_type == other.channel_type
+            and self.name == other.name
+            and not self.channel_ids & other.channel_ids
+        ):
+            fused_operation = SignalOperation(
+                channels_ids=self.channel_ids | other.channel_ids,
+                channel_type=self.channel_type,
+                relaxed=(self.name == Instruction.relaxed_signal),
+            )
+        if isinstance(other, SyncOperation):
```
Binyang2014: Why return self if merge with sync operation?

python/mscclpp/language/internal/operations.py
```diff
@@ -217,6 +298,26 @@ def __init__(
         self.channel_ids = channel_ids
         self.channel_type = channel_type
 
+    def __add__(self, other):
+        fused_operation = None
+        if (
+            isinstance(other, PutOperation)
+            and self.name == Instruction.put or self.name == Instruction.put_with_signal or self.name == Instruction.put_with_signal_and_flush
```
Binyang2014: According to this, can we fuse `proxyChannel.put()` and `proxyChannel.putWithSignal()` ?

python/mscclpp/language/internal/optmizer.py
```diff
@@ -0,0 +1,26 @@
...
+    fused_operations = []
+    while operation_index < len(operations):
+        next_operation_index = operation_index + 1
+        current_operation = operations[operation_index]
+        previous_operation = None
+        fused_operation = current_operation
+        
+        while next_operation_index < len(operations):
+            next_operation = operations[next_operation_index]
+            fused_operation = current_operation + next_operation
```
Binyang2014, line 15: We have some fusion rules such as reduce + put = reduce-send. Do we still have this?

python/mscclpp/language/general.py
caiocbr: I didn't keep it. But I guess that is a good idea give the operation to disable it.

python/mscclpp/language/internal/operations.py
caiocbr: Imagine that I have something like:
signal(tb=0, sync=SyncType.after)
wait(tb=0, sync=SyncType.before)

We will generate something like: signal nop nop wait. We don't need these 2 nop operations. If you check the script that fuses instructions, it will try to fuse until it can't. It stores the last operation, and if the last operation is NOP, it will process it again. For instance:
1. Current op sequence: signal nop nop wait -> It will try to fuse signal and nop, it will work (not return None), so it will process to the next step
2. Current op sequence: signal nop wait -> It will try to fuse signal and nop, it will work (not return None), so it will process to the next step
3. Current op sequence: signal wait -> It will try to fuse signal and wait, it will fail (return None), so it will check the previous merged operation, which is a NOP operation, and reprocess that operation
4. Current op sequence: signal nop wait. It will start from the nop

python/mscclpp/language/internal/operations.py
caiocbr: No. it needs to be the same operation, so it could be two put operation or two put_with_signal or two put_with_signal_and_flush

python/mscclpp/language/internal/optmizer.py
caiocbr: Yes, if you check the reduce operation in the operations file, it will have a method to add the reduce with put operation creating the reduce-send

python/mscclpp/language/tests/allreduce.py
```diff
@@ -0,0 +1,106 @@
...
+def allreduce_example(name, gpu_size, num_threads_per_block, min_message_size, max_message_size):
+    chunksperloop = gpu_size
+    collective = AllReduce(gpu_size, chunksperloop, True)
+    with MSCCLPPProgram(
+        name,
+        collective,
+        gpu_size,
+        protocol="LL",
+        num_threads_per_block=num_threads_per_block,
+        use_double_scratch_buffer=False,
```
Binyang2014: The field `use_double_scratch_buffer` still a bit confusing, need to think about the meaning of this field. Seems it should be true here

python/mscclpp/language/internal/types.py
```diff
@@ -100,31 +100,34 @@ def to_json(self):
 class RemoteBuffer:
     __remote_buffer_count = defaultdict(int)
 
-    def __init__(self, rank: int, type: BufferType, channel_access: ChannelType, set_id: bool = False):
+    def __init__(
+        self, local_rank: int, remote_rank: int, type: BufferType, channel_access: ChannelType, set_id: bool = False
+    ):
         if set_id:
```
Binyang2014: Why we not set id automaticly constructor but provide another API

python/mscclpp/language/internal/operations.py
Binyang2014: I got your idea, but a bit strange for me to fuse signal and nop. Here we fuse to <signal, wait>, then back to <signal, nop, wait>. Can we follow the normal fuse process, signal + nop can not be fused -> break -> go to nop + nop -> fuse to nop -> go to nop + wait -> can not fuse, break, go to wait....

python/mscclpp/language/tests/unit_tests/group_load_reduce_test.py
```diff
@@ -27,6 +27,9 @@ def group_load_reduce_test(num_threads_per_block, min_message_size, max_message_
         ch = SwitchChannel(rank_list=[0, 1], buffer_type=BufferType.input)
         ch.group_load_reduce(buffer_offset=0, size=1, tb=0, dst_chunk=dst_chunk[0:1])
```
Binyang2014: Need to update this?

python/mscclpp/language/tests/unit_tests/flush/flush_fuse_test.py
```diff
@@ -0,0 +1,41 @@
...
+def flush_test(num_threads_per_block, min_message_size, max_message_size):
+    gpus = 3
+    collective = TestCollective(gpus, 0, 0)
+    with MSCCLPPProgram(
+        "flush_test",
+        collective,
+        gpus,
+        protocol="Simple",
+        num_threads_per_block=num_threads_per_block,
+        use_double_scratch_buffer=False,
```
Binyang2014, line 21: Do we have default value for `use_double_scratch_buffer`?

python/mscclpp/language/internal/operations.py
caiocbr: I understood you point, but I did this because when I encounter something like: signal nop nop signal. I need to merge the two signal operations. Therefore, I need a mechanism to ignore the nop operations and identify the next operation that is not a nop.

python/mscclpp/language/tests/allreduce.py
caiocbr: This field indicates that we will double the scratch buffer to prevent conflicts when running this algorithm multiple times consecutively.

python/mscclpp/language/internal/types.py
caiocbr: The context where the RemoteBuffer is used is as follows:
1. The user calls an operation, which refers to a remote buffer.
2. I create this remote buffer object and try to insert it into the GPU.
3. If this object was previously inserted into the GPU, I already have the ID; if not, I set the ID.

This is why I created a different interface for it. I'm not sure if this is the best solution; maybe I could manage the ID outside the remote buffer. What do you think is better?

python/mscclpp/language/tests/unit_tests/flush/flush_fuse_test.py
caiocbr: Yes

python/mscclpp/language/channel.py
```diff
@@ -7,65 +7,33 @@
 
 @dataclass
 class Channel:
```
Binyang2014: Change to memoryChannel, if we have separate class for portChannel

python/mscclpp/language/tests/unit_tests/signal_wait/relaxed_signal_wait_test.py
```diff
@@ -25,7 +25,7 @@ def relaxed_signal_wait_test(num_threads_per_block, min_message_size, max_messag
         for src_rank in range(gpus):
             for dst_rank in range(gpus):
                 if src_rank != dst_rank:
-                    ch = Channel(dst_rank, src_rank, ChannelType.memory)
+                    ch = Channel(dst_rank, src_rank)
```
Binyang2014: Maybe create  MemoryChannel here directly

PR #549, opened 2025-06-13
Fix a FIFO correctness bug

chhwang:
* Add a FIFO test code that reproduced a correctness issue
* Fix the correctness issue by using pinned memory instead of cudaMemcpy


include/mscclpp/fifo_device.hpp
```diff
@@ -15,98 +15,89 @@
...
+
+    // Set last bit non-zero for safe polling; host will revert.
+    trigger.snd ^= (uint64_t{1} << uint64_t{63});
+
+    // Proceed if tail advanced or target slot is 0.
+    // As tail value is cached on device, the device doesn't need to access host memory every time.
+    uint64_t numInflights = prevHead - *tailCache;
+    if (numInflights >= size / 2) {
+      numInflights = prevHead - (*tailCache = uncachedLoad(tail));
+    }
```
Binyang2014: Do we really need this? I think the cost is almost same: Call load every size/2 elems vs call load every size elems. The POLL_MAYBE_JARIBREAK just a simple while loop and don't have too much overhead.

Tested in H200 and A100, the perf is almost similar.

Binyang2014 (approved):

include/mscclpp/fifo_device.hpp
chhwang: Removed this part

include/mscclpp/fifo_device.hpp
```diff
@@ -15,98 +15,92 @@
...
+    uint64_t* ticketTail = &triggerTicketTails[triggerIdx];
+    uint64_t ticket = atomicFetchAdd<uint64_t, scopeDevice>(ticketHead, 1, memoryOrderRelaxed);
+    POLL_MAYBE_JAILBREAK((atomicLoad<uint64_t, scopeDevice>(ticketTail, memoryOrderRelaxed) != ticket), maxSpinCount);
+
+    // Flip the last bit for safe polling; host will revert.
+    constexpr uint64_t flipMask = uint64_t{1} << uint64_t{63};
+    trigger.snd ^= flipMask;
+
+    // Wait until the trigger is freed by the host.
+    POLL_MAYBE_JAILBREAK((hostLoadRelaxed(&(triggers[triggerIdx].fst)) != 0), maxSpinCount);
```
Binyang2014: We access the host data for every push operation. Will it bring addtional cost?

Binyang2014 (approved):

PR #551, opened 2025-06-17
Ensure Consistency among Data Accesses in the same Thread Block

caiomcbr:
The PR ensures that if two or more operations impact the same memory location, it will insert a mechanism to synchronize the thread block between theses operations, preventing any errors. For instance:

copy chunk 0 -> chunk 1
put chunk 1 -> chunk 2

Here we can see the copy operation and put operation have chunk 1 in common for writing and reading, respectively. In this case, I need to make sure the copy operation ends before starting the put operation. Therefore, we should have a synchronization mechanism between them to avoid any conflicts.

In this PR, we will insert synchronization mechanisms between the operations under the following conditions:

Operation using Write Data Access followed by Operation using Write Data Access to the same memory location
Operation using Read Data Access followed by Operation using Write Data Access to the same memory location
Operation using Write Data Access followed by Operation using Read Data Access to the same memory location


python/mscclpp/language/internal/dsl_types.py
Binyang2014, line 1: So we can not use types as the module name

python/mscclpp/language/internal/buffer_access.py
```diff
@@ -0,0 +1,82 @@
+from sortedcontainers import SortedDict
```
Binyang2014, line 1: We need to put this in our reuqirements.txt?

Binyang2014: Another concern is the latest sortedcontainers release is 2021, seems no updates since then

python/mscclpp/language/internal/dsl_types.py
caiocbr: Yes, I think so, actually I try different approaches, but I couldn't keep the same name.

python/mscclpp/language/internal/buffer_access.py
caiocbr: Interesting. If you want I can try to find another library or code the struct by myself as well.

python/mscclpp/language/channel.py
```diff
@@ -1,4 +1,4 @@
-from mscclpp.language.internal.types import RemoteBuffer, SyncType, ReduceOperationType, Chunk, RankGroup
+from mscclpp.language.internal.dsl_types import RemoteBuffer, SyncType, ReduceOperationType, Chunk, RankGroup
```
Binyang2014: So we still need change package name here?

python/mscclpp/language/channel.py
```diff
@@ -49,7 +49,7 @@ def get(self, dst_chunk: Chunk, src_chunk: Chunk, tb: int):
         tb_channel_ids = get_program().setup_channel(tb, self)
 
         op = GetOperation(
-            src_buff=[RemoteChunk(tb_chunk_id, src_chunk.index, src_chunk.size)],
+            src_buff=[RemoteChunk(src_chunk.buffer, src_chunk.index, src_chunk.size, tb_chunk_id)],
```
Binyang2014, line 52: What's the meaning of tb_chunk_id? Do we need to expose this to user?

Binyang2014 (approved):

python/mscclpp/language/channel.py
caiocbr: This refers to the remote buffer ID. When a user performs an operation involving remote memory access, I need this ID to create the necessary operations. Essentially, I obtain the ID and use it to set up the operations. There is no need to make this visible to the user.

PR #557, opened 2025-06-21
FIFO improvements

chhwang:
* Revert `MSCCLPP_FIFO_USE_TAIL_REPLICA=1` back to the default.
* Optimize `FifoDeviceHandle`.
* Do not use `cudaHostAllocWriteCombined` that increases latency.
* Pin host memory for `Host2DeviceSemaphore::outboundSemaphore_`.
* Fix proxy NUMA binding issues.
* Prevent graph capture inside proxy threads.
* Now `CudaIpcConnection` skips stream sync when unnecessary.
* Now any type of connection needs to hold a shared pointer to the context for memory safety.
* Now a context should be always managed by a shared pointer for memory safety.
* Minor docs & interface improvements.
* Minor fix in `mscclpp-test` correctness test.


src/fifo.cc
```diff
@@ -13,71 +14,75 @@ namespace mscclpp {
...
-  // (when `env()->fifoUseTailReplica == true`), it always holds that *tailReplica <= *hostTail.
-  std::shared_ptr<uint64_t> hostTail;
-
   // for transferring fifo tail
   CudaStreamWithFlags stream;
 
   Impl(int size)
       : triggers(detail::gpuCallocHostUnique<ProxyTrigger>(size)),
         head(detail::gpuCallocUnique<uint64_t>()),
+        tailHost(env()->fifoUseTailReplica ? std::make_shared<uint64_t>(0) : detail::gpuCallocHostShared<uint64_t>()),
```
Binyang2014, line 27: Use pin memory to avoid memory swap?

src/proxy.cc
```diff
@@ -12,53 +13,61 @@
 
 namespace mscclpp {
 
-const int ProxyStopCheckPeriod = 1000;
+constexpr int ProxyStopCheckPeriod = 1000;
 
 // Unless explicitly requested, a flush of the tail to device memory is triggered for every ProxyFlushPeriod.
 // As long as the FIFO size is large enough, having a stale tail is not a problem.
-const int ProxyFlushPeriod = 4;
+constexpr int ProxyFlushPeriod = 4;
```
Binyang2014, line 20: Why set proxyFlushPeriod to 4? Is this too small?

chhwang: I just follow existing code. We will see if it needs to be changed later, but I think 4 is fine.

PR #562, opened 2025-07-04
Adding DSL Support for increase Instances

caiomcbr:


python/mscclpp/language/program.py
```diff
@@ -15,6 +15,7 @@ def __init__(
         name: str,
         collective: Collective,
         num_ranks: int,
+        instances: int,
```
Binyang2014: Maybe we can set default value to 1 here

python/mscclpp/language/internal/operations.py
```diff
@@ -21,6 +21,9 @@ class BaseOperation:
     def local_data_access(self, sync_purpose=True):
         return []
 
+    def shift_buffers(self, instance, num_instances):
+        return
```
Binyang2014, line 25: Throw error here?

python/mscclpp/language/internal/gpu.py
```diff
@@ -65,6 +66,49 @@ def resolve_data_dependency(self):
...
+        if ChannelType.memory in self.__channels:
+            self.__channels[ChannelType.memory].connected_to = new_channels[ChannelType.memory]
+        if ChannelType.port in self.__channels:
+            self.__channels[ChannelType.port].connected_to = new_channels[ChannelType.port]
+        self.__nvls_channels = new_channels[ChannelType.switch]
+
+        for threadblock in self.threadblocks:
+            for instance in range(instances):
+                tb = copy.deepcopy(threadblock)
+                tb.id = threadblock.id * instances + instance
```
Binyang2014, line 93: If we have different replica policy, how to modify this? Since the code is sperated in serveral files, it may not easy to make some changes

python/mscclpp/language/internal/operations.py
caiocbr: Actually, we have some operations, such as signal and wait, that do not require data access, so by default, they should perform no action.

python/mscclpp/language/internal/gpu.py
```diff
@@ -65,6 +66,49 @@ def resolve_data_dependency(self):
         for tb in self.threadblocks:
             tb.resolve_data_dependency()
 
+    def replicate_instances(self, instances, replication_function):
+        threadblocks = []
+        channels_base_shift = {
+            ChannelType.memory: (
+                len(self.__channels[ChannelType.memory].connected_to) if ChannelType.memory in self.__channels else 0
```
Binyang2014: Why using double leading underscore here? Not single leading underscore?

Binyang2014 (approved):

PR #566, opened 2025-07-10
Support any GPUs per node for NCCL_API

Binyang2014:
Support any GPUs per node for NCCL API


CMakeLists.txt
```diff
@@ -58,7 +58,9 @@ else()
         set(MSCCLPP_USE_CUDA OFF)
         set(MSCCLPP_USE_ROCM ON)
     else()
-        message(FATAL_ERROR "Neither NVIDIA/CUDA nor AMD/ROCm is found.")
+        message(WARNING "Neither NVIDIA/CUDA nor AMD/ROCm is found, defaulting to NVIDIA/CUDA.")
```
chhwang: Can we make it smarter like if `hip` package is found, use AMD/ROCm.
https://github.com/microsoft/mscclpp/blob/main/cmake/CheckAmdGpu.cmake#L11

Binyang2014: Thanks, fixed in the new commit

PR #567, opened 2025-07-11
Fix pytest failure

Binyang2014:


Binyang2014:
refer: https://github.com/NVIDIA/nccl/blob/7c12c627c62ef4e5a2485777a8d9dce58f3f562f/src/transport/nvls.cc#L113-L115

chhwang (approved):

PR #569, opened 2025-07-16
 Supporting Pipeline Operation at DSL

caiomcbr:
Introduce a mechanism that allows create a pipeline for a sequence of operations, specifying both the number of chunks and the unit size.


python/mscclpp/language/internal/buffer_access.py
```diff
@@ -16,7 +16,12 @@ def __init__(self):
     def process_operations(self, operations):
         result_operations = []
         for operation in operations:
-            if operation.name == Instruction.nop or operation.name == Instruction.barrier:
+            if operation.name == Instruction.pipeline:
+                pipeline_result_operations = self.process_pipeline_operations(operation.operations)
+                if pipeline_result_operations[1]:
```
Binyang2014: What's the meaning of `pipeline_result_operations[1]`?

python/mscclpp/language/internal/types.py
```diff
@@ -205,3 +206,9 @@ class ReplicationPolicy(Enum):
 
     def __str__(self):
         return self.value
+
+
+class FusionStatus(Enum):
+    fused = "fused"
+    vanished = "vanished"
```
Binyang2014: Why need vanished?

PR #571, opened 2025-07-19
Add indirect connection support

hfhongzy:
This PR introduces a new feature called indirect connection, enabling data transfer from endpoint A to endpoint C via forwarding through endpoint B. It defines two abstractions: buffer resources (send and receive buffers on the forwarding endpoint) and scheduler (which manages the data movement logic). We provide an implementation based on Vortex (Yichao, Lin, and Nishil. https://arxiv.org/abs/2502.09541), which uses a toggling double buffer and a pipeline style to overlap the two transfer paths efficiently.

### DIFF
--- a/src/connection.cc
+++ b/src/connection.cc
@@ -421,4 +421,104 @@ void EthernetConnection::recvMessages() {
   }
 }
 
+// IndirectConnection
+
+VortexScheduler::VortexScheduler(std::shared_ptr<Context> context, uint64_t granularity, Device device) : granularity_(granularity), streams_({std::make_shared<CudaStreamWithFlags>(0), std::make_shared<CudaStreamWithFlags>(0)}), forwarding_device_(device) {
+  if (device.type != DeviceType::GPU) {
+    throw std::runtime_error("The forwarding device must be a GPU");
+  }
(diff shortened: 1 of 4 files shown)


Binyang2014:
Thanks @hfhongzy, let me go through the paper first. Since we introduced `VortexScheduler`, thinking if we could have a more general implementation for this kind of requirements

Binyang2014:
Hi @hfhongzy, I have a question for your PR. From your implementation, seems you can create your own connection by inherits from the `Connection` class.  So, do we really need to change MSCCL++ code? I mean if you can implement similar thing in your code repo, such as:
```cpp
namespace your_namespace {
class IndirectConnection  : public mscclpp::Connection {
...
};
};
```
Then create the connection via `auto conn = IndirectConnection(ctx, localEndpoint, scheduler_ptr)`

I think MSCCL++ should support customized p2p connection, trying to better understand your requirements

hfhongzy:
Hi @Binyang2014, thanks for your reply! I believe we (University of Michigan) have been discussing with Changho @chhwang about adding this feature in MSCCL++. Apologies for the confusion here.

PR #572, opened 2025-07-21
Adjust Replication Policy for Ids

caiomcbr:


python/mscclpp/language/internal/gpu.py
```diff
@@ -70,9 +70,7 @@ def resolve_data_dependency(self):
         for tb in self.threadblocks:
             tb.resolve_data_dependency()
 
-    def replicate_instances(
-        self, instances, channel_replication_function, buffer_replication_function, semaphore_replication_function
-    ):
+    def replicate_instances(self, instances, id_replication_function, buffer_replication_function):
```
Binyang2014: The same is strange to me. If we want to unify two functions, maybe just call `channel_and_threadblock_replication_policy` or a shorter name

python/mscclpp/language/internal/gpu.py
```diff
@@ -70,9 +70,7 @@ def resolve_data_dependency(self):
         for tb in self.threadblocks:
             tb.resolve_data_dependency()
 
-    def replicate_instances(
-        self, instances, channel_replication_function, buffer_replication_function, semaphore_replication_function
-    ):
+    def replicate_instances(self, instances, std_replication_function, buffer_replication_function):
```
Binyang2014: std_replication_function mean standard?

caiocbr: Yes, do you prefer use standard directly or another name?

Binyang2014: I prefer `default_replication_function`

Binyang2014 (approved):

PR #574, opened 2025-07-22
Fix multinode test failure

Binyang2014:
Add CPU based connection to fix multi-node test failure issue


Binyang2014:
> The PR description looks too lengthy compared to the changes

Generated by AI... Let me modify the description

PR #575, opened 2025-07-22
Small Fix Attribute Name

caiomcbr:


python/mscclpp/language/tests/unit_tests/sw_broadcast.py
Binyang2014, line 1: Use full name? switch_broadcast.py. Also switch_reduce.py

Binyang2014 (approved):
Pls make sure the allreduce script is working

PR #579, opened 2025-07-24
New DSL implementation

Binyang2014:
The PR contains following changes:
Python side:
- Channel based DSL implementation: decouple channel with chunk.
  - Users create channel explicitly, only need local_rank, remote_rank and channel_type
  - Adjust executor json file, add remote_buffer fields, different op can use different channel and remote buffers combination.
- Reimplement operation fusion, data dependency check mechanism
- Add new op such as semaphore, pipeline 
- Clean code and enhance document
C++ side: 
- Support new execution file json format
- Support semaphore and pipeline operation
- code clean, support non-zero copy scenario


.github/instructions/lisence-add.instructions.md
```diff
@@ -0,0 +1,26 @@
+---
+applyTo: '**/*.cc, **/*.hpp, **/*.py'
```
chhwang, line 2: Add `*.cu`?

chhwang: Add `*.cu`?

python/mscclpp/language/channel.py
```diff
@@ -0,0 +1,872 @@
+from mscclpp.language.internal.types import RemoteBuffer, SyncType, ReduceOperationType, Chunk, RankGroup
```
chhwang, line 4: No license

python/mscclpp/language/channel.py
```diff
@@ -0,0 +1,872 @@
...
+
+        self.channel_id = MemoryChannel._channel_counts[src_rank]
+        MemoryChannel._channel_counts[src_rank] += 1
+
+        self.dst_rank = dst_rank
+        self.src_rank = src_rank
+        self.channel_type = ChannelType.memory
+        get_program().add_channel(self)
+
+    def signal(self, tb: int, data_sync: SyncType = SyncType.none, relaxed=False):
```
chhwang: type hint is missing `relaxed: bool = False`

python/mscclpp/language/channel.py
```diff
@@ -0,0 +1,872 @@
...
+                Defaults to False.
+
+        Example:
+            >>> channel.signal(tb=0, data_sync=SyncType.before)
+        """
+        tb_channel_ids = get_program().setup_channel(tb, self)
+        op = SignalOperation(tb_channel_ids, self.channel_type, data_sync, relaxed)
+        get_program().add_operation(self.src_rank, tb, op)
+
+    def wait(self, tb: int, data_sync: SyncType = SyncType.none, relaxed=False):
```
chhwang: type hint is missing `relaxed: bool = False`

python/mscclpp/language/general.py
```diff
@@ -0,0 +1,17 @@
+from mscclpp.language.internal.globals import get_program
```
chhwang, line 4: No license

python/mscclpp/language/general.py
chhwang, line 1: Why is this file called `general.py`?

python/mscclpp/language/internal/buffer_access.py
```diff
@@ -0,0 +1,92 @@
+from sortedcontainers import SortedDict
```
chhwang, line 4: No license

python/mscclpp/language/rank.py
```diff
@@ -1,33 +1,391 @@
...
+            dst_chunk (Chunk): The destination chunk to copy data to.
+            src_chunk (Chunk): The source chunk to copy data from.
+            tb (int): The thread block ID that will execute this operation.
+
+        Example:
+            >>> rank.copy(dst_chunk, src_chunk, tb=0)
+        """
+        self._copy(dst_chunk=dst_chunk, src_chunk=src_chunk, tb=tb)
+
+    def unpack_copy_packet(self, dst_chunk: Chunk, src_chunk: Chunk, tb: int):
```
Binyang2014: Maybe we can name it as `unpack_packet`

docs/design/mscclpp-dsl.md
chhwang, line 1: Please check typos and make terms consistent (e.g., "switch channel", "switchChannel", "SwitchChannel")

chhwang, line 1: Please check typos and make terms consistent (e.g., "switch channel", "switchChannel", "SwitchChannel")

python/mscclpp/language/channel.py
```diff
@@ -0,0 +1,875 @@
...
+    def signal(self, tb: int, data_sync: SyncType = SyncType.none, relaxed: bool = False):
+        """Send a signal through the memory channel.
+
+        Signals notify the destination that data is ready or an operation has completed.
+        This is used for synchronization between ranks.
+
+        Args:
+            tb (int): The thread block ID that will execute this signal operation.
+            data_sync (SyncType, optional): The type of data synchronization to perform.
+                Defaults to SyncType.none.
```
chhwang: Please make it more concrete (e.g., equivalent to `__syncthreads()`?). Also, it may be safer to make `SyncType.both` as the default.

python/mscclpp/language/channel.py
```diff
@@ -0,0 +1,875 @@
...
+        op = PutOperation(
+            src_buff=[LocalChunk(src_chunk.buffer, src_chunk.index, src_chunk.size)],
+            dst_buff=[RemoteChunk(dst_chunk.buffer, dst_chunk.index, dst_chunk.size, tb_chunk_id)],
+            channel_ids=tb_channel_ids,
+            channel_type=self.channel_type,
+        )
+
+        get_program().add_operation(self.src_rank, tb, op)
+
+    def read_put_packet(self, dst_chunk: Chunk, src_chunk: Chunk, tb: int):
```
chhwang: How about renaming into `copy_packets()`?

python/mscclpp/language/channel.py
```diff
@@ -0,0 +1,875 @@
...
+            dst_buff=[RemoteChunk(dst_chunk.buffer, dst_chunk.index, dst_chunk.size, tb_chunk_id)],
+            channel_ids=tb_channel_ids,
+            channel_type=self.channel_type,
+            from_packet=True,
+            to_packet=True,
+        )
+
+        get_program().add_operation(self.src_rank, tb, op)
+
+    def put_packet(self, dst_chunk: Chunk, src_chunk: Chunk, tb: int):
```
chhwang: Let's name it in plural form `put_packets()`, following the C++ interface name

python/mscclpp/language/channel.py
```diff
@@ -0,0 +1,875 @@
...
+            to_packet=True,
+        )
+
+        get_program().add_operation(self.src_rank, tb, op)
+
+    def reduce(
+        self,
+        local_src_chunk: Chunk,
+        remote_src_chunks: List[Chunk],
+        tb: int,
```
chhwang, line 289: Since all methods need a `tb` parameter, how about making its position consistent (e.g., make it the last parameter for all methods)

python/mscclpp/language/channel.py
```diff
@@ -0,0 +1,875 @@
...
+        Reduces data from multiple remote source chunks with a local source chunk,
+        storing the result in a local destination chunk. If no destination chunk
+        is specified, the result is stored in the local source chunk.
+
+        Args:
+            local_src_chunk (Chunk): The local source chunk on the source rank.
+            remote_src_chunks (List[Chunk]): List of remote source chunks to reduce with.
+            tb (int): The thread block ID that will execute this operation.
+            local_dst_chunk (Chunk, optional): The local destination chunk. If None,
+                uses local_src_chunk as destination. Defaults to None.
```
chhwang, line 304: I think `local_dst_chunk` should be mandatory and `local_src_chunk` should be optional instead. Because we may not have local data to accumulate.

python/mscclpp/language/channel.py
```diff
@@ -0,0 +1,875 @@
...
+            src_buff=[LocalChunk(src_chunk.buffer, src_chunk.index, src_chunk.size)],
+            dst_buff=[RemoteChunk(dst_chunk.buffer, dst_chunk.index, dst_chunk.size, tb_chunk_id)],
+            channel_ids=tb_channel_ids,
+            channel_type=self.channel_type,
+            with_signal=True,
+        )
+
+        get_program().add_operation(self.src_rank, tb, op)
+
+    def put_with_signal_and_flush(self, dst_chunk: Chunk, src_chunk: Chunk, tb: int):
```
chhwang, line 556: How about combining the `put*()` methods into one, like `def put(self, ..., signal: bool = False, flush: bool = False)`

python/mscclpp/language/pipeline.py
chhwang, line 1: Rename into `loop.py`?

python/mscclpp/language/program.py
```diff
@@ -1,474 +1,193 @@
...
-    if _current_program == None:
-        raise RuntimeError("No Program in context")
-    return _current_program
-
-
-# For msccl++ program, we have one assumption that for channel can be identified by (send_buffer, recv_buffer, type, send_tb/recv_tb)
-# which means the send_tb and recv_tb should be the same for a pair of signal and wait, also same for put/get operation.
-# If one sender what to send data to peer want to use different tb in receiver side. We need to send to same tb in receiver side first,
-# then performance a across tb sync. This is a limitation of current implementation.
 class MSCCLPPProgram:
```
chhwang: How about just `class Program`, or `class CollectiveProgram`?

python/mscclpp/language/rank.py
```diff
@@ -1,33 +1,382 @@
...
+            dst_chunk (Chunk): The destination chunk to copy data to.
+            src_chunk (Chunk): The source chunk to copy data from.
+            tb (int): The thread block ID that will execute this operation.
+
+        Example:
+            >>> rank.copy(dst_chunk, src_chunk, tb=0)
+        """
+        self._copy(dst_chunk=dst_chunk, src_chunk=src_chunk, tb=tb)
+
+    def unpack_packet(self, dst_chunk: Chunk, src_chunk: Chunk, tb: int):
```
chhwang: Let's make it plural form `unpack_packets()`

python/mscclpp/language/internal/threadblock.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def shift_buffers(self, instance, num_instances, replication_function):
+        for op in self.ops:
+            op.shift_buffers(instance, num_instances, replication_function)
+
+    def shift_ids(self, instance, num_instances, replication_function):
+        for op in self.ops:
+            op.shift_ids(instance, num_instances, replication_function)
+
+    def to_json(self) -> dict:
```
chhwang: There are many `to_json()` methods that actually returns a dict. Please rename into `to_dict()`.

python/mscclpp/language/internal/optmizer.py
chhwang, line 1: Typo in the file name `optmizer.py` --> `optimizer.py`

python/mscclpp/language/internal/optmizer.py
```diff
@@ -0,0 +1,61 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+from mscclpp.language.internal.operations import *
+from mscclpp.language.internal.types import SyncType
+
+
+def fuse_instructions(operations):
```
chhwang: Rename into `fuse_operations()`?

python/mscclpp/language/internal/optmizer.py
```diff
@@ -0,0 +1,61 @@
...
+            current_operation = fused_operation
+            next_operation_index += 1
+
+        fused_operations.append(current_operation)
+        operation_index = next_operation_index
+
+    return fused_operations
+
+
+def adding_data_sync(operations):
```
chhwang: Rename into `add_data_sync()`? Besides, this name doesn't sound like a optimization function.

python/mscclpp/language/internal/operations.py
```diff
@@ -0,0 +1,879 @@
...
+    DataAccess,
+    DataAccessType,
+)
+from dataclasses import dataclass, field
+from typing import List
+import uuid
+
+
+@dataclass
+class BaseOperation:
```
chhwang: Make it a virtual class? `class BaseOperation(ABC)`

python/mscclpp/language/internal/operations.py
```diff
@@ -0,0 +1,879 @@
...
+        return []
+
+    def shift_buffers(self, instance, num_instances, replication_function):
+        return
+
+    def shift_ids(self, instance, num_instances, replication_function):
+        return
+
+    def __add__(self, other):
+        return None
```
chhwang: Please add docstrings to define these methods

chhwang (commented):
In progress

python/mscclpp/language/channel.py
caiocbr: Yes, it's equivalent to __syncthreads() Okay, let me adjust the comments and change the default value.

python/mscclpp/language/channel.py
caiocbr: We have an operation called copy_packet, we typically use copy for transfers within the same GPU. I'm concerned this might cause confusion. Even though we already have another operation named copy_packet for intra-GPU data transfers, do you think it would still make sense to rename this to copy_packets?

python/mscclpp/language/channel.py
caiocbr: If we put it in the last position, I would need to assign a default value during the operation, which doesn’t seem logical. Perhaps we can keep it as the last parameter without a default value, or consider placing it as the first parameter. What are your thoughts?

python/mscclpp/language/general.py
caiocbr: I think the purpose of this file is to include some global variables and methods related to the program's context. Perhaps naming it 'global' would be more accurate.

python/mscclpp/language/channel.py
chhwang: Let's make it plural at least, `read_put_packets()`

chhwang (approved):

PR #580, opened 2025-07-24
Disable CI and fix bugs

Binyang2014:
- Temporary disable DSL ci pipeline
- Fix executor issue, only add constOffset for remote buffer


caiomcbr (approved):
LGTM

PR #585, opened 2025-07-25
Use smart pointer for IB structure 

Binyang2014:
Change to use smart pointer for IB structure. Registered memory will own ibMr, ibCtx will not held the reference
- Use smart pointer for IbQp and IbMr
- Update memoryChannel API, keep localRegisteredMemory
- Close fd when registedMemory released


src/include/registered_memory.hpp
```diff
@@ -56,6 +56,9 @@ struct RegisteredMemory::Impl {
   TransportFlags transports;
   std::vector<TransportInfo> transportInfos;
 
+  // Only used for IB transport
+  std::shared_ptr<const IbMr> ibMr;
```
chhwang: Shouldn't it be `unique_ptr`?

src/include/connection.hpp
```diff
@@ -38,7 +38,7 @@ class IBConnection : public Connection {
  private:
   Transport transport_;
   Transport remoteTransport_;
-  IbQp* qp_;
+  std::shared_ptr<IbQp> qp_;
```
chhwang: `Connection` already has a copied `Endpoint` and `Endpoint` owns the QP, so `IBConnection` doesn't need a shared pointer. `weak_ptr` is enough here

src/include/registered_memory.hpp
Binyang2014: Change to `unique_ptr` will make `registered_memory` be move only. I think registered memory needs to be shared in many places (channel/user code ...). 
Keep as `shared_ptr` seems has no side-effect

chhwang: RegisteredMemroy::Impl is already managed by a shared_ptr, so RegisteredMemory is always copiable regardless of unique_ptr inside Impl

caiomcbr (approved):
LGTM

PR #590, opened 2025-07-31
Fix GpuStreamPool to be aware of the device ID of streams

chhwang:


src/gpu_utils.cc
```diff
@@ -23,22 +26,29 @@ CudaStreamWithFlags::~CudaStreamWithFlags() {
...
-  if (!streams_.empty()) {
-    auto stream = streams_.back();
-    streams_.pop_back();
+  int deviceId;
+  MSCCLPP_CUDATHROW(cudaGetDevice(&deviceId));
+  auto& streamVec = streams_[deviceId];
+  if (!streamVec.empty()) {
+    auto stream = streamVec.back();
+    streamVec.pop_back();
     return GpuStream(gpuStreamPool(), stream);
```
Binyang2014, line 52: Make `std::shared_ptr<GpuStreamPool> gGpuStreamPool_` be static?

PR #593, opened 2025-07-31
Support CudaIpc connection within a single process

chhwang:
* Allow CudaIpc connection between GPUs in a single process
* Added an example of connection in a single process
* Minor interface updates


src/context.cc
```diff
@@ -77,6 +77,16 @@ MSCCLPP_API_CPP std::shared_ptr<Connection> Context::connect(Endpoint localEndpo
     int deviceId;
     if (localEndpoint.device().type == DeviceType::GPU) {
       deviceId = localEndpoint.device().id;
+      if (remoteEndpoint.device().type == DeviceType::GPU && localEndpoint.hostHash() == remoteEndpoint.hostHash() &&
+          localEndpoint.pidHash() == remoteEndpoint.pidHash()) {
+        // Connecting two GPUs in the same process - need to enable peer access explicitly
+        MSCCLPP_CUDATHROW(cudaSetDevice(deviceId));
```
Binyang2014: Do we need to restore the device id at the end of block?

src/context.cc
```diff
@@ -77,6 +77,29 @@ MSCCLPP_API_CPP std::shared_ptr<Connection> Context::connect(Endpoint localEndpo
...
+          localEndpoint.pidHash() == remoteEndpoint.pidHash()) {
+        // Connecting two GPUs in the same process - need to enable peer access explicitly
+        if (deviceId < 0) {
+          throw Error("No GPU device ID provided for local endpoint", ErrorCode::InvalidUsage);
+        }
+        int remoteDeviceId = remoteEndpoint.device().id;
+        if (remoteDeviceId < 0) {
+          throw Error("No GPU device ID provided for remote endpoint", ErrorCode::InvalidUsage);
+        }
+        int originalDeviceId;
```
Binyang2014, line 90: Maybe we can use the similar way as `AvoidCudaGraphCaptureGuard` to keep and restore deviceId. Seems we have two places where need this

Binyang2014 (approved):

src/context.cc
chhwang: Will open a PR for this

PR #604, opened 2025-08-08
MNNVL fix

chhwang:


src/semaphore.cc
```diff
@@ -26,6 +27,11 @@ struct SemaphoreStub::Impl {
 };
 
 static std::shared_ptr<uint64_t> gpuCallocToken() {
+#if (CUDA_NVLS_API_AVAILABLE)
+  if (isNvlsSupported()) {
+    return detail::gpuCallocPhysicalShared<uint64_t>(1, 0);
```
Binyang2014, line 32: Add comments for this line? To support multi node NVL communication

Binyang2014 (approved):

src/semaphore.cc
chhwang: Will create another PR

PR #606, opened 2025-08-08
AlltoAll Test Support

caiomcbr:


python/mscclpp/language/tests/single_node/alltoall/alltoall.py
```diff
@@ -0,0 +1,99 @@
...
+def alltoall_example(name, gpu_size, num_threads_per_block, min_message_size, max_message_size):
+    chunksperloop = 1
+    collective = AllToAll(gpu_size, chunksperloop, True)
+    with CollectiveProgram(
+        name,
+        collective,
+        gpu_size,
+        instances=4,
+        protocol="Simple",
+        reuse_resources=True,
```
Binyang2014: Change  reuse_resources to False or remove this field?

python/mscclpp/language/tests/single_node/alltoall/alltoall_pipeline.py
```diff
@@ -0,0 +1,109 @@
...
+def alltoall_example(name, gpu_size, num_threads_per_block, min_message_size, max_message_size):
+    chunksperloop = 1
+    collective = AllToAll(gpu_size, chunksperloop, True)
+    with CollectiveProgram(
+        name,
+        collective,
+        gpu_size,
+        instances=2,
+        protocol="Simple",
+        reuse_resources=True,
```
Binyang2014: Remove reuse_resources

python/mscclpp/language/tests/single_node/alltoall/alltoall_ring_get_naive.py
```diff
@@ -0,0 +1,107 @@
+# Copyright (c) Microsoft Corporation.
```
Binyang2014: Remove this file?

python/mscclpp/language/tests/single_node/alltoall/alltoall_ring_get_op.py
```diff
@@ -0,0 +1,106 @@
...
+
+import argparse
+from mscclpp.language.channel import *
+from mscclpp.language.rank import *
+from mscclpp.language.general import *
+from mscclpp.language.program import *
+from mscclpp.language.collectives import *
+
+
+def alltoall_example(name, gpu_size, num_threads_per_block, min_message_size, max_message_size):
```
Binyang2014: It's out-of-place algo, pls rename the file to let it easy to understand

python/mscclpp/language/tests/single_node/alltoall/alltoall_ring_naive.py
```diff
@@ -0,0 +1,102 @@
...
+import argparse
+from mscclpp.language.channel import *
+from mscclpp.language.rank import *
+from mscclpp.language.general import *
+from mscclpp.language.program import *
+from mscclpp.language.collectives import *
+from mscclpp.language.loop import *
+
+
+def alltoall_example(name, gpu_size, num_threads_per_block, min_message_size, max_message_size):
```
Binyang2014: Maybe we can remove all naive file?

python/mscclpp/language/tests/single_node/alltoall/alltoall.py
caiocbr: This algorithm is non-zero copy, so I should keep it as True, no?

python/mscclpp/language/tests/single_node/alltoall/alltoall_ring_get_op.py
caiocbr: You mean "op" is not enough and I should put explicit "out_of_place"?

python/mscclpp/language/tests/single_node/alltoall/alltoall.py
Binyang2014: I think no. If we want to reuse the scratch buffer, we need to make sure there is space in dst. For now we just assume the space is equal to input.

python/mscclpp/language/tests/single_node/alltoall/alltoall_ring_get_op.py
Binyang2014: Oh, I though it means we use get operation... How about name as oop

chhwang (approved):

PR #618, opened 2025-08-22
Fix NVLS correctness issue

Binyang2014:
Fix NVLS correctness issue when dytpe != fp16

### DIFF
--- a/apps/nccl/src/allreduce.hpp
+++ b/apps/nccl/src/allreduce.hpp
@@ -479,27 +479,43 @@ __global__ void __launch_bounds__(512, 1)
 }
 
 #if defined(__CUDA_ARCH__) && __CUDA_ARCH__ >= 900
+template <class T>
+MSCCLPP_DEVICE_INLINE constexpr std::size_t nElements() {
+  using U = std::remove_cv_t<std::remove_reference_t<T>>;
+  if constexpr (std::is_same_v<U, int> || std::is_same_v<U, unsigned int> || std::is_same_v<U, std::int32_t> ||
+                std::is_same_v<U, std::uint32_t>) {
+    return 1;
+  } else {
+    static_assert(16 % sizeof(U) == 0, "nElements: 16 bytes must be divisible by sizeof(T).");
+    return 16 / sizeof(U);
+  }
+}
+
 template <typename T>
 MSCCLPP_DEVICE_INLINE void handleMultiLoadReduceStore(T* src, T* dst, size_t srcOffset, size_t dstOffset, size_t size,
                                                       int tid, int nThreads) {
   // nvls can only handle 4 bytes alignment
   MSCCLPP_ASSERT_DEVICE(size % 4 == 0, "size must be 4 bytes aligned");
-  const size_t n16B = size / sizeof(mscclpp::f16x8);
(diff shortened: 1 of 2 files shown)


apps/nccl/src/allreduce.hpp
```diff
@@ -479,27 +479,42 @@ __global__ void __launch_bounds__(512, 1)
...
   MSCCLPP_ASSERT_DEVICE(size % 4 == 0, "size must be 4 bytes aligned");
-  const size_t n16B = size / sizeof(mscclpp::f16x8);
-  const size_t srcOffset4 = srcOffset / sizeof(mscclpp::f16x8);
-  const size_t dstOffset4 = dstOffset / sizeof(mscclpp::f16x8);
-  mscclpp::f16x8* src4 = (mscclpp::f16x8*)src;
-  mscclpp::f16x8* dst4 = (mscclpp::f16x8*)dst;
-  for (size_t idx = tid; idx < n16B; idx += nThreads) {
+  constexpr size_t nElem = nElements<T>();
+  using vectorType = mscclpp::VectorType<T, nElem>;
+  const size_t nVec = size / sizeof(vectorType);
```
chhwang, line 500: Do we handle `size < 16` somewhere else?

apps/nccl/src/allreduce.hpp
```diff
@@ -479,27 +479,42 @@ __global__ void __launch_bounds__(512, 1)
 }
 
 #if defined(__CUDA_ARCH__) && __CUDA_ARCH__ >= 900
+template <class T>
+MSCCLPP_DEVICE_INLINE constexpr std::size_t nElements() {
```
chhwang: Function name is misleading. How about `calcVectorSize()`

chhwang (changes requested):

apps/nccl/src/allreduce.hpp
Binyang2014: The nVec will be zero then go to this logic: https://github.com/microsoft/mscclpp/blob/86ab95d79bfd8786b2b6bfc44d4f095e17383ee4/apps/nccl/src/allreduce.hpp#L509-L518

apps/nccl/src/allreduce.hpp
Binyang2014: Fixed

apps/nccl/src/allreduce.hpp
```diff
@@ -479,27 +479,42 @@ __global__ void __launch_bounds__(512, 1)
 }
 
 #if defined(__CUDA_ARCH__) && __CUDA_ARCH__ >= 900
+template <class T>
+MSCCLPP_DEVICE_INLINE constexpr std::size_t calcVectorSize() {
+  using U = std::remove_cv_t<std::remove_reference_t<T>>;
+  if constexpr (std::is_same_v<U, std::int32_t> || std::is_same_v<U, std::uint32_t>) {
+    return 1;
+  } else {
+    static_assert(16 % sizeof(U) == 0, "nElements: 16 bytes must be divisible by sizeof(T).");
```
chhwang: please remove `nElements:` from the comment

chhwang (approved):

PR #621, opened 2025-08-29
Thread Block Group DSL

caiomcbr:
Supporting the creation of a group of thread block to perform some operation.

### DIFF
--- a/docs/guide/mscclpp-dsl.md
+++ b/docs/guide/mscclpp-dsl.md
@@ -219,6 +219,17 @@ The following figure shows how thread blocks and channels are replicated across 
 MSCCL++ DSL Instance Replication Overview
 ```
 
+## Thread Block Group
+
+This feature is currently a prototype. It allows you to define and use a set of thread blocks as a group to execute operations. By grouping thread blocks, you can allocate thread blocks in a non-uniform way, giving different operations different amounts of thread blocks as needed.
+
+For example:
+```python
+# Create a Thread Block Group with 4 thread blocks
+tbg = ThreadBlockGroup(tb_list=[0, 1, 2, 3])
+# Use the Thread Block Group to perform the copy operation
+rank.copy(output_buffer[0:1], input_buffer[0:1], tbg=tbg)
+```
 
 ## Execution plan
 
--- a/docs/guide/mscclpp-dsl.md
+++ b/docs/guide/mscclpp-dsl.md
@@ -228,7 +228,7 @@ For example:
 # Create a Thread Block Group with 4 thread blocks
 tbg = ThreadBlockGroup(tb_list=[0, 1, 2, 3])
 # Use the Thread Block Group to perform the copy operation
-rank.copy(output_buffer[0:1], input_buffer[0:1], tbg=tbg)
+rank.copy(output_buffer[0:1], input_buffer[0:1], tb_group=tbg)
 ```
 
 ## Execution plan
--- a/python/mscclpp/language/channel.py
+++ b/python/mscclpp/language/channel.py
@@ -141,7 +141,11 @@ class MemoryChannel:
                 dst_buff=[LocalChunk(dst_chunk.buffer, dst_chunk.index, dst_chunk.size)],
                 channel_ids=tb_channel_ids,
                 channel_type=self.channel_type,
-                tbg_info=(ThreadBlockGroupInfo(tb_group.get_internal_id(tb_id), tb_group.numtb()) if tb_group is not None else None),
+                tbg_info=(
+                    ThreadBlockGroupInfo(tb_group.get_internal_id(tb_id), tb_group.numtb())
+                    if tb_group is not None
+                    else None
+                ),
             )
             get_program().add_operation(self.src_rank, tb_id, op)
 
@@ -196,7 +200,11 @@ class MemoryChannel:
                 dst_buff=[RemoteChunk(dst_chunk.buffer, dst_chunk.index, dst_chunk.size, tb_chunk_id)],
                 channel_ids=tb_channel_ids,
                 channel_type=self.channel_type,
-                tbg_info=(ThreadBlockGroupInfo(tb_group.get_internal_id(tb_id), tb_group.numtb()) if tb_group is not None else None),
+                tbg_info=(
+                    ThreadBlockGroupInfo(tb_group.get_internal_id(tb_id), tb_group.numtb())
+                    if tb_group is not None
+                    else None
+                ),
             )
             get_program().add_operation(self.src_rank, tb_id, op)
 
@@ -256,7 +264,11 @@ class MemoryChannel:
                 dst_buff=[RemoteChunk(dst_chunk.buffer, dst_chunk.index, dst_chunk.size, tb_chunk_id)],
                 channel_ids=tb_channel_ids,
                 channel_type=self.channel_type,
-                tbg_info=(ThreadBlockGroupInfo(tb_group.get_internal_id(tb_id), tb_group.numtb()) if tb_group is not None else None),
+                tbg_info=(
+                    ThreadBlockGroupInfo(tb_group.get_internal_id(tb_id), tb_group.numtb())
+                    if tb_group is not None
+                    else None
+                ),
                 from_packet=True,
                 to_packet=True,
             )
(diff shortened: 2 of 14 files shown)


python/mscclpp/language/channel.py
```diff
@@ -268,25 +300,38 @@ def put_packets(self, dst_chunk: Chunk, src_chunk: Chunk, tb: int):
...
 
-        get_program().add_operation(self.src_rank, tb, op)
+            get_program().add_operation(self.src_rank, tb_id, op)
 
     def reduce(
         self,
         local_src_chunk: Chunk,
         remote_src_chunks: List[Chunk],
-        tb: int,
+        tb: int | ThreadBlockGroup,
```
Binyang2014: Use separate parameter name for tb and tbg? Which is more clear to user. Check inside the function to make sure only on is set.
```python
    if (tb is None) == (thread_block_group is None):
        raise ValueError("Specify exactly one of `tb ` or `thread_block_group`.")
```

python/mscclpp/language/tests/unit_tests/tbg/tbg_reduce_fuse_test.py
```diff
@@ -0,0 +1,65 @@
...
+
+        # Declare Thread Block Group
+        tbg = ThreadBlockGroup(tb_list=[0, 1, 2, 3])
+        tbg1 = ThreadBlockGroup(tb_list=[0, 1, 2])
+
+        # Perform fused reduce operations: multiple reductions with reduced overhead
+        # First reduce: combine input_buffer[0:1] and [1:2] into output_buffer[0:1]
+        rank.reduce(input_buffer[0:1], [input_buffer[1:2]], tb=tbg, dst_chunk=output_buffer[0:1])
+        # Second reduce: combine input_buffer[0:1] and [2:3] into output_buffer[0:1]
+        rank.reduce(input_buffer[0:1], [input_buffer[2:3]], tb=tbg1, dst_chunk=output_buffer[0:1])
```
Binyang2014: This will insert barrier, right?

python/mscclpp/language/thread_block_group.py
```diff
@@ -0,0 +1,53 @@
...
+
+class ThreadBlockGroup:
+    """
+    A group of thread blocks with unique identifiers.
+
+    This class manages a collection of thread blocks, ensuring uniqueness
+    and providing efficient lookup of thread block IDs.
+    """
+
+    def __init__(self, tb_list: List[Any]):
```
Binyang2014: Why type is `List[Any]` not `List[int]`

src/executor/execution_plan.cc
```diff
@@ -529,6 +538,8 @@ void ExecutionPlan::Impl::setupOperation(const nlohmann::json& op, Operation& op
       auto& buff = op["dst_buff"][i];
       size_t constOffset = 0;
       BufferType bufferType = BufferType::NONE;
+      size_t outputOffset = this->getOffset(this->inputSize, this->outputSize, buff["index"], bufferType) + constOffset;
+      size_t outputBufferSize = this->getBufferSize(this->inputSize, this->outputSize, buff["index"], buff["size"]);
```
Binyang2014: Move to line 559? Keep offset/size calculation logic in a same place

Binyang2014 (commented):
Do we need to add some doc for this?

python/mscclpp/language/tests/multi_node/allgather_tbg.py
```diff
@@ -0,0 +1,186 @@
...
+    chunksperloop = 1
+    collective = AllGather(num_gpus, chunksperloop, False)
+    with CollectiveProgram(
+        name,
+        collective,
+        num_gpus,
+        protocol="Simple",
+        num_threads_per_block=num_threads_per_block,
+        reuse_resources=False,
+        instances=1,
```
Binyang2014: Make it as a input argument

python/mscclpp/language/thread_block_group.py
Binyang2014: How about init as `__init__(self, start: int, stop: int)`, just as `range` function

python/mscclpp/language/thread_block_group.py
```diff
@@ -0,0 +1,53 @@
...
+            The integer ID of the thread block (0-indexed)
+
+        Raises:
+            ValueError: If the thread block is not in this group
+        """
+        if tb not in self._tb_id:
+            raise ValueError(f"Thread block {tb} not found in thread block group")
+        return self._tb_id[tb]
+
+    def __len__(self) -> int:
```
Binyang2014: Don't use len here. Maybe `thread_block_group.numtb()`?

python/mscclpp/language/thread_block_group.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def __init__(self, tb_list: List[Any]):
+        """
+        Initialize a ThreadBlockGroup with a list of thread blocks.
+
+        Args:
+            tb_list: List of thread block objects
+        """
+
+        self.tb_list: List[Any] = tb_list
```
Binyang2014: List[int] ?

python/mscclpp/language/tests/multi_node/allgather_tbg.py
```diff
@@ -0,0 +1,186 @@
...
+                    for peer in range(gpus_per_node):
+                        if peer != gpu:
+                            dst_rank_id = peer + gpus_per_node * node
+                            tbg_id = peer if peer < gpu else peer - 1
+                            remote_index = (peer + gpus_per_node * (node - step)) % num_gpus
+                            if step == 0:
+                                memory_channels[(dst_rank_id, src_rank_id, 0)].put(
+                                    scratch_buffer[dst_rank_id][local_index : local_index + 1],
+                                    input_buffer[0:1],
+                                    tbg=tbgs[tbg_id],
```
Binyang2014: tb_group=tbgs[tbg_id]? Instead of tbg=tbgs[tbg_id] ?

python/mscclpp/language/tests/unit_tests/tbg/tbg_reduce_fuse_test.py
caiocbr: No, because it's using different tbg, actually I fixed this example, because what I would like to ensure here is it will fuse this two reduce operation if I use the same tbg.

python/mscclpp/language/thread_block_group.py
caiocbr: Do you think we should restrict the tbg to only allow continuous ranges? Users wouldn't be able to create thread block groups like tb 0, tb 2; I thought in use a list to be more flexible. About the List, you are right, it should be int.

docs/guide/mscclpp-dsl.md
```diff
@@ -219,6 +219,17 @@ The following figure shows how thread blocks and channels are replicated across
...
+## Thread Block Group
+
+This feature is currently a prototype. It allows you to define and use a set of thread blocks as a group to execute operations. By grouping thread blocks, you can allocate thread blocks in a non-uniform way, giving different operations different amounts of thread blocks as needed.
+
+For example:
+```python
+# Create a Thread Block Group with 4 thread blocks
+tbg = ThreadBlockGroup(tb_list=[0, 1, 2, 3])
+# Use the Thread Block Group to perform the copy operation
+rank.copy(output_buffer[0:1], input_buffer[0:1], tbg=tbg)
```
Binyang2014: Change to: `rank.copy(output_buffer[0:1], input_buffer[0:1], tb_group=tbg)` ?

python/mscclpp/language/tests/single_node/allgather_tbg.py
```diff
@@ -0,0 +1,80 @@
...
+                    # Step 1: source signals to indicate it is ready to receive data
+                    ch.signal(tb=0, relaxed=True)
+                    # Step 2: wait for the destination rank to be ready
+                    ch.wait(tb=0, data_sync=SyncType.after, relaxed=True)
+                    # Step 3: Synchronize thread blocks and perform data transfer
+                    # Barrier ensures all thread blocks in the group are synchronized before data transfer
+                    src_rank.barrier(tb_list=tbg.tb_list)
+                    # Transfer data from source chunk to destination chunk using the memory channel
+                    # The ThreadBlockGroup (tbg) coordinates which thread blocks participate in the transfer
+                    ch.put(dst_chunk, src_chunk, tb=tbg)
```
Binyang2014: `tb_group=tbg`?

python/mscclpp/language/channel.py
```diff
@@ -219,21 +246,35 @@ def read_put_packets(self, dst_chunk: Chunk, src_chunk: Chunk, tb: int):
...
-            channel_type=self.channel_type,
-            from_packet=True,
-            to_packet=True,
-        )
+        if tb is not None:
+            tb_list = [tb]
+        elif tb_group is not None:
+            tb_list = tb_group.tb_list
+        else:
+            raise RuntimeError(
```
Binyang2014: Return `ValueError` here?

Binyang2014 (approved):

PR #628, opened 2025-09-08
Add token pool for cuCreate API

Binyang2014:
Create a tokenPool to allocate token. This feature is used to support inter node NVL and try to reduce the footprint caused by cuCreate

### DIFF
--- a/python/mscclpp/core_py.cpp
+++ b/python/mscclpp/core_py.cpp
@@ -70,9 +70,9 @@ void register_core(nb::module_& m) {
       .def_static("create_unique_id", &TcpBootstrap::createUniqueId)
       .def("get_unique_id", &TcpBootstrap::getUniqueId)
       .def("initialize", static_cast<void (TcpBootstrap::*)(UniqueId, int64_t)>(&TcpBootstrap::initialize),
-           nb::call_guard<nb::gil_scoped_release>(), nb::arg("uniqueId"), nb::arg("timeoutSec") = 30)
+           nb::call_guard<nb::gil_scoped_release>(), nb::arg("unique_id"), nb::arg("timeout_sec") = 30)
       .def("initialize", static_cast<void (TcpBootstrap::*)(const std::string&, int64_t)>(&TcpBootstrap::initialize),
-           nb::call_guard<nb::gil_scoped_release>(), nb::arg("ifIpPortTrio"), nb::arg("timeoutSec") = 30);
+           nb::call_guard<nb::gil_scoped_release>(), nb::arg("if_ip_port_trio"), nb::arg("timeout_sec") = 30);
 
   nb::enum_<Transport>(m, "Transport")
       .value("Unknown", Transport::Unknown)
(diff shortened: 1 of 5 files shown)


src/utils_internal.cc
```diff
@@ -232,4 +233,42 @@ void getRandomData(void* buffer, size_t bytes) {
...
+    }
+    self->allocationMap_[index] &= ~mask;
+  };
+
+  size_t size = allocationMap_.size();
+  for (size_t i = 0; i < size; i++) {
+    uint64_t mask = (i + 1 == size) ? tailMask_ : ~0ULL;
+    uint64_t holes = (~allocationMap_[i]) & mask;
+    if (!holes) continue;
+    size_t bit = __builtin_ctzll(holes);
```
chhwang: Please use portable code. Will std::bitmap help?

src/semaphore.cc
```diff
@@ -20,18 +20,21 @@ struct SemaphoreStub::Impl {
...
 };
 
-static std::shared_ptr<uint64_t> gpuCallocToken() {
-// #if (CUDA_NVLS_API_AVAILABLE)
-//   if (isNvlsSupported()) {
-//     return detail::gpuCallocPhysicalShared<uint64_t>(1, 0);
-//   }
-// #endif  // CUDA_NVLS_API_AVAILABLE
+std::shared_ptr<uint64_t> SemaphoreStub::Impl::gpuCallocToken(std::shared_ptr<Context> context) {
+  // NVLS is not supported in ROCm/HIP
```
chhwang: I’d remove this comment

src/include/context.hpp
```diff
@@ -35,13 +35,17 @@ class CudaIpcStream {
   int deviceId() const { return deviceId_; }
 };
 
+class TokenPool;
 struct Context::Impl {
   std::unordered_map<Transport, std::unique_ptr<IbCtx>> ibContexts_;
   std::vector<std::shared_ptr<CudaIpcStream>> ipcStreams_;
+  std::shared_ptr<TokenPool> tokenPool_;
+  const size_t maxNumTokens_ = 1 << 15;  // 32K tokens
```
chhwang, line 43: Why 32K?

src/utils_internal.cc
Binyang2014: It has `std::countr_zero` but only available after C++20. Which require nvcc version>=12, hipcc version>=6.0

src/include/context.hpp
Binyang2014: I think the min granularity for cuMemCreate is 2MB (2^21). sizeof(token) = 64 (2^6). Then the max token is 2^15 if the tokenPoolSize is 2MB

chhwang: Will it break the code if the min granularity changes?

src/utils_internal.cc
chhwang: We can just do `bit=0; while ((holes&1)==0) {holes>>=1; bit++;}`

src/include/context.hpp
Binyang2014: No, it will not break the code. We just keep more unused memory.

PR #638, opened 2025-09-25
Address teardown issue

Binyang2014:
Ignore cuda/cu errors during teardown. Some pointer may be invalid at this point


caiomcbr (approved):
LGTM

PR #640, opened 2025-09-29
Add MSCCLPP_GIT_COMMIT micro

Binyang2014:
- Add MSCCLPP_GIT_COMMIT micro
- Update docs


.github/workflows/update-version.yml
```diff
@@ -27,30 +27,13 @@ jobs:
...
-
-          # Update CMakeLists.txt
-          sed -i "s/set(MSCCLPP_MAJOR \".*\")/set(MSCCLPP_MAJOR \"${MAJOR}\")/" CMakeLists.txt
-          sed -i "s/set(MSCCLPP_MINOR \".*\")/set(MSCCLPP_MINOR \"${MINOR}\")/" CMakeLists.txt
-          sed -i "s/set(MSCCLPP_PATCH \".*\")/set(MSCCLPP_PATCH \"${PATCH}\")/" CMakeLists.txt
-
-          # Update header files
-          sed -i "s/#define MSCCLPP_MAJOR .*/#define MSCCLPP_MAJOR ${MAJOR}/" include/mscclpp/core.hpp
-          sed -i "s/#define MSCCLPP_MINOR .*/#define MSCCLPP_MINOR ${MINOR}/" include/mscclpp/core.hpp
-          sed -i "s/#define MSCCLPP_PATCH .*/#define MSCCLPP_PATCH ${PATCH}/" include/mscclpp/core.hpp
```
seagater, line 47: Now we no longer touch CMakeList.txt and core.hpp, do we have other files rely on version updates in core.hpp or CMakeLists.txt?

CMakeLists.txt
```diff
@@ -13,8 +16,27 @@ endif()
...
+  endif()
+else()
+    message(WARNING "Git not found, setting GIT_HASH to 'UNKNOWN'")
+endif()
+
+configure_file(
+  "${CMAKE_CURRENT_SOURCE_DIR}/include/mscclpp/version.hpp.in"
+  "${CMAKE_CURRENT_BINARY_DIR}/include/mscclpp/version.hpp"
+  @ONLY
+)
```
seagater, line 39: Do we need to update README or quickstart.md to add note about the new dependency on Git?

.github/workflows/update-version.yml
Binyang2014: Double checked, there is no file rely on version

CMakeLists.txt
Binyang2014: Seems no need, we require `git clone` for installation, which already requires git. No new dependency introduced

.github/workflows/update-version.yml
seagater: Ok. that's great. We can proceed with this change.

CMakeLists.txt
seagater: Got it. It should be ok if there is no new dependency.

seagater (approved):
The PR looks good. There are no other files rely on version and no new dependency besides of Git.

PR #641, opened 2025-09-30
Address corner case when generating version file

Binyang2014:
Address corner case for version file generation


_build_version_metadata.py
```diff
@@ -34,12 +33,19 @@ def _get_version():
...
-        # Convert "0.7.1.dev36+g6e2360d69.d20250926" to "0.7.1.dev36+g6e2360d69"
+        # Convert "x.y.z.dev36+g6e2360d69.d20250926" to "x.y.z.dev36+g6e2360d69"
         version = re.sub(r"\.d\d{8}", "", version)
 
         # Use the value in VERSION as the base version
-        # Change to "0.7.0.dev36+g6e2360d69"
-        version = re.sub(r"^[0-9]+\.[0-9]+\.[0-9]+", base_version, version)
+        # Change x.y.z.dev36+g6e2360d69 or x.y.dev36+g6e2360d69 to "base_version.dev36+g6e2360d69"
+        # Use a more specific pattern to match the version prefix and ensure correct replacement
+        new_version = re.sub(r"^[0-9]+(?:\.[0-9]+)*(\.dev\d+\+g[0-9a-fA-F]+)$", rf"{base_version}\1", version)
```
seagater: What if there is no 'dev#N' in version,  e,g., version = '0.7.1+g6e2360d69' or version = '0.8+g6e2360d69'?
The re.sub will not work. And new_version = version.
Should we change to  re.sub(r"^[0-9]+(?:\\.[0-9]+)*((?:\\.dev\d+)?\\+g[0-9a-fA-F]+)$", rf"{base_version}\1", version)？

seagater (approved):
The changes looks good to me. Since we use default version format "mscclpp-0.8.0.post1.dev0+gc632fee37.d20251007". It's ok to remove the special _build_version_metadata.py

PR #642, opened 2025-10-01
Fix Rocm build issue

Binyang2014:
Fix cross-compiling issue. Add `target_compile_options` to make sure the compile option is correct


seagater (approved):
Looks good to me.

PR #648, opened 2025-10-17
Resolve IBVerbs Loading Issues

caiomcbr:
Some systems do not include libibverbs.so when installing ibverbs; instead, they only provide libibverbs.so1. This PR updates the CMake file to search for this library and modifies the wrapper to load it.


CMakeLists.txt
```diff
@@ -148,6 +148,16 @@ find_package(IBVerbs)
...
 find_package(Threads REQUIRED)
 
+set(CMAKE_COLOR_DIAGNOSTICS ON)
+function(msg_red text)
+  string(ASCII 27 ESC)
+  message("${ESC}[31m${text}${ESC}[0m")
+endfunction()
+
+if(NOT IBVERBS_FOUND)
+  msg_red("libibverbs not found. Install libibverbs-dev or rdma-core-devel.")
```
chhwang, line 158: Let's not make it look like an error message, as IB is not a must.

caiocbr: Are you suggesting we update the message to display only when libibverbs is not found? Maybe we could change the color to orange instead of red. The challenge is that sometimes we intend to use ibverbs, but the environment only provides libibverbs.so1. It's not straightforward to spot this, as our cmake build outputs several messages consecutively. This issue was also reported here too: https://github.com/sgl-project/sglang/issues/6834

chhwang: @caiomcbr We may add a new cmake option `MSCCLPP_USE_IB` to enable explicit failure.

chhwang: Done by #667

PR #656, opened 2025-10-21
Fix docs

chhwang:
* Fix Python doc generation
* Remove `ChannelTrigger` and fix `ProxyTrigger`
* Fixed package versions for consistency

### DIFF
--- a/apps/nccl/src/allreduce.cu
+++ b/apps/nccl/src/allreduce.cu
@@ -123,6 +123,26 @@ struct Allreduce8Adapter {
   }
 };
 
+template <Op OpType, typename T>
+struct AllreduceNvlsPacketAdapter {
+  static cudaError_t call(const void* input, void* scratch, void* output, void*, void*,
+                          mscclpp::DeviceHandle<mscclpp::SwitchChannel>* nvlsChannels,
+                          mscclpp::DeviceHandle<mscclpp::SwitchChannel>*, size_t, size_t, size_t scratchBufferSize,
+                          int rank, int, int worldSize, size_t nelems, cudaStream_t stream, uint32_t* deviceFlag,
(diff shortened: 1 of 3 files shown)


.github/workflows/doc-build.yaml
```diff
@@ -10,25 +10,26 @@ permissions:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
Binyang2014, line 13: The latest image doesn't work?

docs/conf.py
```diff
@@ -3,21 +3,27 @@
...
+sys.path.insert(0, str(python_path))
+
 # -- Project information -----------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
 
 project = "mscclpp"
 copyright = "2025, MSCCL++ Team"
 author = "MSCCL++ Team"
-release = "v0.8.0"
+release = "v" + open("../VERSION").read().strip()
```
Binyang2014: Seems we can delete this github action: https://github.com/microsoft/mscclpp/blob/main/.github/workflows/update-version.yml

Can we get version via this way: https://setuptools-scm.readthedocs.io/en/latest/usage/#usage-from-sphinx? May need to install mscclpp package first

.github/workflows/doc-build.yaml
chhwang: The latest image cannot find Doxygen v1.9.1 by default. We can use later Ubuntu when we can upgrade Doxygen version.

docs/conf.py
chhwang: Please check my updates; I only generated the `_version.py` file instead of installing the package

PR #677, opened 2025-11-07
Supporting New Packet Kernel Operation at Executor

caiomcbr:
This PR introduces three new operations to enhance flexibility and performance at executor.

One operation can be invoked directly via the DSL API and two operations are created through fusion of existing operations, reducing overhead and improving efficiency.

1. Port Channel Put Packet (Direct DSL API Call): Sends data from pkt format to the remote side in pkt format via the port channel. Both source and destination buffers must be scratch.

2. Reduce Copy Packet (Fusion):
Reduce Packet+Copy Packet=Reduce Copy Packet
Triggered when the destination buffer of Reduce Packet matches the source buffer of Copy Packet.
Purpose: Combine reduction and copy into a single step for better performance.

3. Reduce Copy Send Packet (Fusion):
Reduce Copy Packet+Put Packet=Reduce Copy Send Packet (when dst buffer of Reduce Copy Packet matches src buffer of Put Packet)
Reduce Copy Packet+Read Put Packet=Reduce Copy Send Packet (when dst pkt buffer of Reduce Copy Packet matches src buffer of Read Put Packet)
Purpose: Combine reduction, copy, and send operations into one optimized pipeline.

Fusion Diagram
Reduce Packet + Copy Packet → Reduce Copy Packet
Reduce Copy Packet + Put Packet → Reduce Copy Send Packet
Reduce Copy Packet + Read Put Packet → Reduce Copy Send Packet

Beyond this, this PR adjust the AllReduce 2 Node algorithm:

Message Size  |  Latency (µs)
        1K            |     15.34
        2K            |     15.88
        4K            |     15.71
        8K            |     16.01
        16K          |     15.88
        32K          |     16.21
        64K          |     16.90
        128K        |     18.24
        256K        |     20.39
        512K        |     25.26
        1M           |     32.74
        2M           |     53.64

### DIFF
--- a/include/mscclpp/port_channel.hpp
+++ b/include/mscclpp/port_channel.hpp
@@ -19,7 +19,7 @@ class BaseProxyService {
  public:
   BaseProxyService() = default;
   virtual ~BaseProxyService() = default;
-  virtual void startProxy() = 0;
+  virtual void startProxy(bool blocking = false) = 0;
   virtual void stopProxy() = 0;
 };
 
@@ -73,7 +73,8 @@ class ProxyService : public BaseProxyService {
   PortChannel portChannel(SemaphoreId id, MemoryId dst, MemoryId src);
 
   /// Start the proxy service.
-  void startProxy();
+  /// @param blocking Whether to block until the proxy thread has started (default: false).
+  void startProxy(bool blocking = false);
 
   /// Stop the proxy service.
   void stopProxy();
--- a/include/mscclpp/proxy.hpp
+++ b/include/mscclpp/proxy.hpp
@@ -42,7 +42,7 @@ class Proxy {
   ~Proxy();
 
   /// Start proxy.
-  void start();
+  void start(bool blocking = false);
 
   /// Stop proxy.
   void stop();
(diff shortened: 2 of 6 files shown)


python/mscclpp/__main__.py
```diff
@@ -27,11 +27,32 @@
             reuse_resources=True,
             use_double_scratch_buffer=True,
             min_message_size=1 << 10,
+            max_message_size=64 << 10,
+            tags={"default": 1},
+        ),
+        "additional_args": [1],
```
Binyang2014: What's the meaning of `[1]`, maybe we can change to a more meaningful name

python/mscclpp/language/channel.py
```diff
@@ -682,6 +682,55 @@ def put_with_signal_and_flush(self, dst_chunk: Chunk, src_chunk: Chunk, tb: int)
 
         get_program().add_operation(self.src_rank, tb, op)
 
+    def put_packets(self, dst_chunk: Chunk, src_chunk: Chunk, tb: int):
```
Binyang2014, line 685: For this op, we support both memory channel and port channel right?

python/mscclpp/__main__.py
caiocbr: This is the extra function parameter. It means the thread block group size.

python/mscclpp/language/channel.py
caiocbr: Yes, this PR add the support of this operation for port channels. Before it was only supported by memory channel.

python/mscclpp/__main__.py
Binyang2014: Can we use **kargs for additional args, then we could have more meaningful name

PR #680, opened 2025-11-11
`connect()` APIs changed to return an instance instead of a shared_ptr

chhwang:
The key purpose is handling all mscclpp objects' memory internally by hiding shared pointers from user APIs.
* `Connection` class is now a wrapper of `BaseConnection` class that is equivalent to the previous `Connection` class
* `connect()` methods now return `Connection` instead of `std::shared_ptr<Connection>`
* Removed `connectOnSetup()` method


include/mscclpp/core.hpp
```diff
@@ -587,12 +588,7 @@ class RegisteredMemory {
 class Connection {
  public:
   /// Constructor.
-  /// @param context The context associated with the connection.
-  /// @param localEndpoint The local endpoint of the connection.
-  Connection(std::shared_ptr<Context> context, const Endpoint& localEndpoint);
-
-  /// Destructor.
-  virtual ~Connection() = default;
+  Connection() = default;
```
Binyang2014, line 591: Do we need default constructor?

include/mscclpp/port_channel.hpp
```diff
@@ -82,7 +82,7 @@ class ProxyService : public BaseProxyService {
   std::vector<std::shared_ptr<Host2DeviceSemaphore>> semaphores_;
   std::vector<RegisteredMemory> memories_;
   std::shared_ptr<Proxy> proxy_;
-  std::unordered_map<std::shared_ptr<Connection>, int> inflightRequests_;
+  std::unordered_map<const BaseConnection*, int> inflightRequests_;
```
Binyang2014: Can we use ` std::unordered_map<const Connection, int> inflightRequests_` here? Not sure if it's safe to use raw pointer

include/mscclpp/core.hpp
chhwang: Yes, because we have a Connection instance that may not be initialized in `SemaphoreStub::Impl` (which means the stub is from remote)

include/mscclpp/port_channel.hpp
chhwang: Changed into `std::shared_ptr<BaseConnection>` instead of the raw pointer

Binyang2014 (approved):

PR #683, opened 2025-11-13
Fix IB Error in DSL Executor

caiomcbr:


src/executor/executor.cc
```diff
@@ -99,8 +99,10 @@ struct hash<mscclpp::DeviceExecutionPlanKey> {
 }  // namespace std
 
 namespace {
-auto inSameNode = [](int rank1, int rank2, int nranksPerNode) {
-  return rank1 / nranksPerNode == rank2 / nranksPerNode;
+auto useIB = [](int rank1, int rank2, int nranksPerNode) {
+  bool inSameNode = rank1 / nranksPerNode == rank2 / nranksPerNode;
```
Binyang2014, line 111: If inSameNode, we use cudaIpc, else use IB.

caiocbr: You are right! Let me fix it.

PR #686, opened 2025-11-17
Fix Error in Non IB Env at Executor

caiomcbr:


src/executor/executor.cc
```diff
@@ -99,8 +99,17 @@ struct hash<mscclpp::DeviceExecutionPlanKey> {
 }  // namespace std
 
 namespace {
-auto inSameNode = [](int rank1, int rank2, int nranksPerNode) {
-  return rank1 / nranksPerNode == rank2 / nranksPerNode;
+auto hasIBDevices = []() {
+#if defined(USE_IBVERBS)
```
chhwang: No need this `#if` directive since `getIBDeviceCount()` already handles by itself

src/executor/executor.cc
```diff
@@ -99,8 +99,17 @@ struct hash<mscclpp::DeviceExecutionPlanKey> {
...
+  return mscclpp::getIBDeviceCount() > 0;
+#else
+  return false;
+#endif
+};
+
+auto useIB = [](int rank1, int rank2, int nranksPerNode) {
+  bool inSameNode = rank1 / nranksPerNode == rank2 / nranksPerNode;
+  return hasIBDevices() && !inSameNode;
```
chhwang: We can just do `return (getIBDeviceCount() > 0) && !inSameNode;`

PR #692, opened 2025-11-19
Torch integration

Binyang2014:
Reorganize current native algorithm implementation and DSL algorithm implementation.
Provide unified API for DSL algo and native algo and provide interface to tune the algo
Provide interface for pytorch integration with native API and DSL

### DIFF
--- a/.azure-pipelines/templates/integration-test.yaml
+++ b/.azure-pipelines/templates/integration-test.yaml
@@ -202,7 +202,7 @@ steps:
         export PATH=/usr/local/mpi/bin:\$PATH;                     \
         export LD_LIBRARY_PATH=/root/mscclpp/build/lib:\$LD_LIBRARY_PATH; \
         python3 -m pip install .;                                     \
-        mpirun --allow-run-as-root -tag-output -x LD_LIBRARY_PATH=/usr/local/cuda/compat:$LD_LIBRARY_PATH -x MSCCLPP_HOME=/root/mscclpp -np 8 python3 ./python/mscclpp_benchmark/allreduce_bench.py"'
+        mpirun --allow-run-as-root -tag-output -x MSCCLPP_HOME=/root/mscclpp -np 8 python3 ./python/mscclpp_benchmark/allreduce_bench.py"'
       kill $CHILD_PID
     workingDirectory: '$(System.DefaultWorkingDirectory)'
 
--- a/.azure-pipelines/templates/ut-no-ib-env.yaml
+++ b/.azure-pipelines/templates/ut-no-ib-env.yaml
@@ -73,7 +73,7 @@ steps:
         export PATH=/usr/local/mpi/bin:\$PATH                          \
         export LD_LIBRARY_PATH=/root/mscclpp/build/lib:\$LD_LIBRARY_PATH;  \
         cd /root/mscclpp;                                              \
-        mpirun --allow-run-as-root -tag-output -x MSCCLPP_HOME=/root/mscclpp -x LD_LIBRARY_PATH=/usr/local/cuda/compat:$LD_LIBRARY_PATH -np 8 python3 -m pytest ./python/test/test_mscclpp.py::test_executor -x"'
+        mpirun --allow-run-as-root -tag-output -x MSCCLPP_HOME=/root/mscclpp -np 8 python3 -m pytest ./python/test/test_mscclpp.py::test_executor -x"'
       kill $CHILD_PID
     workingDirectory: '$(System.DefaultWorkingDirectory)'
 
--- a/.azure-pipelines/templates/ut.yaml
+++ b/.azure-pipelines/templates/ut.yaml
@@ -119,7 +119,7 @@ steps:
         export PATH=/usr/local/mpi/bin:\$PATH                          \
         export LD_LIBRARY_PATH=/root/mscclpp/build/lib:\$LD_LIBRARY_PATH;  \
         cd /root/mscclpp;                                              \
-        mpirun --allow-run-as-root -tag-output -x MSCCLPP_HOME=/root/mscclpp -x LD_LIBRARY_PATH=/usr/local/cuda/compat:$LD_LIBRARY_PATH -np 8 python3 -m pytest ./python/test/test_mscclpp.py -x"'
+        mpirun --allow-run-as-root -tag-output -x MSCCLPP_HOME=/root/mscclpp -np 8 python3 -m pytest ./python/test/test_mscclpp.py -x"'
       kill $CHILD_PID
     workingDirectory: '$(System.DefaultWorkingDirectory)'
 
--- a/python/mscclpp/utils.py
+++ b/python/mscclpp/utils.py
@@ -44,8 +44,8 @@ class Kernel:
     CU_LAUNCH_PARAM_BUFFER_SIZE = 0x02
     CU_LAUNCH_PARAM_END = 0x00 if not cp.cuda.runtime.is_hip else 0x03
 
-    def __init__(self, ptx: bytes, kernel_name: str):
-        self._module = cp.cuda.driver.moduleLoadData(ptx)
+    def __init__(self, cubin: bytes, kernel_name: str):
+        self._module = cp.cuda.driver.moduleLoadData(cubin)
         self._kernel = cp.cuda.driver.moduleGetFunction(self._module, kernel_name)
 
     def launch_kernel(
@@ -95,8 +95,8 @@ class KernelBuilder:
         self.macros = None
         if file_dir:
             self.macros = ["-D{}={}".format(macro, value) for macro, value in macro_dict.items()]
-        ptx = self._compile_cuda(os.path.join(self._current_file_dir, file), f"{kernel_name}.ptx")
-        self._kernel = Kernel(ptx, kernel_name)
+        cubin = self._compile_cuda(os.path.join(self._current_file_dir, file), f"{kernel_name}.cubin")
+        self._kernel = Kernel(cubin, kernel_name)
         self.kernel_map[kernel_key] = self._kernel
 
     def _compile_cuda(self, source_file, output_file, std_version="c++17"):
@@ -110,13 +110,13 @@ class KernelBuilder:
             command = [
                 nvcc,
                 f"-std={std_version}",
-                "-ptx",
+                "-cubin",
                 "-Xcompiler",
                 "-Wall,-Wextra",
                 f"-I{include_dir}",
                 f"{source_file}",
                 f"--gpu-architecture=compute_{compute_capability}",
-                f"--gpu-code=sm_{compute_capability},compute_{compute_capability}",
+                f"--gpu-code=sm_{compute_capability}",
                 "-o",
                 f"{self._tempdir.name}/{output_file}",
             ]


src/core/algorithm.cc
chhwang, line 1: I don't think it's a good idea to maintain algos in mscclpp core

python/mscclpp/_core/algorithm.py
chhwang, line 1: Why underscore prefix in this file name?

python/csrc/core_py.cpp
```diff
@@ -60,8 +68,6 @@ void register_core(nb::module_& m) {
       .def("recv", static_cast<void (Bootstrap::*)(std::vector<char>&, int, int)>(&Bootstrap::recv), nb::arg("data"),
            nb::arg("peer"), nb::arg("tag"));
 
-  nb::class_<UniqueId>(m, "UniqueId");
```
chhwang, line 64: I think users will need it for custom bootstrapping

python/mscclpp/_core/algorithm.py
Binyang2014: Underscore mean it's an internal module. Treat it as internal module because I want to let users import the Algorithm class from mscclpp, not from mscclpp.algorithm

src/core/algorithm.cc
Binyang2014: Okay, maybe we can treat algos as a  mscclpp extension. let me move the algorithms to another folder and generate another .so file.  Our nccl.so will link mscclpp_ext.so. How do you think this structure? 
mscclpp (project root) -->  core <include, src>, python, ext -> algos.

python/csrc/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@
 
 find_package(Python 3.8 COMPONENTS Interpreter Development.Module REQUIRED)
 include(FetchContent)
-FetchContent_Declare(nanobind GIT_REPOSITORY https://github.com/wjakob/nanobind.git GIT_TAG v1.4.0)
+FetchContent_Declare(nanobind GIT_REPOSITORY https://github.com/wjakob/nanobind.git GIT_TAG v1.9.2)
```
chhwang, line 6: For what do we upgrade nanobind?

include/mscclpp/algorithm.hpp
```diff
@@ -14,26 +14,142 @@
...
+  std::vector<MemoryChannel> memoryChannels;
+  std::vector<SwitchChannel> switchChannels;
+  std::vector<PortChannel> portChannels;
+  std::vector<std::shared_ptr<NvlsConnection>> nvlsConnections;
+  std::shared_ptr<DeviceHandle<MemoryChannel>> memoryChannelDeviceHandles;
+  std::shared_ptr<DeviceHandle<SwitchChannel>> switchChannelDeviceHandles;
+  std::shared_ptr<DeviceHandle<PortChannel>> portChannelDeviceHandles;
+  std::vector<std::shared_ptr<MemoryDevice2DeviceSemaphore>> memorySemaphores;
+  std::vector<std::shared_ptr<Host2DeviceSemaphore>> hostSemaphores;
   std::unordered_map<std::string, std::shared_ptr<void>> extras;
```
chhwang: Are they need to be public? Note this is a user-facing header.

chhwang:
> Putting all my minor comments below.
> 
> `examples/torch-integration/customized_allgather.cu`: "MIT license" -> "MIT License" `python/csrc/ext/algorithm_collection_builder.cpp`: rename the file into `python/csrc/ext/algorithm_collection_builder_py.cpp` `python/mscclpp/_core/__init__.py`: "MIT license" -> "MIT License" `python/mscclpp/_core/buffer.py`: "MIT license" -> "MIT License" `include/mscclpp/ext/collectives/algorithm_collection_builder.hpp`: missing license `src/core/include/logger.hpp`: let `case LogSubsys::COUNT:` case fallback to default (return "UNKNOWN"), since it's only for counting the number of `LogSubsys` types and using this directly for logging is not a valid behavior. Let's put a regarding comment here since copilot keeps suggesting this. `src/core/algorithm.cc`: "MIT license" -> "MIT License" `src/core/CMakeLists.txt`: "MIT license" -> "MIT License" `src/core/gpu_utils.cc`: fix the comment "!defined(**HIP_PLATFORM_AMD**)" -> "!defined(MSCCLPP_USE_ROCM)" `src/ext/collectives/allgather/allgather_fullmesh2.cu`: rename the file into `src/ext/collectives/allgather/allgather_fullmesh_2.cu` (would be better if we give a proper name rather than "_2") `src/ext/collectives/allreduce/allreduce_nvls_with_copy2.cu`: rename the file into `src/ext/collectives/allreduce/allreduce_nvls_with_copy_2.cu` (would be better if we give a proper name rather than "_2") `src/ext/collectives/include/allgather/allgather_fullmesh.hpp`: missing #ifdef guard in this header file - use `#ifdef MSCCLPP_EXT_<FILE_NAME>_HPP` `src/ext/collectives/include/allgather/allgather_fullmesh2.hpp`: rename the file into `src/ext/collectives/include/allgather/allgather_fullmesh_2.hpp` (would be better if we give a proper name rather than "_2"). Also, missing #ifdef guard in this header file - use `#ifdef MSCCLPP_EXT_<FILE_NAME>_HPP` `src/ext/collectives/include/allreduce/allreduce_allpair_packet.hpp`: "MIT license" -> "MIT License" `src/ext/collectives/include/allreduce/allreduce_nvls_packet.hpp`: "MIT license" -> "MIT License" `src/ext/collectives/include/allreduce/allreduce_nvls_with_copy2.hpp`: rename the file into `src/ext/collectives/include/allreduce/allreduce_nvls_with_copy_2.hpp` (would be better if we give a proper name rather than "_2") `src/ext/collectives/include/collective_utils.hpp`: follow `#ifdef MSCCLPP_EXT_<FILE_NAME>_HPP` format for the header guard `src/ext/collectives/algorithm_collection_builder.cc`: missing license `src/ext/collectives/CMakeLists.txt`: "MIT license" -> "MIT License" `src/ext/nccl/CMakeLists.txt`: "MIT license" -> "MIT License" `src/ext/CMakeLists.txt`: missing license `include/mscclpp/algorithm.hpp`: enum value names of `CommResult` (e.g., commSuccess) is inconsistent with other enums. But it would be too long if we use CAPITAL_CASE style, so I'd recommend CamelCase style, including in `CommResult` and in other enums in this header.
> 
> Overall: Review the documentations under `docs/` to see if all the texts are still aligned with the changed paths. Especially, we need to update the binary paths (such as `mp_unit_tests` build path) mentioned in documentations.

@copilot work on this

python/csrc/CMakeLists.txt
Binyang2014: For supporting ptyhon3.12

include/mscclpp/algorithm.hpp
Binyang2014: Users need this to implement customized algorithm, for example: https://github.com/microsoft/mscclpp/blob/105239fc6c8c246e77a6f54504cb2836a67ca234/examples/customized-collective-algorithm/customized_allgather.cu

Also, this structure is not stable, will change to use BaseChannel instead Channel which memory attached in future.

Binyang2014: Move to the internal header folder

python/mscclpp/_core/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+from .algorithm import *
+from .comm import *
+from .compiler import *
+from .buffer import *
+
+__all__ = []
+__all__ += algorithm.__all__
+__all__ += comm.__all__
+__all__ += compiler.__all__
+__all__ += buffer.__all__
```
chhwang, line 13: unneeded

python/mscclpp/_core/algorithm.py
```diff
@@ -0,0 +1,230 @@
...
+
+from __future__ import annotations
+from typing import Optional, Tuple, Dict
+from functools import cached_property
+
+
+from mscclpp._mscclpp import (
+    Algorithm as _Algorithm,
+    DslAlgorithm as _DslAlgorithm,
+    AlgorithmType as _AlgorithmType,
```
chhwang, line 12: Python-binded objects would be named differently in the future (e.g., adding `Cpp*` prefix) to avoid confusion

python/mscclpp/_core/compiler.py
```diff
@@ -0,0 +1,350 @@
...
+                        "nranks_per_node": algo_spec.nranks_per_node,
+                        "world_size": algo_spec.world_size,
+                        "instances": algo_spec.instances,
+                        "protocol": algo_spec.protocol,
+                    },
+                }
+            )
+        ).hexdigest()
+
+        plan_dir = os.environ.get("MSCCLPP_EXECUTION_PLAN_DIR", Path.home() / ".cache/mscclpp")
```
chhwang, line 141: 1. All env variables should be collected and documented in `env.cpp`, even if it's used only at Python level.
2. Is execution plan a part of "cache", meaning, this belongs to the system not to the user?
3. I'd prefer `MSCCLPP_CACHE_DIR` env and we internally decide what to put where.

python/mscclpp/_core/compiler.py
```diff
@@ -0,0 +1,350 @@
...
+    This compiler takes CUDA or HIP source files containing custom collective
+    algorithm kernels and compiles them into loadable Python modules using
+    pybind11 bindings.
+
+    The compiler automatically detects whether to use NVCC (CUDA) or HIPCC (ROCm)
+    based on the runtime environment. Compiled modules are cached to avoid
+    recompilation.
+
+    The cache location can be configured via the `MSCCLPP_NATIVE_CACHE_DIR`
+    environment variable (defaults to `~/.cache/mscclpp/native`).
```
chhwang, line 183: I'd prefer `MSCCLPP_CACHE_DIR` env and we internally decide what to put where.

python/mscclpp/_core/compiler.py
```diff
@@ -0,0 +1,350 @@
...
+        """Get the path to the appropriate compiler.
+
+        Returns:
+            Path to nvcc (CUDA) or hipcc (ROCm) compiler.
+        """
+        if self._is_hip:
+            rocm_home = os.environ.get("ROCM_HOME")
+            return os.path.join(rocm_home, "bin/hipcc") if rocm_home else "hipcc"
+        else:
+            cuda_home = os.environ.get("CUDA_HOME")
```
chhwang, line 243: Why not `MSCCLPP_CUDA_HOME`/`MSCCLPP_ROCM_HOME`?

python/mscclpp/ext/__init__.py
```diff
@@ -0,0 +1,6 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT license.
+
+from .algorithm_collection_builder import *
+
+__all__ = algorithm_collection_builder.__all__
```
chhwang, line 6: unneeded

src/core/executor/execution_kernel.cu
```diff
@@ -67,6 +67,10 @@ void ExecutionKernel::launchKernel(int rank, int nthreadblocks, int nthreads, vo
       );
 #endif
       break;
+    case DataType::FP8_E4M3:
+    case DataType::FP8_E5M2:
+      // FP8 is not supported in CUDA execution kernel.
```
chhwang, line 72: Should we warn or throw here?

Binyang2014:
Will resolve all the comments in another PR

python/mscclpp/_core/compiler.py
Binyang2014: I think `CUDA_HOME` and `ROCM_HOME` are common evn used by many applications. We don't need to introduce another env name

PR #711, opened 2025-12-23
Use uncached memory on Rocm platform to avoid hang

qishilu:
MSCCLPP_DEVICE_HIP is undefined because it is defined in device.hpp


Binyang2014:
Thanks for the contribution, could you tell us which test is hang. Then we could verify in our env

Binyang2014:
Did a simple test in our env. I think you are right.

chhwang (approved):
A better way is including `<mscclpp/device.hpp>` here but this is fine.

PR #718, opened 2026-01-07
create CI pipeline for rocm

Binyang2014:
Create CI pipeline for AMD GPU.

### DIFF
--- a/docker/build.sh
+++ b/docker/build.sh
@@ -69,13 +69,11 @@ docker build -t ${TAG_TMP} \
 
 if [[ ${TARGET} == rocm* ]]; then
     echo "Building ROCm base image..."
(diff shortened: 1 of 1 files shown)


python/requirements_rocm6.txt
```diff
@@ -0,0 +1,10 @@
+mpi4py==4.1.1
+cupy==13.6.0
```
chhwang, line 2: Does it support ROCm?

chhwang (approved):

python/requirements_rocm6.txt
Binyang2014: Yes, we just need to set some envs before pip install:
```
export CUPY_INSTALL_USE_HIP=1
export ROCM_HOME=/opt/rocm
```

PR #719, opened 2026-01-07
Use `GpuIpcMem` for NVLS connections

chhwang:
* Now `NvlsConnection` internally reuses `GpuIpcMem` for multicast memory handling.
* Removed unnecessary barriers from `connectNvlsCollective()` (CUDA API handles this automatically).
* Updated `GpuIpcMem::map()` and `GpuIpcMem::mapMulticast()` to return a shared pointer with custom deleter for unmapping, which prevents misuse of raw pointers and reduces states to be stored in the `GpuIpcMem` instance.
* Now for `RuntimeIpc` type handles, for consistency with other types, `cudaIpcOpenMemHandle` will be called in `GpuIpcMem::map()` instead of the ctor of `GpuIpcMem`.

### DIFF
--- a/.azure-pipelines/integration-test.yml
+++ b/.azure-pipelines/integration-test.yml
@@ -46,6 +46,7 @@ jobs:
       subscription:     mscclpp-ci
       vmssName:         mscclpp-ci
       sshKeySecureFile: mscclpp.pem
+      gpuArch:          '80'
 
 - job: IntegrationTestH100
   displayName: Integration test H100
@@ -66,3 +67,4 @@ jobs:
       vmssName:         mscclpp-h100-ci
       sshKeySecureFile: mscclpp.pem
       perfBaselineFile: test/deploy/perf_ndmv5.jsonl
+      gpuArch:          '90'
--- a/.azure-pipelines/templates/integration-test.yaml
+++ b/.azure-pipelines/templates/integration-test.yaml
@@ -8,7 +8,8 @@ parameters:
 - name: perfBaselineFile
   type: string
   default: 'test/deploy/perf_ndmv4.jsonl'
-
+- name: gpuArch
+  type: string
 
 steps:
 - task: Bash@3
@@ -18,7 +19,7 @@ steps:
     targetType: inline
     script: |
       mkdir build && cd build
-      cmake -DCMAKE_BUILD_TYPE=Release -DMSCCLPP_BYPASS_GPU_CHECK=ON -DMSCCLPP_USE_CUDA=ON ..
+      cmake -DCMAKE_BUILD_TYPE=Release -DMSCCLPP_BYPASS_GPU_CHECK=ON -DMSCCLPP_USE_CUDA=ON -DMSCCLPP_GPU_ARCHS=${{ parameters.gpuArch }} ..
       make -j
     workingDirectory: '$(System.DefaultWorkingDirectory)'
 
--- a/.azure-pipelines/templates/ut-no-ib-env.yaml
+++ b/.azure-pipelines/templates/ut-no-ib-env.yaml
@@ -5,6 +5,8 @@ parameters:
   type: string
 - name: sshKeySecureFile
   type: string
+- name: gpuArch
+  type: string
 
 steps:
 - task: Bash@3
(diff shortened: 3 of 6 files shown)


src/switch_channel.cc
```diff
@@ -1,230 +1,104 @@
...
-  std::copy_n(reinterpret_cast<char*>(&bufferSize_), sizeof(bufferSize_), std::back_inserter(result));
-  std::copy_n(reinterpret_cast<char*>(&minMcGran_), sizeof(minMcGran_), std::back_inserter(result));
-  std::copy_n(reinterpret_cast<char*>(&mcGran_), sizeof(mcGran_), std::back_inserter(result));
-  std::copy_n(reinterpret_cast<char*>(&rootPid_), sizeof(rootPid_), std::back_inserter(result));
-  std::copy_n(reinterpret_cast<char*>(&rootFd_), sizeof(rootFd_), std::back_inserter(result));
+  detail::serialize(result, *localGpuIpcMemHandle_);
+  detail::serialize(result, minMcGran_);
+  detail::serialize(result, numDevices_);
   return result;
 }
```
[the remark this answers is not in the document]
chhwang: Fixed

src/switch_channel.cc
```diff
@@ -1,230 +1,104 @@
...
-
-  INFO(MSCCLPP_COLL,
-       "NVLS handle created on root with size %ld. minGranularity %ld and recommendedGranularity %ld buffer size is "
-       "%ld, adjusted size is %ld",
-       mcProp_.size, minMcGran_, mcGran_, bufferSize, bufferSize_);
+  // Compute minimum granularity for user buffer alignment
+  CUmulticastObjectProp mcProp = {};
+  mcProp.size = localGpuIpcMemHandle_->baseSize;
+  mcProp.numDevices = numDevices;
+  mcProp.handleTypes = CU_MEM_HANDLE_TYPE_POSIX_FILE_DESCRIPTOR;
```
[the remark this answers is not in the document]
chhwang: Fixed

src/include/gpu_ipc_mem.hpp
```diff
@@ -77,33 +77,32 @@ static_assert(std::is_trivially_copyable_v<GpuIpcMemHandle>);
 /// GpuIpcMem represents a GPU memory region that has been imported using a GpuIpcMemHandle.
 /// If a RegisteredMemory instance represents an imported GPU memory, it will manage a unique
 /// GpuIpcMem instance for that memory region.
-class GpuIpcMem {
+class GpuIpcMem : public std::enable_shared_from_this<GpuIpcMem> {
  public:
   GpuIpcMem(const GpuIpcMemHandle &handle);
```
Binyang2014: Do we need to make the constructor be private and provide a function to create std::shared_ptr<GpuIpcMem>. Since GpuIpcMem inherits from std::enable_shared_from_this, create instance directly maybe dangerous

src/gpu_ipc_mem.cc
```diff
@@ -351,12 +298,24 @@ GpuIpcMem::~GpuIpcMem() {
...
     THROW(GPU, Error, ErrorCode::InvalidUsage, "GpuIpcMemHandle type is None, cannot map memory");
-  } else if (dataPtr_ != nullptr) {
-    // Already mapped
-    return dataPtr_;
+  }
+
+  if (type_ == GpuIpcMemHandle::Type::RuntimeIpc) {
+    // RuntimeIpc: Open handle and return shared_ptr with cleanup in deleter
+    void* basePtr = nullptr;
+    MSCCLPP_CUDATHROW(cudaIpcOpenMemHandleWrapper(&basePtr, handle_.runtimeIpc.handle));
```
Binyang2014, line 313: We can use cuda API to get basePtr from dataPtr right?

Binyang2014 (approved):

src/gpu_ipc_mem.cc
chhwang: Yes, if the caller side needs the base address, they can use `cuPointerGetAttribute` with `CU_POINTER_ATTRIBUTE_RANGE_START_ADDR` attribute.

src/include/gpu_ipc_mem.hpp
chhwang: Tackled in [b5fe18a](https://github.com/microsoft/mscclpp/pull/719/commits/b5fe18a01891ddc9fc855e2ac1e14905a6661998)

PR #722, opened 2026-01-14
Update `copilot-instructions.md`

chhwang:


.github/copilot-instructions.md
```diff
@@ -40,7 +40,7 @@ cd ..
 
 For testing after successful build:
 ```bash
-# To run all tests
+# To run tests with two GPUs - two is enough for most tests
 mpirun -np 2 ./build/test/mp_unit_tests
```
Binyang2014: We just moved the binary path to ./build/bin/mp_unit_tests

chhwang: Fixed

PR #726, opened 2026-01-15
Support multi-node in `MemoryChannel` tutorial

chhwang:


examples/tutorials/03-memory-channel/bidir_memory_channel.cu
```diff
@@ -216,23 +215,35 @@ void worker(int gpuId) {
...
+    if (status0 != 0 || status1 != 0) {
+      log("One of the processes failed.");
+      return -1;
+    }
+    log("Succeed!");
+    return 0;
+  } else if (argc == 4) {
+    std::string ipPort = argv[1];
+    int rank = std::atoi(argv[2]);
+    int gpuId = std::atoi(argv[3]);
```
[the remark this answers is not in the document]
chhwang: We will keep it simple as this is a tutorial code.

docs/tutorials/03-memory-channel.md
```diff
@@ -251,6 +251,37 @@ columns 2
 
 Since the flags take 50% of the packet size, the goodput of communication using packets is only 50% compared to transferring raw data. However, this doesn't matter because packets are designed for small data transfers. Packets transfer small data efficiently because the integrity of the user data is guaranteed by only waiting for the correct flags (done by `unpackPackets()`); explicit memory synchronization (signal and wait) is not needed.
 
+(mc-cross-node)=
+## Cross-node Execution
+
+For **inter-node** communication between GPUs on different nodes, using `PortChannel` (will be explained in the following tutorial) is usually a more accessible option that leverages more widely-used networking interfaces. However, `MemoryChannel` can still be used as long as the underlying hardware allows memory mapping between the two GPUs, such as [Multi-Node NVLink (MNNVL)](https://docs.nvidia.com/multi-node-nvlink-systems/mnnvl-user-guide/overview.html) on NVIDIA Grace Blackwell platforms.
```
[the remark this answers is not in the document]
chhwang: Fixed

docs/tutorials/03-memory-channel.md
```diff
@@ -251,6 +251,37 @@ columns 2
...
+
+We can use the same example code to test inter-node `MemoryChannel`. Before proceeding, please make sure that your environment is configured correctly for inter-node memory mapping. For example, [verify your MNNVL environment](https://docs.nvidia.com/multi-node-nvlink-systems/mnnvl-user-guide/verifying.html).
+
+Run the program on two nodes with command line arguments:
+
+```
+./bidir_memory_channel [<ip_port> <rank> <gpu_id>]
+```
+
+For example, assume we use `192.168.0.1:50000` as the bootstrap IP address and port, and both nodes use GPU 0 on their local.
```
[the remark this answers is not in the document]
chhwang: Fixed

docs/tutorials/03-memory-channel.md
```diff
@@ -251,6 +251,37 @@ columns 2
...
 
+(mc-cross-node)=
+## Cross-node Execution
+
+For **inter-node** communication, using `PortChannel` (will be explained in the following tutorial) is usually a more accessible option that leverages more widely-used networking interfaces. However, `MemoryChannel` can still be used as long as the underlying hardware allows memory mapping between the two GPUs, such as [Multi-Node NVLink (MNNVL)](https://docs.nvidia.com/multi-node-nvlink-systems/mnnvl-user-guide/overview.html) on NVIDIA Grace Blackwell platforms.
+
+We can use the same example code to test inter-node `MemoryChannel`. Before proceeding, please make sure that your environment is configured correctly for inter-node memory mapping. For example, [verify your MNNVL environment](https://docs.nvidia.com/multi-node-nvlink-systems/mnnvl-user-guide/verifying.html).
```
mahdiehghazim: Just curious do we have proper checking's in the code so if the environment does not support inter-node memorychannel, the user gets proper error message?

mahdiehghazim: pushed a commit to address this comment

examples/tutorials/03-memory-channel/bidir_memory_channel.cu
```diff
@@ -216,23 +262,37 @@ void worker(int gpuId) {
...
```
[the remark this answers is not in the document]
mahdiehghazim: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/microsoft/mscclpp/pull/726#discussion_r2743316756)

examples/tutorials/03-memory-channel/bidir_memory_channel.cu
```diff
@@ -95,21 +96,66 @@ __global__ void bidirPutPacketKernel(mscclpp::MemoryChannelDeviceHandle *devHand
   devHandle->unpackPackets(pktBufOffset, dstOffset, copyBytes, tid, blockDim.x * gridDim.x, flag);
 }
 
-void worker(int gpuId) {
+int nvlink_check(int gpuId) {
```
Binyang2014: This function just check the nvlink status, but according to this doc: https://docs.nvidia.com/multi-node-nvlink-systems/mnnvl-user-guide/verifying.html, to check if MNNVL enabled, we need to check many services includes IMEX etc...

docs/tutorials/03-memory-channel.md
Binyang2014: Maybe we can just check the error if MNVNL is not supported and throw an exception for this? To verify if current env support MNNVL is not easy

examples/tutorials/03-memory-channel/bidir_memory_channel.cu
mahdiehghazim: I just added IMEX checking's to the code and updated the document that for additional checks the user can refer to the document

examples/tutorials/03-memory-channel/bidir_memory_channel.cu
```diff
@@ -13,6 +13,10 @@
 #include <mscclpp/memory_channel_device.hpp>
 #include <sstream>
 
+#if defined(__CUDACC__) && !defined(__HIP_PLATFORM_AMD__)
```
Binyang2014: Do we need this macro, as MNNVL is only for Nvidia

examples/tutorials/03-memory-channel/bidir_memory_channel.cu
```diff
@@ -95,21 +99,103 @@ __global__ void bidirPutPacketKernel(mscclpp::MemoryChannelDeviceHandle *devHand
...
+  if (!status.empty() && status.back() == '\n') status.pop_back();
+
+  if (status != "active") {
+    log("IMEX is not active");
+    return -1;
+  }
+
+  return 0;
+}
+#endif
```
Binyang2014: I am not sure if this is a good way. As systemctl  might not available inside the container. @chhwang any thoughts here. Maybe we just skip the env check and just keep the main logic. As it's users' responsibility to make sure the env is setup properly

Binyang2014 (approved):

examples/tutorials/03-memory-channel/Makefile
```diff
@@ -12,11 +12,12 @@ endif
 
 TARGET = bidir_memory_channel
 SRC = bidir_memory_channel.cu
+COMPILE_FLAG := -I$(PWD)/../../../include/ -I$(PWD)/../../../build/include -L$(PWD)/../../../build/lib/
```
Binyang2014: No need for this. We assume user already install mscclpp and configure the include path and lib path correctly.

mahdiehghazim: @Binyang2014 Where should the users configure include and lib path? should they update the makefile? I've been following the instruction, and it didn't mention anything about configuring the paths. So, we should either update the instruction asking users to configure the paths on their own or we should properly set the paths in makefile so the user can build with a simple make command. otherwise, it'll be confusing for the users.

Binyang2014: No, user need to install mscclpp via make install. After that the headers and libs will be installed into /usr/local <default> or user specified installation path.

PR #728, opened 2026-01-20
Add a new IB stack impl that doesn't use RDMA atomics

chhwang:
* Added configurable InfiniBand (IB) signaling mode. `EndpointConfig::Ib::Mode` enum selects the mode (`Default`, `Host`, `HostNoAtomic`). `Default` is equivalent to `Host` unless specified different by envrionment `MSCCLPP_IBV_MODE`. `Host` corresponds to the previous implementation using RDMA atomics for signaling, while `HostNoAtomic` uses write-with-immediate instead.
* Regarding updates in Python bindings and API.


include/mscclpp/env.hpp
```diff
@@ -54,6 +54,12 @@ class Env {
   /// default libibverbs library found in the system.
   const std::string ibvSo;
 
+  /// Env name: `MSCCLPP_IBV_MODE`. Selects the IB stack implementation for PortChannel.
+  /// Allowed values:
+  /// - "host": use the host stack with RDMA atomics (default).
+  /// - "host-no-atomic": use the host stack with write-with-immediate signaling (no RDMA atomics).
+  const std::string ibvMode;
```
mahdiehghazim, line 61: so, we have two options for IB stack implementation. I'm curious why we have three options in core.hpp. how about setting Host and HostNoAtomic as class mode and set Host as default?

/// IB mode for signaling, used to select between different implementations.
    enum class Mode {
      Default,      // Use the MSCCLPP_IBV_MODE environment variable (or "host" if unset).
      Host,         // Use the host stack with RDMA atomics.
      HostNoAtomic  // Use the host stack with write-with-immediate signaling (no RDMA atomics).
    };

src/connection.cc
```diff
@@ -180,25 +186,127 @@ void CudaIpcConnection::flush(int64_t timeoutUsec) {
...
+    auto qp = qp_.lock();
+    for (int i = 0; i < 16; ++i) {
+      qp->stageRecv(/*wrId=*/0);
+    }
+    qp->postRecv();
+    // Start the background thread to poll recv CQ
+    recvThread_ = std::thread([this]() { this->recvThreadFunc(); });
+    INFO(CONN, "IBConnection via ", getIBDeviceName(transport_), " created with write-with-imm mode");
+  } else {
+    INFO(CONN, "IBConnection via ", getIBDeviceName(transport_), " created");
```
mahdiehghazim: we can either say "created with atomic mode" or remove this line because nothing is done here if useWriteImm_ is 0

src/core/connection.cc
```diff
@@ -180,25 +186,127 @@ void CudaIpcConnection::flush(int64_t timeoutUsec) {
 
 // IBConnection
 
+void IBConnection::recvThreadFunc() {
```
mahdiehghazim, line 184: @chhwang is this the function that you wanted to make more organized?

include/mscclpp/core.hpp
```diff
@@ -381,6 +381,13 @@ struct EndpointConfig {
   /// These settings are only used when the transport is an InfiniBand type (IB0-IB7); they are ignored for other
   /// transports.
   struct Ib {
+    /// IB mode for signaling, used to select between different implementations.
+    enum class Mode {
```
Binyang2014, line 385: Can we use ibv_query_device() to check if atomic is supported by device or not?

src/connection.cc
```diff
@@ -180,25 +186,127 @@ void CudaIpcConnection::flush(int64_t timeoutUsec) {
...
+        continue;
+      }
+
+      // The imm_data contains newValue (32-bit, extended to 64-bit)
+      // Note: getRecvWcImmData already converts from network byte order via ntohl
+      unsigned int immData = qp->getRecvWcImmData(i);
+      newValueHost = static_cast<uint64_t>(immData);
+
+      // Memory barrier to ensure we see the RDMA write data
+      std::atomic_thread_fence(std::memory_order_acquire);
```
Binyang2014: Do we need this memory fence? Since we already get ImmData from CQ, does this guarantee the data already resides in GPU memory?

src/connection.cc
```diff
@@ -180,25 +186,127 @@ void CudaIpcConnection::flush(int64_t timeoutUsec) {
...
+      // Read dstGpuAddr from the local stored address (set by setLocalInboundAddr)
+      uint64_t dstGpuAddr = localInboundAddr_;
+      if (dstGpuAddr != 0) {
+        uint64_t* dstPtr = reinterpret_cast<uint64_t*>(dstGpuAddr);
+
+        // Use cudaMemcpyAsync with our dedicated stream to avoid blocking on the default stream
+        MSCCLPP_CUDATHROW(
+            cudaMemcpyAsync(dstPtr, &newValueHost, sizeof(uint64_t), cudaMemcpyHostToDevice, signalStream_));
+        // Synchronize our stream to ensure the copy is complete before the GPU kernel reads it
+        MSCCLPP_CUDATHROW(cudaStreamSynchronize(signalStream_));
```
Binyang2014: Do we need to sync stream here? I think a GPU thread will poll this flag.

include/mscclpp/env.hpp
chhwang: If we don't have `Default` and if the user wants to say "follow whatever the env variable says", the only way is setting that manually:
```cpp
EndpointConfig cfg;
cfg.ib.mode = mscclpp::env()->ibvMode;
```
because if we leave it unset, it will be "host" regardless of the env variable.

src/core/connection.cc
```diff
@@ -248,12 +356,28 @@ void IBConnection::updateAndSync(RegisteredMemory dst, uint64_t dstOffset, uint6
...
+    // The remote's recvThreadFunc will use its stored localInboundAddr_ to write
+
+    // Put newValue in imm_data (truncated to 32-bit; semaphore counters should fit)
+    unsigned int immData = static_cast<unsigned int>(newValue);
+
+    // Send 0-byte write-with-imm; use dstMrInfo as target (we don't actually write anything)
+    qp_.lock()->stageSendWriteWithImm(nullptr, dstMrInfo,
+                                      /*size=*/0, /*wrId=*/0,
+                                      /*srcOffset=*/0, /*dstOffset=*/0,
+                                      /*signaled=*/true, /*immData=*/immData);
```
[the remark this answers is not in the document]
chhwang: Fixed

src/endpoint.cc
```diff
@@ -23,9 +26,14 @@ Endpoint::Impl::Impl(const EndpointConfig& config, Context::Impl& contextImpl)
...
     }
+
+    // If write-with-imm mode is enabled (HostNoAtomic), we need recv WRs for the recv CQ
+    bool useWriteImm = (env()->ibvMode == "host-no-atomic");
+    int maxRecvWr = useWriteImm ? 16 : 0;
+
     ibQp_ = contextImpl.getIbContext(config_.transport)
                 ->createQp(config_.ib.port, config_.ib.gidIndex, config_.ib.maxCqSize, config_.ib.maxCqPollNum,
-                           config_.ib.maxSendWr, 0, config_.ib.maxWrPerSend);
+                           config_.ib.maxSendWr, maxRecvWr, config_.ib.maxWrPerSend);
```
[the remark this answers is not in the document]
chhwang: Fixed

include/mscclpp/core.hpp
```diff
@@ -629,6 +641,11 @@ class Connection {
   /// @param timeoutUsec Timeout in microseconds. Default: -1 (no timeout)
   void flush(int64_t timeoutUsec = -1);
 
+  /// Set the local GPU address where incoming signals should write.
+  /// Must be called by the receiver before the sender calls updateAndSync.
+  /// @param addr The local GPU address for incoming writes.
+  void setLocalInboundAddr(uint64_t addr);
```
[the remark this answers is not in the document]
chhwang: Fixed

test/mp_unit/port_channel_tests.cu
```diff
@@ -477,21 +497,46 @@ void PortChannelOneToOneTest::testPacketPingPongPerf(bool useIb) {
...
   GTEST_SKIP() << "This test requires IBVerbs that the current build does not support.";
 #endif  // !defined(USE_IBVERBS)
 }
 
-TEST_F(PortChannelOneToOneTest, PacketPingPongPerf) { testPacketPingPongPerf(false); }
+TEST_F(PortChannelOneToOneTest, PingPongIbHostNoAtomicMode) {
+#if defined(USE_IBVERBS)
+  testPingPong(PingPongTestParams{
+      .useIPC = false, .useIB = true, .useEthernet = false, .waitWithPoll = false, .ibMode = IbMode::HostNoAtomic});
+#else   // !defined(USE_IBVERBS)
```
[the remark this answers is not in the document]
chhwang: Fixed

src/connection.cc
chhwang: Fixed

src/core/connection.cc
chhwang: No, it was `setLocalInboundAddr()` and I fixed it in later commits

include/mscclpp/core.hpp
chhwang: Implemented

src/connection.cc
chhwang: Fixed (removed)

src/connection.cc
chhwang: No we don't. Moved this sync to IBConnection dtor

include/mscclpp/env.hpp
mahdiehghazim: I don’t think the user needs to explicitly specify “follow whatever the env variable says.”
If the environment variable is set, we should follow it; otherwise, we can default to host e.g., 

 ibv_mode = getenv("MSCCLPP_IBV_MODE");

    if (ibv_mode == nullptr) {
        mode=host;
    } else {
        mode=ibv_mode;
    }

PR #733, opened 2026-01-23
Address comments for PR #692

Binyang2014:
Rename nanobind-exposed C++ types to Cpp*
Replace MSCCLPP_EXECUTION_PLAN_DIR / MSCCLPP_NATIVE_CACHE_DIR with MSCCLPP_CACHE_DIR across C++ and Python.


python/mscclpp/_core/compiler.py
```diff
@@ -138,7 +136,7 @@ def compile(
             )
         ).hexdigest()
 
-        plan_dir = os.environ.get("MSCCLPP_EXECUTION_PLAN_DIR", Path.home() / ".cache/mscclpp")
+        plan_dir = os.environ.get("MSCCLPP_CACHE_DIR", Path.home() / ".cache/mscclpp")
```
chhwang: We could do:
```python
from mscclpp._mscclpp import env
plan_dir = Path(env().cache_dir)
```

python/mscclpp/_core/compiler.py
```diff
@@ -226,7 +224,7 @@ def __init__(self):
             "-L" + os.path.join(self._lib_home, "lib"),
             "-lmscclpp",
         ]
-        cache_root = os.environ.get("MSCCLPP_NATIVE_CACHE_DIR", Path.home() / ".cache/mscclpp/native")
+        cache_root = os.environ.get("MSCCLPP_CACHE_DIR", Path.home() / ".cache/mscclpp" / "native")
```
chhwang: We could do:
```python
from mscclpp._mscclpp import env
plan_dir = Path(env().cache_dir) / "native"
```

src/ext/collectives/algorithm_collection_builder.cc
```diff
@@ -105,9 +105,9 @@ AlgorithmCollection AlgorithmCollectionBuilder::buildDefaultDslAlgorithms(int ra
     return oss.str();
   };
 
-  std::string planDir = env()->executionPlanDir;
+  std::string planDir = env()->cacheDir + "/default";
```
chhwang: Better to follow modern cpp style:
```cpp
std::filesystem::path cacheDir(env()->cacheDir);
auto planDir = cacheDir / "default";
```

PR #741, opened 2026-02-06
Refactor algo selection logic and introduce symmetric_memory env

Binyang2014:
This PR refactors the algorithm selection logic in MSCCL++ and introduces support for symmetric memory configuration through environment variables.

1. Algorithm Selection Refactoring
Use separate class for algo selection. Could introduce more complex logic for algo selection based on message size, arch, if cuda graph is enabled and memory allocation method

2. Symmetric Memory Support
Introduced symmetricMemory parameter in algorithm context key generation. Remove disableChannelCache env as is ambiguous 

3. Add new args for build_default_algorithms 
Add  flag_buffer, and flag_buffer_size args to build default algorithm. Then we could use unified flag buffer for different algorithms, avoid application hanging when switch algo for different message size.

### DIFF
--- a/python/mscclpp/language/internal/operations.py
+++ b/python/mscclpp/language/internal/operations.py
@@ -534,6 +534,7 @@ class PutOperation(BaseOperation):
         self.dst_buff = dst_buff
         self.channel_ids = channel_ids
         self.channel_type = channel_type
+        self.from_packet = from_packet
         self.to_packet = to_packet
         self.with_signal = with_signal
         self.with_signal_and_flush = with_signal_and_flush
@@ -575,6 +576,25 @@ class PutOperation(BaseOperation):
                 channel_ids=self.channel_ids + other.channel_ids,
                 channel_type=self.channel_type,
                 tbg_info=self.tbg_info,
+                to_packet=self.to_packet,
+                with_signal=self.with_signal,
+                with_signal_and_flush=self.with_signal_and_flush,
+            )
+        elif (
+            isinstance(other, PutOperation)
+            and self.name == Instruction.read_put_packet
+            and self.name == other.name
+            and self.src_buff == other.src_buff
+            and self.channel_type == other.channel_type
+            and self.tbg_info == other.tbg_info
+        ):
+            fused_operation = PutOperation(
+                src_buff=self.src_buff,
+                dst_buff=self.dst_buff + other.dst_buff,
+                channel_ids=self.channel_ids + other.channel_ids,
+                channel_type=self.channel_type,
+                tbg_info=self.tbg_info,
+                from_packet=self.from_packet,
                 to_packet=self.to_packet,
                 with_signal=self.with_signal,
                 with_signal_and_flush=self.with_signal_and_flush,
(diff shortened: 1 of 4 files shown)


docs/guide/mscclpp-torch-integration.md
```diff
@@ -116,8 +118,11 @@ class CustomizedComm:
         dlpack = mscclpp.RawGpuBuffer(1 << 27).to_dlpack(data_type=str(torch.float16))
         self.scratch_buffer = torch.utils.dlpack.from_dlpack(dlpack)
 
+        # Allocate flag buffer (required for packet-based algorithms)
+        self.flag_buffer = torch.ones(128, dtype=torch.uint32, device='cuda')
```
chhwang: Any reasons why we let users allocate this not doing it in our code?

src/ext/nccl/datatype_conversion.hpp
```diff
@@ -6,6 +6,7 @@
 
 #include <mscclpp/ext/nccl/nccl.h>
 
+#include <cassert>
 #include <cstddef>
```
chhwang: Please use our `THROW` macro instead of `assert`

python/csrc/core_py.cpp
```diff
@@ -44,7 +44,9 @@ void register_core(nb::module_& m) {
       .value("uint32", DataType::UINT32)
       .value("float16", DataType::FLOAT16)
       .value("float32", DataType::FLOAT32)
-      .value("bfloat16", DataType::BFLOAT16);
+      .value("bfloat16", DataType::BFLOAT16)
+      .value("float8_e4m3", DataType::FP8_E4M3)
+      .value("float8_e5m2", DataType::FP8_E5M2);
```
chhwang, line 49: `FP8` had to be `FLOAT8` to be consistent with `FLOAT16` or `FLOAT32`. We may fix in the future

chhwang (changes requested):

docs/guide/mscclpp-torch-integration.md
Binyang2014: It because we let user to customize the algorithm. So, if user want to create a new packet-based algorithm, they need to get flag buffers.
One option is we let user manage the buffer and pass to default algo and customized algo.
Another option is we can create a default flag buffer and let user to get it from our API. Then we need to manage the buffer life cycle. The buffer will be a static buffer, may could some issue when clear up the cuda runtime. (Maybe we could have a try for this way)

src/ext/nccl/datatype_conversion.hpp
Binyang2014: Fixed

docs/guide/mscclpp-torch-integration.md
Binyang2014: Okay, create a function to get default flag buffer. For default algo will use this buffer automatically

python/csrc/core_py.cpp
Binyang2014: Will fix this in another PR

PR #743, opened 2026-02-10
Removing MPI Dependency

caiomcbr:


chhwang (approved):

python/requirements_cuda11.txt
```diff
@@ -1,4 +1,5 @@
 mpi4py
+torch
```
Binyang2014: Do we need to depend on torch here?  Maybe we can remove this line.

CMakeLists.txt
```diff
@@ -47,7 +47,7 @@ list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 
 # Options
 option(MSCCLPP_ENABLE_TRACE "Enable tracing" OFF)
-option(MSCCLPP_BUILD_TESTS "Build tests" ON)
+option(MSCCLPP_BUILD_TESTS "Build tests" OFF)
```
Binyang2014, line 50: Why change this flag?

PR #745, opened 2026-02-13
Disabling Nanobind Memory Leak Warnings in Release Builds

caiomcbr:


python/csrc/core_py.cpp
```diff
@@ -306,6 +306,9 @@ void register_core(nb::module_& m) {
 }
 
 NB_MODULE(_mscclpp, m) {
+#ifdef MSCCLPP_DISABLE_NB_LEAK_WARNINGS
+  nb::set_leak_warnings(false);
+#endif
```
Binyang2014, line 312: Can we move it to the top of the file?
And maybe we can just use this:
```cpp
#if !defined(DEBUG_BUILD)
  nb::set_leak_warnings(false);
#endif
```
Then we can remove the definition of MSCCLPP_DISABLE_NB_LEAK_WARNINGS

caiocbr: Sure!

Binyang2014 (approved):

python/csrc/core_py.cpp
```diff
@@ -307,6 +307,9 @@ void register_core(nb::module_& m) {
 }
 
 NB_MODULE(_mscclpp, m) {
+#if !defined(DEBUG_BUILD)
```
Binyang2014: Thinking if it's better to add sperate option for this. Mixed with debug option may let us hard to find memory leak issue

PR #749, opened 2026-02-20
address flagBuffer ownership issue

Binyang2014:
This pull request updates the handling of the default flag buffer in the C++ and Python bindings to ensure proper memory management when interfacing with Python. 

Make sure the buffer will not be deallocated when transfer ownership from cpp to python

### DIFF
--- a/src/core/algorithm.cc
+++ b/src/core/algorithm.cc
@@ -204,18 +204,18 @@ static std::weak_ptr<void> gDefaultFlagBufferWeak;
 static size_t gDefaultFlagCount = 128;
 
 std::pair<std::shared_ptr<void>, size_t> getFlagBuffer() {
-  auto sp = gDefaultFlagBufferWeak.lock();
-  if (!sp) {
+  auto ptr = gDefaultFlagBufferWeak.lock();
+  if (!ptr) {
     if (!gDefaultFlagBuffer) {
       // Intentionally never freed — CUDA driver reclaims GPU memory at process exit.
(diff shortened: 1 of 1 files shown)


python/mscclpp/_core/algorithm.py
```diff
@@ -241,15 +241,22 @@ def register_algorithm(self, collective: str, algo_name: str, algorithm: Algorit
...
     This buffer is used internally by default algorithms to store selection flags.
     It is allocated as a shared GPU buffer and can be accessed from Python.
+    The result is cached so all callers share the same buffer.
 
     Returns:
         A CuPy array representing the flag buffer on the GPU.
     """
-    buffer_ptr, buffer_size = cpp_get_default_flag_buffer()
-    memptr = cp.cuda.MemoryPointer(cp.cuda.UnownedMemory(buffer_ptr, buffer_size, None), 0)
-    return cp.ndarray((buffer_size // 4,), dtype=cp.uint32, memptr=memptr)
```
caiocbr, line 255: I think we can keep the old code now, since we're handling it correctly in C++. Each time, we'll return the same buffer, so maybe there's no reason to keep it in global space.

Binyang2014: Change to this way because I want to make `flag_buffer` easy to maintain. 
For python side, only one instance, others just refer this instance.
For cpp side, one instance, all obj refer to this one.

Python/Cpp may use different instance for flag_buffer, but these instances point to the same raw ptr

PR #753, opened 2026-02-24
IB `host-no-atomic`: GDRCopy + mlx5dv Data Direct for memory-consistent low-latency signaling

chhwang:
Major enhancements to the IB signal forwarding mechanisms (`host-no-atomic` mode), primarily adding support for GDRCopy and MLX5 Direct Verbs, and refactoring the signal forwarding path for IB HostNoAtomic mode. The changes fix memory consistency issues and reduce signaling latency.
- GDRCopy and MLX5 Direct Verbs MR integration
- Signal forwarding path redesign
- Semaphore and connection API updates
- Environment (`MSCCLPP_FORCE_DISABLE_GDR`) and documentation updates

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -5,7 +5,7 @@
 # from the environment for the first two.
 SPHINXOPTS    ?=
 SPHINXBUILD   ?= sphinx-build
-SPHINXMULTIVERSION ?= sphinx-multiversion
+SPHINXMULTIVERSION ?= python3 build_multiversion.py
 SOURCEDIR     = .
 BUILDDIR      = _build
 
--- a/docs/build_multiversion.py
+++ b/docs/build_multiversion.py
@@ -1,0 +1,49 @@
+#!/usr/bin/env python3
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+"""Wrapper around sphinx-multiversion that patches copy_tree to generate
+_version.py in each tag checkout. This is needed because setuptools_scm
+generates _version.py at build time, but sphinx-multiversion uses
+`git archive` which only contains committed files.
+
+Usage (called by Makefile):
+    python3 build_multiversion.py <sourcedir> <outputdir> [sphinx-opts...]
+"""
+
+import os
+import re
+import subprocess
+import sys
+
+import sphinx_multiversion.git as smv_git
+from sphinx_multiversion import main as smv_main
+
+# Save the original copy_tree
+_original_copy_tree = smv_git.copy_tree
+
+
+def _patched_copy_tree(gitroot, src, dst, reference, sourcepath="."):
+    """Call original copy_tree, then generate _version.py from the VERSION file."""
+    _original_copy_tree(gitroot, src, dst, reference, sourcepath)
+
+    # Extract version from the tag name (e.g., "v0.9.0" -> "0.9.0")
+    refname = getattr(reference, "refname", "") or ""
+    match = re.search(r"v(\d+\.\d+\.\d+)", refname)
+    if not match:
+        return
+
+    version = match.group(1)
+    version_py_dir = os.path.join(dst, "python", "mscclpp")
+    if os.path.isdir(version_py_dir):
+        version_py = os.path.join(version_py_dir, "_version.py")
+        if not os.path.exists(version_py):
+            with open(version_py, "w") as f:
+                f.write(f'__version__ = "{version}"\n')
+
+
+# Monkey-patch
+smv_git.copy_tree = _patched_copy_tree
+
+if __name__ == "__main__":
+    sys.exit(smv_main(sys.argv[1:]))
--- a/python/mscclpp/language/channel.py
+++ b/python/mscclpp/language/channel.py
@@ -140,7 +140,7 @@ class MemoryChannel:
 
         for tb_id in tb_list:
             tb_chunk_id = get_program().setup_remote_chunk(self.src_rank, tb_id, remote_chunk, self.channel_type)
-            tb_channel_ids = get_program().setup_channel(tb, self)
+            tb_channel_ids = get_program().setup_channel(tb_id, self)
             op = GetOperation(
                 src_buff=[RemoteChunk(src_chunk.buffer, src_chunk.index, src_chunk.size, tb_chunk_id)],
                 dst_buff=[LocalChunk(dst_chunk.buffer, dst_chunk.index, dst_chunk.size)],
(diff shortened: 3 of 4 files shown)


src/core/connection.cc
```diff
@@ -220,22 +219,32 @@ void IBConnection::recvThreadFunc() {
...
-      // The imm_data contains newValue (32-bit, extended to 64-bit)
-      // Note: getRecvWcImmData already converts from network byte order via ntohl
-      unsigned int immData = qp->getRecvWcImmData(i);
-      newValueHost = static_cast<uint64_t>(immData);
+      // Read the token value written by the remote sender.
+#if defined(DEBUG_CUFLUSH) && defined(MSCCLPP_USE_CUDA)
+      // cuFlush path: read from imm_data then flush NIC->GPU write pipeline for visibility.
+      newValueHost = static_cast<uint64_t>(qp->getRecvWcImmData(i));
+      MSCCLPP_CUTHROW(cuFlushGPUDirectRDMAWrites(CU_FLUSH_GPU_DIRECT_RDMA_WRITES_TARGET_CURRENT_CTX,
+                                                 CU_FLUSH_GPU_DIRECT_RDMA_WRITES_TO_OWNER));
```
Binyang2014: Do we need to keep this code here?

src/core/connection.cc
```diff
@@ -220,22 +219,32 @@ void IBConnection::recvThreadFunc() {
...
-        MSCCLPP_CUDATHROW(
-            cudaMemcpyAsync(dstPtr, &newValueHost, sizeof(uint64_t), cudaMemcpyHostToDevice, signalStream_));
-
-        INFO(CONN, "IBConnection recvThreadFunc: updated GPU ptr ", dstPtr, " to ", newValueHost, " (immData=", immData,
-             ")");
+        if (remoteUpdateDstAddrMap_ && remoteUpdateDstAddrMap_->valid()) {
+          // Direct host-side write to GPU memory via GDRCopy BAR1 mapping
+          remoteUpdateDstAddrMap_->copyTo(&newValueHost, sizeof(uint64_t));
+        } else {
+          *dstPtr = newValueHost;
```
Binyang2014: Is this valid for CUDA? Maybe we can throw error if the dstAddrMap is invalid for cuda env

src/core/connection.cc
```diff
@@ -220,22 +219,32 @@ void IBConnection::recvThreadFunc() {
...
+      // Read the 64-bit token from the local signal GPU buffer via volatile load.
+      // localSignalGpuPtr_ points to either a GDRCopy BAR1 mapping (CUDA) or the
+      // GPU buffer directly (ROCm system-coherent/uncached memory). volatile is not
+      // strictly needed here (uncacheable memory and intervening function calls prevent
+      // stale reads), but is kept as a convention for NIC-written memory.
+      newValueHost = *static_cast<volatile uint64_t*>(localSignalGpuPtr_);
+#endif
 
       // Read dstGpuAddr from the local stored address (set by setRemoteUpdateDstAddr)
       uint64_t dstGpuAddr = remoteUpdateDstAddr_;
```
Binyang2014: A bit confused about this var. If we use host2hostSemaphore, this addr is host addr?

src/core/ib.cc
```diff
@@ -84,9 +87,16 @@ IbMr::IbMr(ibv_pd* pd, void* buff, std::size_t size) : mr_(nullptr), buff_(buff)
     MSCCLPP_CUTHROW(cuMemGetHandleForAddressRange(&fd, addr, pages * pageSize, CU_MEM_RANGE_HANDLE_TYPE_DMA_BUF_FD, 0));
```
Binyang2014: We need to use `cuMemGetHandleForAddressRange(&fd, addr, pages * pageSize, CU_MEM_RANGE_HANDLE_TYPE_DMA_BUF_FD, CU_MEM_RANGE_FLAG_DMA_BUF_MAPPING_TYPE_PCIE)` to get max performance

src/core/semaphore.cc
```diff
@@ -177,15 +208,16 @@ MSCCLPP_API_CPP void Host2HostSemaphore::wait(int64_t maxSpinCount) {
...
          (*expectedInboundToken_)) {
     if (maxSpinCount >= 0 && spinCount++ == maxSpinCount) {
-      throw Error("Host2HostSemaphore::wait timed out", ErrorCode::Timeout);
+      THROW(CONN, Error, ErrorCode::Timeout, "Host2HostSemaphore::wait timed out");
     }
   }
 }
 
 MSCCLPP_API_CPP MemoryDevice2DeviceSemaphore::MemoryDevice2DeviceSemaphore(const Semaphore& semaphore)
     : semaphore_(semaphore), expectedInboundToken_(detail::gpuCallocUnique<uint64_t>()) {
```
Binyang2014, line 217: Maybe we can remove the outboundToken_ here. We update remote token directly

src/core/connection.cc
```diff
@@ -250,60 +256,104 @@ IBConnection::IBConnection(std::shared_ptr<Context> context, const Endpoint& loc
...
+             ibCtx->getDevName(),
+             " is not a VF. "
+             "If you experience data corruption, use VF devices with Data Direct or IbMode::Host.");
+      }
+    }
 
-    // Pre-post receive requests for incoming write-with-imm
+    // Pre-post receive requests for incoming WRITE_WITH_IMM notifications.
+    // The recv CQE guarantees the preceding data WRITE has been committed to GPU memory.
     auto qp = qp_.lock();
```
Binyang2014, line 311: For non-atomic mode. We use `detail::gpuCallocShared` to allocate the signal buffer. In this case the GPU buffer is not dataDirectEnabled, will it fallback to slow path?

src/core/connection.cc
```diff
@@ -250,60 +256,104 @@ IBConnection::IBConnection(std::shared_ptr<Context> context, const Endpoint& loc
...
       qp->stageRecv(/*wrId=*/0);
     }
     qp->postRecv();
-    // Start the background thread to poll recv CQ
-    recvThread_ = std::thread([this]() { this->recvThreadFunc(); });
-    INFO(CONN, "IBConnection via ", getIBDeviceName(transport_), " created with no-atomic mode");
+    // The recv thread is started later in startSignalForwarding() when the semaphore
+    // provides the signal forwarding destination. This ensures the thread lifetime is
+    // bounded by the GdrMap lifetime (created before start, destroyed after stop).
+    INFO(CONN, "IBConnection via ", getIBDeviceName(transport_), " created with signal forwarding (HostNoAtomic) mode");
```
Binyang2014, line 320: Have a bit concern for this. In our test, we usally set 4 QPs per to provide  good resilience. If so we will create too many threads

src/core/semaphore.cc
chhwang: Isn't it already removed in main?

src/core/connection.cc
chhwang: Will address in another PR

src/core/ib.cc
chhwang: Fixed

PR #758, opened 2026-02-28
Adding Support to Setting Message Size Range in Native Algorithm API

caiomcbr:


python/csrc/algorithm.cpp
```diff
@@ -60,6 +60,16 @@ void register_algorithm(nb::module_& m) {
           .def_prop_ro("name", &Algorithm::name)
           .def_prop_ro("collective", &Algorithm::collective)
           .def_prop_ro("message_range", &Algorithm::messageRange)
+          .def(
+              "set_message_range",
+              [](Algorithm& self, size_t minMessageSize, size_t maxMessageSize) {
+                auto* native = dynamic_cast<NativeAlgorithm*>(&self);
+                if (!native) {
+                  throw nb::type_error("set_message_range is only supported for native algorithms");
+                }
```
Binyang2014: We can check this in python side

python/mscclpp/_core/algorithm.py
```diff
@@ -114,11 +114,22 @@ def collective(self) -> str:
...
         return (self._algorithm.message_range[0], self._algorithm.message_range[1])
 
+    def set_message_size_range(self, min_message_size: int, max_message_size: int):
+        """Set the valid message size range in bytes.
+
+        Args:
+            min_message_size: Minimum supported message size in bytes.
+            max_message_size: Maximum supported message size in bytes.
+
+        Only supported for native algorithms. Raises TypeError for DSL algorithms.
```
Binyang2014, line 129: We can check the type here

Binyang2014 (approved):

PR #776, opened 2026-04-08
Fix missing directory of document for new tag v0.9.0 

seagater:
The v0.9.0 conf.py (introduced in #775) dynamically loads the version from python/mscclpp/_version.py.

This file is generated at build time by setuptools_scm and is listed in .gitignore — it is never committed to the repo. Earlier tags (v0.8.0 and below) used a hardcoded release (e.g., "v0.8.0") in conf.py, so they had no dependency on generated files.
sphinx-multiversion checks out each tag using git archive, which only extracts committed files. 
Since _version.py is not committed, the v0.9.0 checkout is missing it, and conf.py crashes on import. All future tags will have this same problem.

**Three changes:**
1. docs/build_multiversion.py (new): A wrapper around sphinx-multiversion that monkey-patches copy_tree to generate _version.py in each tag checkout after extraction. The version string is parsed from the tag name (e.g., v0.9.0 → __version__ = "0.9.0").
2. Makefile: The multiversion target now calls build_multiversion.py instead of sphinx-multiversion directly.
3. conf.py: Added a fallback so that if _version.py doesn't exist, it reads the version from the VERSION file instead. This makes conf.py resilient for any future scenario where _version.py is missing.

**Testing**
Verified locally:
• make multiversion now successfully builds all 11 versions (v0.4.0 through v0.9.0)
• v0.9.0 docs are correctly generated under _build/html/v0.9.0/
Version selector shows v0.9.0 as latest

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -5,7 +5,7 @@
 # from the environment for the first two.
 SPHINXOPTS    ?=
 SPHINXBUILD   ?= sphinx-build
-SPHINXMULTIVERSION ?= sphinx-multiversion
+SPHINXMULTIVERSION ?= python3 build_multiversion.py
 SOURCEDIR     = .
 BUILDDIR      = _build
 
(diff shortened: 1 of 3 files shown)


docs/conf.py
```diff
@@ -30,10 +30,16 @@ def __getattr__(self, name):
...
+_version_file = python_path / "mscclpp" / "_version.py"
+if _version_file.exists():
+    spec = importlib.util.spec_from_file_location("_version", _version_file)
+    version_module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(version_module)
+    version = version_module.__version__
+else:
+    # Fallback: read VERSION file (always present in the repo)
+    version_file = project_root / "VERSION"
+    version = version_file.read_text().strip() if version_file.exists() else "0.0.0"
```
Binyang2014: Maybe we don't need to change this, we will generate the __version.py based on tag

docs/Makefile
Binyang2014: Maybe set SPHINXMULTIVERSION=python3 build_multiversion.py here?

docs/conf.py
seagater: Resolved. Revert the code changes to use _version.py.

docs/Makefile
seagater: Updated.

Binyang2014 (approved):

PR #781, opened 2026-04-10
Fix multi-node H100 CI: CUDA compat, deploy improvements

Binyang2014:
## Summary

- **Multi-node H100 CI setup**: Improve architecture detection and GPU configuration
- **Remove hardcoded VMSS hostnames** from deploy files
- **Fix CUDA compat library issue**: Remove stale compat paths from Docker image for CUDA 12+. Instead, `peer_access_test` now returns a distinct exit code (2) for CUDA init failure, and `setup.sh` conditionally adds compat libs only when needed. This fixes `cudaErrorSystemNotReady` (error 803) when the host driver is newer than the container's compat libs.
- **Speed up deploy**: Replace recursive `parallel-scp` with tar+scp+untar to avoid per-file SSH overhead.


src/core/registered_memory.cc
```diff
@@ -159,10 +159,34 @@ RegisteredMemory::Impl::Impl(const std::vector<char>::const_iterator& begin,
...
+        try {
+          auto gpuIpcMem = GpuIpcMem::create(entry.gpuIpcMemHandle);
+          this->remoteMemMap = gpuIpcMem->map();
+          this->data = this->remoteMemMap.get();
+        } catch (const Error& e) {
+          if (!hasIB) {
+            throw Error("Cross-node Fabric import failed and no IB transport available: " + std::string(e.what()),
+                        ErrorCode::InvalidUsage);
+          }
+          INFO(GPU, "Cross-node Fabric import failed, falling back to IB transport");
```
chhwang: Why would CudaIpc transport fall back to IB?

src/core/registered_memory.cc
```diff
@@ -159,10 +159,34 @@ RegisteredMemory::Impl::Impl(const std::vector<char>::const_iterator& begin,
...
+      // Cross-node memory: CudaIpc only works via Fabric (requires IMEX daemon).
+      // PosixFd uses unix domain sockets which are node-local.
+      // If Fabric import fails or is unavailable, fall back to IB transport.
+      bool hasFabric = (entry.gpuIpcMemHandle.typeFlags & GpuIpcMemHandle::Type::Fabric) != 0;
+      bool hasIB = (transports & AllIBTransports).any();
+      if (hasFabric) {
+        try {
+          auto gpuIpcMem = GpuIpcMem::create(entry.gpuIpcMemHandle);
+          this->remoteMemMap = gpuIpcMem->map();
+          this->data = this->remoteMemMap.get();
```
chhwang: Please remove redundant patterns

src/core/registered_memory.cc
```diff
@@ -159,10 +159,34 @@ RegisteredMemory::Impl::Impl(const std::vector<char>::const_iterator& begin,
...
+        } catch (const Error& e) {
```
chhwang: Exactly which errors are we catching here?

chhwang (changes requested):

src/core/registered_memory.cc
Binyang2014: Because we support this: `transports = mscclpp::Transport::CudaIpc | mscclpp::Transport::IB0`. In this case, we need try to import handle with cudaIpc first, if failed try to import via IB

src/core/registered_memory.cc
Binyang2014: Failed to import via fabric due to MNNVL not supported

chhwang: I mean, can we specify that here? We may catch CudaError and identify using the error code. Also if we will catch all related errors here, we don't need to check `hasFabric` ahead.

src/core/registered_memory.cc
chhwang: Makes sense, but then the log message is inconsistent here. If `transports = mscclpp::Transport::CudaIpc | mscclpp::Transport::IB0` and it is deserialized on a different host that cannot support CudaIpc, it will print this fallback message. However, if it is deserialized on the same host that cannot support CudaIpc, it won't print any messages.

src/core/registered_memory.cc
Binyang2014: Just refactor this piece of code. The error could be system error such as file does not exist if using unix socket for handle sharing or cuda error for fabric handle sharing. So just catch `BaseError` here

src/core/registered_memory.cc
Binyang2014: Thanks, fixed

chhwang (approved):

PR #783, opened 2026-04-11
Support TBG inside Pipeline

caiocbr:
This PR adds support for linked thread block groups in pipelines, allowing multiple thread blocks to cooperate on processing each pipeline slice and increasing the available parallelism within pipeline iterations.

image

As shown in the image, each pipeline iteration processes a different slice of data, with each thread block within the threadblock group managing a portion of that slice. The final slice in the pipeline may differ in size from the others, and in such cases, the thread block groups will adjust accordingly to handle it.


src/core/executor/execution_plan.cc
```diff
@@ -537,14 +537,21 @@ void ExecutionPlan::Impl::setupOperation(const nlohmann::json& op, Operation& op
       }
       size_t inputOffset = this->getOffset(this->inputSize, this->outputSize, buff["index"], bufferType) + constOffset;
       size_t inputBufferSize = this->getBufferSize(this->inputSize, this->outputSize, buff["index"], buff["size"]);
-      inputOffset += calcOffset(inputBufferSize, tbId, tbgSize);
-      inputBufferSize = calcSize(inputBufferSize, tbId, tbgSize);
+      inputOffset += calcOffset(inputBufferSize, 0, 1);
+      inputBufferSize = calcSize(inputBufferSize, 0, 1);
```
Binyang2014, line 541: What's the meaning of 0 and 1 here. Maybe give a meaningful name?

src/core/executor/execution_plan.cc
```diff
@@ -567,8 +574,8 @@ void ExecutionPlan::Impl::setupOperation(const nlohmann::json& op, Operation& op
       }
       size_t outputOffset = this->getOffset(this->inputSize, this->outputSize, buff["index"], bufferType) + constOffset;
       size_t outputBufferSize = this->getBufferSize(this->inputSize, this->outputSize, buff["index"], buff["size"]);
-      outputOffset += calcOffset(outputBufferSize, tbId, tbgSize);
-      outputBufferSize = calcSize(outputBufferSize, tbId, tbgSize);
+      outputOffset += calcOffset(outputBufferSize, 0, 1);
+      outputBufferSize = calcSize(outputBufferSize, 0, 1);
```
Binyang2014, line 578: Same here

src/core/include/execution_kernel.hpp
```diff
@@ -66,6 +66,19 @@ MSCCLPP_DEVICE_INLINE uint32_t getOffset(BufferType bufferType, uint32_t offset)
   }
 }
 
+MSCCLPP_DEVICE_INLINE uint32_t calcOffset(uint32_t size, uint32_t index, uint32_t slices) {
+  constexpr uint32_t alignment = 16;
```
Binyang2014, line 70: Do we need alignment here?

caiocbr: I think so, because I need to calculate this when using tbg, so each tb will need to use this. Although I need to gather this information from the execution plan, so let me adjust this.

PR #784, opened 2026-04-13
Add `accumulate` method for `PortChannel`

chhwang:
## Summary

Adds `Connection::accumulate()` and `PortChannel::accumulate()` for remote signed 64-bit addition. Callers provide only their contribution, allowing concurrent writers to share a counter.

The operation uses the `TriggerAccumulate` FIFO opcode. Its operand occupies the trigger's first word. FIFO readiness comes from lap-parity commit bits, so zero-valued operands are valid.

## Transport support

| Transport | Support |
|---|---|
| IB | NIC atomic fetch-and-add; rejected in no-atomic mode |
| Ethernet | Receive-side read-modify-write, serialized across connections |
| CudaIpc on ROCm | Atomic kernel on the connection stream |
| CudaIpc on CUDA | Rejected; no safe, practical proxy-side RMW |

## Validation and ordering

- Accumulate targets must contain a naturally aligned 8-byte word at the requested offset.
- Bounds checks are overflow-safe and run before issuing transport work.
- Accumulate is asynchronous. A later signal or flush on the same connection preserves the connection's operation ordering.
- Packed trigger fields reuse the existing debug-only checked constructor without adding release-mode validation branches.

## Tests

Coverage includes:

- positive, negative, and zero operands;
- concurrent and multi-rank fan-in;
- accumulate-before-signal/flush ordering;
- unsupported CUDA CudaIpc and IB no-atomic modes;
- undersized, exact-end, partial-word, huge-offset, and misaligned targets;
- collective capability/topology gating for IB and CudaIpc tests.

Follow-up validation for the current head:

- 8× A100, CUDA 12.9: consolidated one-to-one accumulate **4/4 passed**; fan-in **2/2 passed**.
- 8× MI300X, ROCm 6.2.4: consolidated accumulate **7/7 passed**; full 8-rank non-IB aggregate **30 passed / 8 expected skips**.

The original implementation was also validated on H200/CUDA 12.9 and MI300X/ROCm 7.2.


include/mscclpp/fifo_device.hpp
```diff
@@ -15,10 +15,27 @@
...
-constexpr TriggerType TriggerFlag = 0x2;  // Trigger a signaling.
-constexpr TriggerType TriggerSync = 0x4;  // Trigger a flush.
+constexpr TriggerType TriggerNone = 0;                   // Not an operation; invalid for ProxyService.
+constexpr TriggerType TriggerPut = 1;                    // Transfer data.
+constexpr TriggerType TriggerSignal = 2;                 // Signal the remote semaphore.
+constexpr TriggerType TriggerPutWithSignal = 3;          // Transfer data, then signal.
+constexpr TriggerType TriggerFlush = 4;                  // Flush the connection.
+constexpr TriggerType TriggerAccumulate = 6;             // Add a value to remote memory.
+constexpr TriggerType TriggerPutWithSignalAndFlush = 7;  // Transfer data, signal, then flush.
+// 5 is unassigned.
```
Binyang2014, line 32: Do we need to reserve 5 here? You mean user may build customized algo with old API?

src/core/accumulate_kernel.cu
```diff
@@ -0,0 +1,33 @@
...
+  setStreamIfNeeded();
+  // Submit to this connection's stream, which orders the add ahead of any signal or flush that
+  // follows. On ROCm a kernel runs while the caller's kernel occupies the GPU, so the proxy does
+  // not wait for the caller.
+  accumulateU64Kernel<<<1, 1, 0, *stream_>>>(dst, value);
+  MSCCLPP_CUDATHROW(cudaGetLastError());
+  dirty_ = true;
+}
+
+}  // namespace mscclpp
```
Binyang2014: Do we need launch a kernel here or we can access peer GPU directly in CPU side?

src/core/accumulate_kernel.cu
```diff
@@ -0,0 +1,33 @@
...
+
+void CudaIpcStream::accumulate(uint64_t* dst, uint64_t value) {
+  CudaDeviceGuard deviceGuard(deviceId_);
+  setStreamIfNeeded();
+  // Submit to this connection's stream, which orders the add ahead of any signal or flush that
+  // follows. On ROCm a kernel runs while the caller's kernel occupies the GPU, so the proxy does
+  // not wait for the caller.
+  accumulateU64Kernel<<<1, 1, 0, *stream_>>>(dst, value);
+  MSCCLPP_CUDATHROW(cudaGetLastError());
+  dirty_ = true;
```
Binyang2014, line 28: Why need to set dirty here? It will be reset some where?

src/core/connection.cc
```diff
@@ -759,14 +868,29 @@ void EthernetConnection::recvMessages() {
...
-      mscclpp::gpuMemcpy(ptr + (recvSize / sizeof(char)), recvBuffer_.data(), messageSize, cudaMemcpyHostToDevice);
-      recvSize += messageSize;
+    if (isAccumulate && size == sizeof(uint64_t)) {
+      // Accumulate modulo 2^64: receive the operand, then read, add, and write back.
+      uint64_t addValue;
+      if (!receiveFramePart(&addValue, sizeof(uint64_t), false)) return;
+
+      // Every peer terminates its socket in this process, so several recv threads can be here
+      // at once for one address. The read-modify-write below is not atomic, so serialize it
+      // against the other recv threads.
```
Binyang2014, line 878: Why there will be several threads here. For each device we just launch one recv thread, no?

include/mscclpp/fifo_device.hpp
chhwang: It's just unused so unassigned, but probably we'd better reserve for a future operation. Will do

Binyang2014 (approved):

chhwang:
@copilot resolve the merge conflicts in this pull request

Binyang2014 (approved):

PR #786, opened 2026-04-14
Fix nccl-test CI building for all GPU architectures

Binyang2014:
## Problem

`nccl-test.yml` was the only CI template calling `deploy.yml` without passing `gpuArch`. Since the CI build machine has no GPU, CMake fell back to building for **all** supported architectures (`80;90;100;120`), unnecessarily slowing down CI builds.

## Fix

- Add `gpuArch` parameter to `nccl-test.yml` and forward it to `deploy.yml`
- Pass `gpuArch: '80'` (A100) and `gpuArch: '90'` (H100) from `nccl-api-test.yml`

All other templates were already passing `gpuArch` correctly.

### DIFF
--- a/.azure-pipelines/nccl-api-test.yml
+++ b/.azure-pipelines/nccl-api-test.yml
@@ -44,6 +44,7 @@ jobs:
     parameters:
       subscription:     mscclpp-ci
       vmssName:         mscclpp-ci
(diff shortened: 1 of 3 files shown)


caiomcbr (approved):
LGTM

PR #798, opened 2026-05-06
Introduce GPUBufferPool for symmetric memory support

Binyang2014:
## Summary

Enables application workloads to use MSCCL++ symmetric memory through deterministic GPU buffer sub-allocation. Applications can reserve one communication-friendly `GpuBuffer` slab per rank and allocate aligned buffers at matching offsets by making the same allocation calls on every rank.

Pooled allocations reuse the existing `GpuBuffer` API and remain valid across Python and DLPack consumers until the last reference is released. This lets application-managed input, output, and workspace buffers participate in symmetric-memory NVL/NVLS collectives without introducing a separate buffer type.

## Changes

- Add a C++ `GpuBufferPool` with deterministic aligned allocation, block reuse and coalescing, thread-safe bookkeeping, and storage lifetime tied to returned `GpuBuffer` instances.
- Add the Python `GpuBufferPool` API and preserve pooled-buffer ownership across DLPack tensor conversion.
- Document pooled-buffer semantics: returned memory may contain data from a previous use, and buffers must remain alive until all GPU work using them completes.
- Add C++ and Python coverage for alignment, reuse, accounting, cross-rank matching offsets, and DLPack lifetime.
- Represent FP8 NVLS multimem support as an explicit `0`/`1` compile-time capability and use value-based checks consistently.


Binyang2014:
@copilot+gpt-5.5 help me fix lint and rocm build issue

python/mscclpp/__init__.py
```diff
@@ -84,6 +86,8 @@
     "ExecutionPlan",
     "PacketType",
     "RawGpuBuffer",
+    "RawGpuBufferPool",
+    "RawGpuBufferPoolBuffer",
```
Binyang2014: Why we expose interal data here?

python/csrc/gpu_utils_py.cpp
```diff
@@ -52,65 +61,77 @@ static DLDataType getDlType(std::string type) {
...
+  return makeDlpack(buffer.data(), buffer.nelems(), buffer.deviceId(), std::move(owner), dataType, shape, strides);
+}
+
+static nb::capsule toDlpack(std::shared_ptr<GpuBufferPool::Buffer> buffer, std::string dataType,
+                            std::vector<int64_t>& shape, std::vector<int64_t>& strides) {
+  void* data = buffer->data();
+  size_t bytes = buffer->bytes();
+  int deviceId = buffer->deviceId();
+  return makeDlpack(data, bytes, deviceId, std::move(buffer), dataType, shape, strides);
+}
```
Binyang2014: Why need two functions with same name?

python/csrc/gpu_utils_py.cpp
```diff
@@ -52,65 +61,77 @@ static DLDataType getDlType(std::string type) {
   }
 }
 
-static nb::capsule toDlpack(GpuBuffer<char> buffer, std::string dataType, std::vector<int64_t>& shape,
-                            std::vector<int64_t>& strides) {
+static void dlpackDeleter(DLManagedTensor* self) { delete static_cast<DlpackContext*>(self->manager_ctx); }
```
Binyang2014: Can we just make it as a lambda?

include/mscclpp/gpu_utils.hpp
```diff
@@ -403,6 +404,96 @@ class GpuBuffer {
...
+/// relying on normal Python reference counting to return buffers to the local pool.
+///
+class GpuBufferPool {
+ public:
+  /// Represents one buffer returned by `GpuBufferPool`.
+  ///
+  /// This object owns a sub-range of a pool allocation. Destroying the last reference to this object returns the
+  /// sub-range to the pool. Destruction only updates local pool bookkeeping and does not perform any
+  /// cross-rank synchronization.
+  class Buffer {
```
Binyang2014: Can we resuse the API for GPU buffer, not create a new class. Maybe need a field to indicate from GPUBufferPool or not

python/mscclpp/_core/buffer.py
```diff
@@ -35,3 +39,65 @@ def __new__(
...
+
+    @property
+    def device_id(self) -> int:
+        """CUDA/HIP device id of the pool allocation."""
+        return self._pool.device_id()
+
+    def allocate(
+        self,
+        nbytes: int,
+        alignment: int = 256,
```
Binyang2014, line 88: Why use 256 here?

python/csrc/gpu_utils_py.cpp
```diff
@@ -15,6 +17,13 @@ using namespace mscclpp;
 
 constexpr int BYTE_BITS = 8;
 
+struct DlpackContext {
```
Empyreus, line 20: Is there value in creating this struct just for makeDlpack instead of just temp variables within the function?

Empyreus (approved):

PR #804, opened 2026-05-15
Clean up completed communicator receives

Binyang2014:
## Summary
- Release the reference after last requests are ready.
- Keep ordered receive chaining for repeated rank/tag operations while cleaning up completed receive bookkeeping.

### DIFF
--- a/src/core/communicator.cc
+++ b/src/core/communicator.cc
@@ -3,9 +3,60 @@
 
 #include "communicator.hpp"
 
+#include <utility>
+
 #include "api.h"
 
 namespace mscclpp {
+
+namespace {
+
+template <typename Fn>
+class ScopeGuard {
+ public:
+  explicit ScopeGuard(Fn fn) : fn_(std::move(fn)) {}
+  ScopeGuard(const ScopeGuard&) = delete;
+  ScopeGuard& operator=(const ScopeGuard&) = delete;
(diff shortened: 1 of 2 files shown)


src/core/communicator.cc
```diff
@@ -3,10 +3,61 @@
...
+                               if (item && it != impl->lastRecvItems_.end() && it->second == item) {
+                                 impl->lastRecvItems_.erase(it);
+                               }
+                             });
+
+                             if (lastRecvItem) {
+                               // Recursive call to the previous receive items
+                               lastRecvItem->wait();
+                             }
+                             return func();
```
chhwang: Can't we just do cleanup here directly like `auto ret = func(); cleanup(); return ret;`, instead of setting a callback?

Binyang2014: Thanks, fixed

src/core/communicator.cc
```diff
@@ -3,10 +3,42 @@
 
 #include "communicator.hpp"
 
+#include <utility>
```
chhwang: Do we need this?

chhwang (approved):

src/core/communicator.cc
Binyang2014: For `std::move`, but seems this already be included in other header

chhwang (approved):

PR #807, opened 2026-05-19
add nBlocks check for allreduce_allpair_packet algo

Binyang2014:
- Fix the correctness issue for allreduce_allpair_packet algo. Make sure no overwrite for input buffer. Use same tb for send/reduce/write-back.
- Check if nBlocks/nthreads validate for packet algorithm.
- Add more logs
- Modify flag update logic, make it work for the case:  nthreadPerNBlock < nflags


mahdiehghazim (approved):
LGTM

mahdiehghazim (approved):

PR #810, opened 2026-05-20
GB200 support: SendRecv DSL collective and per-channel executor connections

Binyang2014:
## Summary
 
 GB200 support work: introduces point-to-point send/receive in the MSCCL++ DSL
 and extends the executor for split-NVL-domain topologies where some ranks are
 NVL-connected within a node and other ranks must communicate across the network.
 
 ### DSL
 - New `SendRecv` collective with separate input/output buffers
   (`python/mscclpp/language/collectives.py`).
 - New multi-node sendrecv DSL example
   (`python/mscclpp/language/tests/multi_node/send_recv.py`) with `--split_mask`
   (group size − 1) and `--instances` CLI options. Documents the channel-ordering
   trick that keeps signal tags cross-matched between paired peers when
   `prev == next`.
 - `BaseBuffer.__getitem__` now accepts slices with `None` start/stop
   (e.g., `buf[:]`).
 
 ### Executor
 - One connection (unique QP) per channel entry instead of one per peer.
   Required for HostNoAtomic IB mode where each QP can forward signals to a
   single semaphore. Uses per-peer tag counters so paired ranks agree on tag
   ordering regardless of the order peers appear in each rank's `connected_to`
   list.
 - MEMORY channels now unconditionally use `Transport::CudaIpc`; only PORT
   channels can use IB. Matches the invariant already enforced by
   `getTransportFlags`.
 - `ExecutionContext::connections` is now a `vector<Connection>` indexed by
   channel order (was `unordered_map<int, Connection>` keyed by peer). Removes
   redundant semaphore fields from `ExecutionContext`.
 - TODO: explicit NVL-domain check in `useIB`


src/core/include/execution_kernel.hpp
```diff
@@ -174,11 +174,11 @@ MSCCLPP_DEVICE_INLINE void handlePut(const Operation& op, void* input, void* out
       uint32_t dstOffset =
           dstOffsets[tid] + getOffset<ReuseScratch>(portChannelBufferTypes_[op.outputBufferRefs[tid].id], offset);
       uint32_t srcOffset = srcOffsets[tid] + getOffset<ReuseScratch>(op.inputBufferRefs[tid].type, offset);
-      if constexpr (PutWithSignal) {
-        portChannels_[channelIndexes[tid]].putWithSignal(dstMemoryId, dstOffset, srcMemoryId, srcOffset, size);
-      } else if constexpr (PutWithSignalAndFlush) {
+      if constexpr (PutWithSignalAndFlush) {
```
mahdiehghazim, line 177: any reason we changed the order of if statements here?

mahdiehghazim (approved):

src/core/include/execution_kernel.hpp
Binyang2014: Seems the logic is same. Maybe copilot reformat the code. I think we can keep this change here

PR #811, opened 2026-05-22
Add check to filter invalid nblock/nthread candidates

Binyang2014:
Add check for invalid nblock/nthread candidate


mahdiehghazim (approved):
LGTM

PR #815, opened 2026-06-04
Expose NVLS multicast granularity option for GpuBuffer

Binyang2014:
Add a public Granularity enum (MultiCastMinimum, MultiCastRecommended) and let GpuBuffer choose the NVLS multicast allocation granularity via a constructor argument, defaulting to MultiCastMinimum to minimize memory usage. Expose the same option through the C++ and Python (nanobind) APIs.


mahdiehghazim (approved):
LGTM

PR #817, opened 2026-06-08
Add support for allgather nvls algorithms

Empyreus:
Add support for testing with NVSL by creating MSCCLPP_FORCE_DISABLE_IB to disable IB for NVLS runs.

Add needed functions for allgather implementation

Implement initial allgather nvls algorithm. 

Results with allgather_nvls_zero_copy.py:

image


python/mscclpp/language/internal/operations.py
```diff
@@ -926,11 +926,12 @@ def shift_buffers(self, instance, num_instances, replication_function):
...
-        result["buffer_type"] = self.buffer_type.value
-        result["buffer_offset"] = self.buffer_offset
-        result["size"] = self.size
-        result["channel_ids"] = self.channel_ids
+        result["src_buff"] = [
+            {"type": self.src_chunk.buffer.value, "index": self.src_chunk.index, "size": self.size}
+        ]
+        result["dst_buff"] = [
+            {"switch_channel_id": self.channel_ids[0], "index": self.buffer_offset, "size": self.size}
+        ]
```
Binyang2014: @caiomcbr does this change make sense to you?

src/core/executor/executor.cc
```diff
@@ -290,7 +292,7 @@ struct Executor::Impl {
     context.localMemoryIdBegin = context.proxyService->nextMemoryId(3);
     for (auto& bufferType : {BufferType::INPUT, BufferType::OUTPUT, BufferType::SCRATCH}) {
       TransportFlags flags = Transport::CudaIpc;
-      if (hasIBDevices()) flags |= IBs[rank % this->nranksPerNode];
+      if (hasIBDevices() && !env()->forceDisableIb) flags |= IBs[rank % this->nranksPerNode];
```
Binyang2014: Maybe we can change to if (useIB()) ...

src/core/include/execution_kernel.hpp
```diff
@@ -575,6 +575,58 @@ MSCCLPP_DEVICE_INLINE void handleMultiLoadReduceStore(const Operation& op, uint3
...
+    }
+    done += 8;
+  }
+  if (size - done >= 4) {
+    if (threadIdx.x == 0) {
+      u32x1* src4 = reinterpret_cast<u32x1*>(srcBase + done);
+      u32x1* dst4 = reinterpret_cast<u32x1*>(dstBase + done);
+      SwitchChannelDeviceHandle::multimemStore(src4[0], dst4);
+    }
+  }
```
Binyang2014: Can we use different thread to send remainder at once?

src/core/include/execution_kernel.hpp
```diff
@@ -575,6 +575,58 @@ MSCCLPP_DEVICE_INLINE void handleMultiLoadReduceStore(const Operation& op, uint3
...
+  assert(reinterpret_cast<uintptr_t>(dstBase) % 16 == 0);
+  const size_t n16 = size / 16;
+  f32x4* src16 = reinterpret_cast<f32x4*>(srcBase);
+  f32x4* dst16 = reinterpret_cast<f32x4*>(dstBase);
+  for (size_t idx = threadIdx.x; idx < n16; idx += blockDim.x) {
+    SwitchChannelDeviceHandle::multimemStore(src16[idx], dst16 + idx);
+  }
+
+  // Remainder (size is a multiple of 4, so at most 12 bytes left): one 8-byte then one 4-byte store.
+  size_t done = n16 * 16;
```
Binyang2014: Maybe change to another name for `done` and `n16`? Also we can use `n16 << 4` instead of  `*16`

python/mscclpp/language/internal/operations.py
Empyreus: My understanding is that the original didn't have information needed for allgather, such as offsets, sizes.

src/core/include/execution_kernel.hpp
Empyreus: I updated the variable names, but I think at least for me '* 16' feels cleaner and doesn't affect performance since the compiler should optimize this out. If you still want it changed, I can move to << 4 instead.

src/core/executor/executor.cc
```diff
@@ -290,7 +292,7 @@ struct Executor::Impl {
     context.localMemoryIdBegin = context.proxyService->nextMemoryId(3);
     for (auto& bufferType : {BufferType::INPUT, BufferType::OUTPUT, BufferType::SCRATCH}) {
       TransportFlags flags = Transport::CudaIpc;
-      if (hasIBDevices()) flags |= IBs[rank % this->nranksPerNode];
+      if (hasIBDevices() && mscclpp::env()->forceDisableIb) flags |= IBs[rank % this->nranksPerNode];
```
Binyang2014: `hasIBDevices() && !mscclpp::env()->forceDisableIb`?

Binyang2014 (approved):
LGTM, pls address copilot PR first

Empyreus:
results
Updated Results

python/mscclpp/language/internal/operations.py
```diff
@@ -926,11 +926,42 @@ def shift_buffers(self, instance, num_instances, replication_function):
 
     def to_dict(self):
         result = {"name": self.name.value}
-        result["src_chunk"] = self.src_chunk.to_dict()
-        result["buffer_type"] = self.buffer_type.value
-        result["buffer_offset"] = self.buffer_offset
-        result["size"] = self.size
-        result["channel_ids"] = self.channel_ids
+        result["src_buff"] = [{"type": self.src_chunk.buffer.value, "index": self.src_chunk.index, "size": self.size}]
```
caiocbr: You can use directly `chunk.to_dict()`, like:
```
result["src_buff"] = []
result["src_buff"].append(self.src_chunk.to_dict())
```

python/mscclpp/language/internal/operations.py
```diff
@@ -926,11 +926,42 @@ def shift_buffers(self, instance, num_instances, replication_function):
...
+        result["dst_buff"] = [
+            {"switch_channel_id": self.channel_ids[0], "index": self.buffer_offset, "size": self.size}
+        ]
+        result["channel_type"] = self.channel_type.value
+        return result
+
+
+class GroupStorePacket(BaseOperation):
```
caiocbr: I think maybe here we can follow the design of the other instructions. Instead of splitting in GroupStore and GroupStorePacket keep only GroupStore and add some parameter like use_packet to change the behavior for pkt.

python/mscclpp/language/channel.py
```diff
@@ -959,6 +959,51 @@ def broadcast(self, rank, src_chunk: Chunk, buffer_offset, size, tb):
         op = GroupStore(src_chunk, self.buffer_type, buffer_offset, size, tb_channel_ids, self.channel_type)
         get_program().add_operation(self.src_rank, tb, op)
 
+    def broadcast_packets(self, rank, src_chunk: Chunk, buffer_offset, size, tb):
+        """Broadcast packet-formatted data from source chunk to all ranks in the switch channel.
+
+        Packet variant of :meth:`broadcast`. Emits a ``gstorepkt`` (MULTI_STORE_PKT)
+        operation that multicasts LL-protocol packets (data + flag) from the source
+        chunk to the specified buffer region across all ranks in the rank group, with
+        no explicit barrier required (the packet flag provides synchronization).
```
caiocbr: You might want to add a comment here noting that this reads the data to ensure it's ready. There could be other implementations where reading the data isn't necessary, so it may be helpful to mention that in the description.

python/mscclpp/language/channel.py
```diff
@@ -959,6 +959,51 @@ def broadcast(self, rank, src_chunk: Chunk, buffer_offset, size, tb):
         op = GroupStore(src_chunk, self.buffer_type, buffer_offset, size, tb_channel_ids, self.channel_type)
         get_program().add_operation(self.src_rank, tb, op)
 
+    def broadcast_packets(self, rank, src_chunk: Chunk, buffer_offset, size, tb):
+        """Broadcast packet-formatted data from source chunk to all ranks in the switch channel.
+
+        Packet variant of :meth:`broadcast`. Emits a ``gstorepkt`` (MULTI_STORE_PKT)
```
caiocbr: Here, we can make explicit the use of the scratch buffer for the packet approach. You can refer `read_put_packets` description.

python/mscclpp/language/channel.py
```diff
@@ -959,6 +959,51 @@ def broadcast(self, rank, src_chunk: Chunk, buffer_offset, size, tb):
...
+        no explicit barrier required (the packet flag provides synchronization).
+
+        Args:
+            rank (int): The rank that will execute this broadcast operation.
+            src_chunk (Chunk): The source chunk containing packet data to broadcast.
+            buffer_offset (int): The offset in the destination buffer where data will be stored.
+            size (int): The size of data to broadcast.
+            tb (int): The thread block ID that will execute this operation.
+
+        Raises:
```
caiocbr, line 976: In this situation, we need to make sure the scratch buffer is being used for both the source and destination ranks.

python/mscclpp/language/channel.py
```diff
@@ -959,6 +959,51 @@ def broadcast(self, rank, src_chunk: Chunk, buffer_offset, size, tb):
...
+            >>> channel.broadcast_packets(rank=0, src_chunk=chunk, buffer_offset=0, size=1, tb=0)
+        """
+        self.src_rank = rank
+        if src_chunk.rank not in self.rank_group.ranks:
+            raise RuntimeError(
+                f"Destination chunk rank {src_chunk.rank} is not part of the rank group {self.rank_group.ranks}."
+            )
+        if src_chunk.size != size:
+            raise RuntimeError(f"Destination chunk size {src_chunk.size} does not match the required size {size}.")
+
```
caiocbr, line 995: Could you add a check to verify if the buffer type is scratch?

python/mscclpp/language/channel.py
```diff
@@ -959,6 +959,51 @@ def broadcast(self, rank, src_chunk: Chunk, buffer_offset, size, tb):
...
+        for rank in self.rank_group.ranks:
+            if self.buffer_type == BufferType.scratch:
+                buffer_size = get_program().gpus[rank].scratch_chunks
+            else:
```
caiocbr: We only should have buffer type scratch in this operation

python/mscclpp/language/tests/single_node/allgather_nvls_zero_copy_pkt.py
```diff
@@ -0,0 +1,85 @@
...
+    # scratch via the switch (gstorepkt / MULTI_STORE_PKT), and unpacks locally.
+    chunksperloop = 1
+    collective = AllGather(gpu_size, chunksperloop, True)
+    with CollectiveProgram(
+        name,
+        collective,
+        gpu_size,
+        instances=instances,
+        protocol="LL",
+        num_threads_per_block=num_threads_per_block,
```
caiocbr, line 31: You might want to include `auto_sync=False` to make sure it doesn't add any unwanted sync.

src/core/include/execution_kernel.hpp
```diff
@@ -575,6 +575,79 @@ MSCCLPP_DEVICE_INLINE void handleMultiLoadReduceStore(const Operation& op, uint3
...
+  // any type conversion, so the data type is irrelevant here -- we move raw bytes using the widest
+  // available multimem store unit (16 -> 8 -> 4 bytes). This keeps the op fully type agnostic (works for
+  // any dtype, including uint8_t and FP8, on any arch that supports MULTI_STORE).
+  const uint32_t size = min(op.inputBufferSizes[0] - offset, unitSize);
+  if (size == 0) {
+    return;
+  }
+  const uint32_t srcOffset = op.inputOffsets[0] + getOffset<ReuseScratch>(op.inputBufferRefs[0].type, offset);
+  const uint32_t dstOffset = op.outputOffsets[0] + getOffset<ReuseScratch>(op.nvlsOutputBufferType, offset);
+  // `multimem.st` has a 4-byte minimum granularity, so the moved region must be 4-byte aligned and sized.
```
caiocbr: I think if we have some alignment requirement for this function would be better add this check at DSL level.

python/mscclpp/language/channel.py
```diff
@@ -959,6 +959,54 @@ def broadcast(self, rank, src_chunk: Chunk, buffer_offset, size, tb):
         op = GroupStore(src_chunk, self.buffer_type, buffer_offset, size, tb_channel_ids, self.channel_type)
         get_program().add_operation(self.src_rank, tb, op)
 
+    def broadcast_packets(self, rank, src_chunk: Chunk, buffer_offset, size, tb):
+        """Broadcast data in packet format from the source chunk to all ranks' scratch buffers in the switch channel.
```
caiocbr, line 963: Fix typo

caiocbr (approved):
LGTM

src/core/include/execution_kernel.hpp
```diff
@@ -575,6 +575,65 @@ MSCCLPP_DEVICE_INLINE void handleMultiLoadReduceStore(const Operation& op, uint3
...
+    return;
+  }
+
+  const uint32_t srcOffset = op.inputOffsets[0] + getOffset<ReuseScratch>(op.inputBufferRefs[0].type, offset);
+  const uint32_t dstOffset = op.outputOffsets[0] + getOffset<ReuseScratch>(op.nvlsOutputBufferType, offset);
+  char* srcBase = static_cast<char*>(getBuffer(input, output, scratch, op.inputBufferRefs[0].type)) + srcOffset;
+  char* dstBase = reinterpret_cast<char*>(nvlsChannels_[op.nvlsOutputIndex].mcPtr) + dstOffset;
+
+  size_t processed = 0;
+  const bool alignedf32x4 = (reinterpret_cast<uintptr_t>(srcBase) % sizeof(f32x4) == 0) &&
```
Binyang2014, line 593: Change this to assert? If baseaddr not align with f32x4, seems there will be some correctness issue

python/mscclpp/language/tests/single_node/allgather_nvls_zero_copy_pkt.py
Binyang2014, line 1: I think we can move the algo to default folder, same as nvls_zero_copy. Then user could consume these algo directly. We can add these in other pr

Binyang2014 (approved):

PR #819, opened 2026-06-17
support rocm7.2

Binyang2014:
This pull request introduces support for ROCm 7.2 across the build system, CI pipelines, Docker images, and documentation, while also improving ROCm FP8 type selection and CUDA IPC memory handle management. It updates dependencies and configurations to ensure compatibility with ROCm 7.2, adds new options for native FP8 variants, and refines some benchmarking and internal memory handling logic.

Pls notice: there is an issue in rocm7.2 (rocm7.2 user lib + rocm6.2 driver) when execution code in this order: allocating memory -> ipc communication -> allocate new memory -> free old memory.


include/mscclpp/gpu_data_types.hpp
```diff
@@ -16,16 +16,16 @@ using __bfloat16 = __hip_bfloat16;
...
+// MSCCLPP_ROCM_USE_FNUZ_FP8 option controls whether native FP8 aliases use FNUZ.
 #if defined(HIP_VERSION_MAJOR) && (HIP_VERSION_MAJOR >= 6)
 #include <hip/hip_fp8.h>
 
 // Create aliases matching CUDA naming convention for cross-platform compatibility.
 // Define __FP8_E4M3_IS_FNUZ__ / __FP8_E5M2_IS_FNUZ__ when the platform-native FP8 is the
 // "fnuz" variant (no infinities, NaN-only at 0x80, bias differs from OCP). Dispatch layers
 // use these macros to throw on unsupported variants requested via DataType.
-#if (HIP_VERSION_MAJOR == 6) || (HIP_VERSION_MAJOR > 6 && HIP_FP8_TYPE_FNUZ && !HIP_FP8_TYPE_OCP)
+#if defined(MSCCLPP_ROCM_FP8_FNUZ) || (HIP_FP8_TYPE_FNUZ && !HIP_FP8_TYPE_OCP)
```
mahdiehghazim: if MSCCLPP_ROCM_FP8_FNUZ is disabled in cmake and "(HIP_FP8_TYPE_FNUZ && !HIP_FP8_TYPE_OCP)" is true, should we give a warning to the user that fp8_fnuz is being used?

CMakeLists.txt
```diff
@@ -183,11 +184,20 @@ elseif(MSCCLPP_USE_CUDA)
 	    endif()
     endif()
 elseif(MSCCLPP_USE_ROCM)
-    set(MSCCLPP_GPU_ARCHS gfx90a gfx941 gfx942)
```
mahdiehghazim, line 186: any reason we removed gfx941?

include/mscclpp/gpu_data_types.hpp
mahdiehghazim: nvm, i see it falls back to prints hip default in cmakelists.txt

python/mscclpp_benchmark/bench_collective.py
```diff
@@ -607,6 +606,9 @@ def main(argv: list[str] | None = None) -> None:
             )
             if comm.rank == 0:
                 print(".", end="", flush=True)
+            # TODO: remove this after rocm7.2 bug is fixed
```
mahdiehghazim: should we guard it so it's only enabled with rocm7.2?

src/core/gpu_ipc_mem.cc
```diff
@@ -97,43 +95,56 @@ struct CudaIpcMemHandleEqual {
...
 
-static std::unordered_map<cudaIpcMemHandle_t, void*, CudaIpcMemHandleHash, CudaIpcMemHandleEqual>
+static std::unordered_map<cudaIpcMemHandle_t, std::weak_ptr<void>, CudaIpcMemHandleHash, CudaIpcMemHandleEqual>
     openCudaIpcMemHandleMap;
 
 static std::mutex openCudaIpcMemHandleMapMutex;
 
-// Cache open ipc handles to avoid opening multiple times (ROCm may exceed system limit on vm.max_map_count).
-static inline cudaError_t cudaIpcOpenMemHandleWrapper(void** addr, cudaIpcMemHandle_t ipcHandle) {
+static inline cudaError_t cudaIpcOpenMemHandleWrapper(std::shared_ptr<void>& basePtr, cudaIpcMemHandle_t ipcHandle) {
```
mahdiehghazim, line 103: are the changes in this file a workaround for rocm 7 bug or they are more general?

CMakeLists.txt
Binyang2014: gfx941 is the early version of MI300X, not production ready. We don't need to support it

src/core/gpu_ipc_mem.cc
Binyang2014: more general. Previous our memory handle management is not correct. In some case we may close a handle which still has reference

src/ext/collectives/allreduce/allreduce_fullmesh.cu
```diff
@@ -220,31 +209,34 @@ CommResult AllreduceFullmesh::allreduceKernelFunc(
...
                                                                bool symmetricMemory) {
-  static int tag = 0;
-  size_t recvBytes;
-  CUdeviceptr recvBasePtr;
-  MSCCLPP_CUTHROW(cuMemGetAddressRange(&recvBasePtr, &recvBytes, (CUdeviceptr)output));
   symmetricMemory_ = symmetricMemory;
   if (!symmetricMemory_) {
-    return AlgorithmCtxKey{nullptr, (void*)recvBasePtr, 0, recvBytes, tag++};
+    return AlgorithmCtxKey{const_cast<void*>(input), output, size, size, 0};
   }
```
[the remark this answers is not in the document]
Binyang2014: That expected behavior, to make sure context reset will reset all handle. Otherwise we will keep some internal structure for input buffer, bad for handle management

python/mscclpp_benchmark/bench_collective.py
Binyang2014: Thx, fixed

include/mscclpp/gpu_data_types.hpp
Binyang2014: Yeah, we will print a message for this

CMakeLists.txt
```diff
@@ -192,9 +192,9 @@ message(STATUS "GPU architectures: ${MSCCLPP_GPU_ARCHS}")
 if(MSCCLPP_USE_ROCM)
     if(MSCCLPP_ROCM_USE_FNUZ_FP8)
         add_compile_definitions(MSCCLPP_ROCM_FP8_FNUZ)
-        message(STATUS "ROCm native FP8 variant: fnuz")
+        message(STATUS "ROCm native FP8 aliases: FNUZ (MSCCLPP_ROCM_USE_FNUZ_FP8=ON)")
     else()
-        message(STATUS "ROCm native FP8 variant: HIP default")
+        message(STATUS "ROCm native FP8 aliases: OCP (MSCCLPP_ROCM_USE_FNUZ_FP8=OFF)")
```
mahdiehghazim, line 197: so you changed it to rely on MSCCLPP_ROCM_USE_FNUZ_FP8 only and ignore HIP default. i think it makes it more transparent for the user. is there any other reason we do not rely on HIP default?

mahdiehghazim (approved):

CMakeLists.txt
Binyang2014: Because `HIP_FP8_TYPE_OCP` only works for device code, in our host code we also need to know the supported fp8 type. So just sue `MSCCLPP_ROCM_USE_FNUZ_FP8` to control it. Otherwise it may cause some confusion

PR #822, opened 2026-06-24
Migrate to C++20 and drop CUDA 11 support

Empyreus:
This PR bumps the project's C++ language standard from C++17 to C++20 (host, CUDA, and HIP) and removes support for CUDA 11, which cannot compile C++20. CI matrices are updated to drop CUDA 11.8 and add CUDA 13.0 alongside 12.9.

 - C++20 enables newer language features across the codebase.
 - C++20 requires CUDA ≥ 12.0, so CUDA 11.8 can no longer be supported.
 - Blackwell / sm_100 targets require CUDA ≥ 12.8, so CUDA 13.x is added to CI coverage.


.devcontainer/devcontainer_amd.json
```diff
@@ -32,7 +32,7 @@
                     "/usr/include"
                 ],
                 "C_Cpp.default.cStandard": "c17",
```
Binyang2014, line 34: Do we need to change this?

CMakeLists.txt
```diff
@@ -189,10 +189,11 @@ endif()
 message(STATUS "GPU architectures: ${MSCCLPP_GPU_ARCHS}")
 
 # Declare project
-set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra")
 if(MSCCLPP_USE_CUDA)
-    set(CMAKE_CUDA_STANDARD 17)
+    set(CMAKE_CUDA_STANDARD 20)
+    set(CMAKE_CUDA_EXTENSIONS OFF)
```
Binyang2014, line 206: Why need this line?

Empyreus: Since we're now above nvcc's default of C++17, CMake passes an explicit -std flag for the CUDA language. With CMAKE_CUDA_EXTENSIONS left at its default (ON), that flag is -std=gnu++20, which nvcc rejects (it only accepts the c++NN dialects, not the gnu++NN variants). Setting CMAKE_CUDA_EXTENSIONS OFF makes CMake emit -std=c++20 instead, which nvcc accepts. This only affects CUDA compilation, host C++ still keeps GNU extensions and since we don't rely on any GNU dialect features in the CUDA code, turning it off has no downside.

Binyang2014 (approved):
LGTM

Binyang2014 (approved):

PR #826, opened 2026-06-30
Support FP8 NVLS multimem reduction

Binyang2014:
## Summary
- add FP16 accumulation support for FP8 NVLS multimem load-reduce
- probe and gate FP8 NVLS algorithms on device support
- keep MNNVL algorithm sizing and buffer-pool changes out of this sub-PR

## Validation
- ./tools/lint.sh
- cmake --build build --target mscclpp_collectives mscclpp_nccl -j$(nproc)


caiocbr (approved):
This PR adds FP16 accumulation for the multimem.ld_reduce path when the data type is FP8 (via the optional AccumT template parameter on SwitchChannelDeviceHandle::multimemLoadReduce and the dispatchFp8Accum helper), and threads an accumDtype parameter (default DataType::AUTO) through Algorithm::execute.

What this PR intentionally leaves for a future PR is wiring the accumulation type into algorithm selection. Today the accumulation type only influences the kernel dispatch after an algorithm has already been chosen; the selector itself does not know about it. Follow-up work should close that gap:

Carry accumDtype in the request. Add an accumDtype field to CollectiveRequest and thread it from the public/NCCL entry point (where it is currently defaulted to AUTO) down into selection, instead of only into execute().

Let algorithms declare accumulation-type support. Extend the Algorithm interface (e.g. a supportsAccumDtype(dtype, accumDtype) predicate, a capability set, or a new field in Constraint/tags) so each algorithm advertises which (dtype, accumDtype) combinations it can handle. For example, the NVLS/multimem path can support FP8 data with FP16 or FP32 accumulation, while packet/other paths may only support native accumulation.

Filter/prefer during selection. Update matchExecutionPlan and the native selectors (selectSingleNodeAllreduceBlackwell / any AlgoSelectFunc) to only pick algorithms that support the requested accumDtype, with a well-defined fallback (e.g. fall back to AUTO/native accumulation) when no candidate supports it.

Approving as-is since the accumulation plumbing is self-contained and correct; the selection changes above can be handled in a follow-up.

PR #843, opened 2026-07-16
Add switch-channel NVLS cross-rank barrier

Empyreus:
Adds a device-side cross-rank barrier built on NVLS (multimem) atomics

### DIFF
--- a/include/mscclpp/switch_channel.hpp
+++ b/include/mscclpp/switch_channel.hpp
@@ -4,11 +4,10 @@
 #ifndef MSCCLPP_SWITCH_CHANNEL_HPP_
 #define MSCCLPP_SWITCH_CHANNEL_HPP_
 
-#include <mscclpp/gpu_utils.hpp>
-#include <mscclpp/switch_channel_device.hpp>
-
 #include <cstdint>
 #include <memory>
+#include <mscclpp/gpu_utils.hpp>
+#include <mscclpp/switch_channel_device.hpp>
 
 namespace mscclpp {
 
--- a/include/mscclpp/switch_channel.hpp
+++ b/include/mscclpp/switch_channel.hpp
@@ -7,6 +7,9 @@
 #include <mscclpp/gpu_utils.hpp>
 #include <mscclpp/switch_channel_device.hpp>
 
+#include <cstdint>
+#include <memory>
+
 namespace mscclpp {
 
 class NvlsConnection;
@@ -16,6 +19,12 @@ struct SwitchChannel {
   void* devicePtr_;
   std::shared_ptr<void> mcPtr_;
   size_t bufferSize_;
+  // Barrier state inherited from the owning NvlsConnection (see NvlsConnection::bindAllocatedMemory).
+  // All are null / zero if the connection was created without barrier support.
+  uint32_t* barrierLocalFlag_ = nullptr;
+  uint32_t* barrierMcFlag_ = nullptr;
+  uint32_t* barrierGen_ = nullptr;
+  int barrierNRanks_ = 0;
 
  public:
   using DeviceHandle = SwitchChannelDeviceHandle;
@@ -41,9 +50,30 @@ class NvlsConnection {
   /// @return SwitchChannel with devicePtr, mcPtr and bufferSize
   SwitchChannel bindAllocatedMemory(CUdeviceptr devicePtr, size_t size);
 
+  /// Attach a device-side barrier resource shared by all SwitchChannels created from this
+  /// connection. After this call, `SwitchChannel::deviceHandle().barrier()` can synchronize all
+  /// ranks in the multicast group without a separate mesh of memory-channel semaphores. This is set
+  /// up automatically by `connectNvlsCollective`; it is an internal setup hook and is not intended
+  /// to be called directly.
+  /// @param barrierConn Auxiliary NVLS connection backing the barrier flag (kept alive).
+  /// @param barrierBuffer Storage for the barrier flag (kept alive); element 0 is the shared arrival
+  /// counter, element 1 is this rank's generation counter.
+  /// @param barrierChannel The bound barrier channel (kept alive for its multicast pointer).
+  /// @param nRanks Number of ranks participating in the multicast group.
+  void attachBarrier(std::shared_ptr<NvlsConnection> barrierConn, std::shared_ptr<void> barrierBuffer,
+                     std::shared_ptr<SwitchChannel> barrierChannel, int nRanks);
+
  private:
   class Impl;
   std::shared_ptr<Impl> pimpl_;
+
+  // Barrier resources, owned by this connection and shared by every SwitchChannel it creates.
+  std::shared_ptr<NvlsConnection> barrierConn_;
+  std::shared_ptr<void> barrierBuffer_;
+  std::shared_ptr<SwitchChannel> barrierChannel_;
+  uint32_t* barrierLocalFlag_ = nullptr;
+  uint32_t* barrierMcFlag_ = nullptr;
+  int barrierNRanks_ = 0;
 };
 
 class Communicator;
--- a/include/mscclpp/switch_channel.hpp
+++ b/include/mscclpp/switch_channel.hpp
@@ -20,9 +20,9 @@ struct SwitchChannel {
   size_t bufferSize_;
   // Barrier state inherited from the owning NvlsConnection (see NvlsConnection::bindAllocatedMemory).
   // All are null / zero if the connection was created without barrier support.
-  uint32_t* barrierLocalFlag_ = nullptr;
-  uint32_t* barrierMcFlag_ = nullptr;
-  uint32_t* barrierGen_ = nullptr;
+  uint64_t* barrierLocalFlag_ = nullptr;
+  uint64_t* barrierMcFlag_ = nullptr;
+  uint64_t* barrierGen_ = nullptr;
   int barrierNRanks_ = 0;
 
  public:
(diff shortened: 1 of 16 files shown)


include/mscclpp/switch_channel.hpp
```diff
@@ -16,6 +18,12 @@ struct SwitchChannel {
   void* devicePtr_;
   std::shared_ptr<void> mcPtr_;
   size_t bufferSize_;
+  // Barrier state inherited from the owning NvlsConnection (see NvlsConnection::bindAllocatedMemory).
+  // All are null / zero if the connection was created without barrier support.
+  uint32_t* barrierLocalFlag_ = nullptr;
+  uint32_t* barrierMcFlag_ = nullptr;
+  uint32_t* barrierGen_ = nullptr;
+  int barrierNRanks_ = 0;
```
Binyang2014: Do you think it's better to have a switchChannelSemaphore to store these fields? SwitchChannel only hold the semaphore obj, same as memoryChannel

include/mscclpp/switch_channel_device.hpp
```diff
@@ -25,8 +27,76 @@ struct SwitchChannelDeviceHandle {
...
+      asm volatile("multimem.red.release.sys.add.u32 [%0], %1;" ::"l"(mcBarrierFlag), "r"(1U) : "memory");
+    }
+
+    cuda::memory_order waitOrder =
+        (memoryOrder == cuda::memory_order::relaxed) ? cuda::memory_order::relaxed : cuda::memory_order::acquire;
+    // Wait until every rank has arrived. The signed (wrap-safe) compare means "counter is behind target".
+    POLL_MAYBE_JAILBREAK(
+        (static_cast<int32_t>(atomicLoad<uint32_t, scopeSystem>(localBarrierFlag, waitOrder) - target) < 0),
+        maxSpinCount);
+  }
```
Binyang2014: How about split this API into signal and wait, same as the other channel.

python/mscclpp/language/internal/buffer_access.py
```diff
@@ -19,7 +19,11 @@ def __init__(self):
     def process_operations(self, operations):
         result_operations = []
         for operation in operations:
-            if operation.name == Instruction.nop or operation.name == Instruction.barrier:
+            if (
+                operation.name == Instruction.nop
+                or operation.name == Instruction.barrier
+                or operation.name == Instruction.group_barrier
```
Binyang2014: Also here change to switchChannel.signal() swithcChannel.wait(), not barrier API

include/mscclpp/switch_channel_device.hpp
chhwang, line 1: Please add C++ unit tests for new channel methods.

include/mscclpp/switch_channel.hpp
chhwang: +1

include/mscclpp/switch_channel.hpp
```diff
@@ -16,6 +18,12 @@ struct SwitchChannel {
   void* devicePtr_;
   std::shared_ptr<void> mcPtr_;
   size_t bufferSize_;
+  // Barrier state inherited from the owning NvlsConnection (see NvlsConnection::bindAllocatedMemory).
+  // All are null / zero if the connection was created without barrier support.
+  uint32_t* barrierLocalFlag_ = nullptr;
```
chhwang: 32-bit flag is smaller than our other semaphores that use 64-bit flags. Is this overflow-safe?

src/core/switch_channel.cc
```diff
@@ -249,18 +295,25 @@ MSCCLPP_API_CPP std::shared_ptr<NvlsConnection> connectNvlsCollective(std::share
...
-    std::vector<char> data;
-    bootstrap->recv(data, rootRank, 0);
-    conn = std::make_shared<NvlsConnection>(data);
-  }
+  auto conn = makeNvlsConnection(bootstrap, allRanks, rank, rootRank, isRoot, bufferSize, 0);
+
+#if (CUDA_NVLS_API_AVAILABLE)
+  // Set up a small auxiliary multicast that backs a device-side barrier for this connection. Using a
+  // dedicated connection (rather than carving from the user buffer) keeps the user buffer's sizing
+  // untouched. Flag layout: element 0 is the shared arrival counter, element 1 is this rank's
```
chhwang: I don't like this idea. Why not appending the flag size to the requested buffer size?

src/core/switch_channel.cc
```diff
@@ -220,19 +220,65 @@ std::vector<char> NvlsConnection::serialize() { return pimpl_->serialize(); }
...
 
 SwitchChannel::DeviceHandle SwitchChannel::deviceHandle() const {
   SwitchChannel::DeviceHandle device;
   device.devicePtr = devicePtr_;
   device.mcPtr = mcPtr_.get();
   device.bufferSize = bufferSize_;
+  device.mcBarrierFlag = barrierMcFlag_;
+  device.localBarrierFlag = barrierLocalFlag_;
+  device.barrierGen = barrierGen_;
+  device.nRanks = barrierNRanks_;
```
chhwang: `barrierNRanks_` looks duplicated with `NvlsConnection::Impl::numDevices_`.

src/core/switch_channel.cc
```diff
@@ -249,18 +295,25 @@ MSCCLPP_API_CPP std::shared_ptr<NvlsConnection> connectNvlsCollective(std::share
...
+  auto barrierBuffer = std::make_shared<GpuBuffer<uint32_t>>(2);
+  size_t barrierBytes = barrierBuffer->bytes();
+  auto barrierConn = makeNvlsConnection(bootstrap, allRanks, rank, rootRank, isRoot, barrierBytes, 1);
+  auto barrierChannel = std::make_shared<SwitchChannel>(
+      barrierConn->bindAllocatedMemory(CUdeviceptr(barrierBuffer->data()), barrierBytes));
+  conn->attachBarrier(barrierConn, barrierBuffer->memory(), barrierChannel, static_cast<int>(allRanks.size()));
+  // One-time sync so every rank has bound its barrier flag into the multicast before any rank issues
+  // a multimem operation on it at kernel time. Synchronize only allRanks (not the whole bootstrap,
+  // which may include ranks outside this NVLS group).
+  bootstrap->groupBarrier(allRanks);
```
chhwang: I don't think we need to wait for flag binding here, since `cuMemMap` will block until all [`cuMulticastAddDevice`](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MULTICAST.html#group__CUDA__MULTICAST_1g2df07b38ff506c519e9f799c5ddf7e5d) are done from all devices, which basically separates this concern.

chhwang (changes requested):

include/mscclpp/switch_channel.hpp
Empyreus: This implementation is overflow-safe, overflowing is expected in our wait method.

src/core/switch_channel.cc
Empyreus: I am reviewing this and my understanding is that the request buffer size for multicast regions is fixed, so in order to request extra we cannot do flag size but must request the full chunk again. We can do this but it will require us to double the multicast region memory usage for every switch_channel.

include/mscclpp/switch_channel.hpp
```diff
@@ -41,9 +49,28 @@ class NvlsConnection {
...
   class Impl;
   std::shared_ptr<Impl> pimpl_;
+
+  // Barrier resources, owned by this connection and shared by every SwitchChannel it creates. The
+  // flag is carved from this connection's own multicast region, so no auxiliary connection is kept.
+  std::shared_ptr<void> barrierBuffer_;
+  std::shared_ptr<SwitchChannel> barrierChannel_;
+  uint64_t* barrierLocalFlag_ = nullptr;
+  uint64_t* barrierMcFlag_ = nullptr;
+  int barrierNRanks_ = 0;
```
Binyang2014: Can we use semaphore object here

src/core/include/execution_kernel.hpp
```diff
@@ -634,6 +634,45 @@ MSCCLPP_DEVICE_INLINE void handleMultiStorePkt(const Operation& op, void* input,
...
+//   if (blockIdx.x == 0 && threadIdx.x == 0) {
+//     nvlsChannels_[op.nvlsInputIndex].barrier();
+//   }
+//   syncer->sync(op.nThreadBlocks);
+// }
+
+template <bool Relaxed = false>
+MSCCLPP_DEVICE_INLINE void handleMultiSignal(const Operation& op) {
+  DeviceSyncer* syncer = &deviceSyncers[op.deviceSyncerIndex];
+  syncer->sync(op.nThreadBlocks);
```
Binyang2014: Why we sync all thread blocks here? Leave this to user? Users need to call barrier explicitly.

src/core/include/execution_kernel.hpp
```diff
@@ -634,6 +634,45 @@ MSCCLPP_DEVICE_INLINE void handleMultiStorePkt(const Operation& op, void* input,
...
+      nvlsChannels_[op.nvlsInputIndex].relaxedSignal();
+    } else {
+      nvlsChannels_[op.nvlsInputIndex].signal();
+    }
+  }
+}
+
+template <bool Relaxed = false>
+MSCCLPP_DEVICE_INLINE void handleMultiWait(const Operation& op) {
+  DeviceSyncer* syncer = &deviceSyncers[op.deviceSyncerIndex];
```
Binyang2014: Same here

src/core/switch_channel.cc
```diff
@@ -249,18 +313,24 @@ MSCCLPP_API_CPP std::shared_ptr<NvlsConnection> connectNvlsCollective(std::share
...
-  }
+#if (CUDA_NVLS_API_AVAILABLE)
+  // Back the device-side barrier with a flag carved from this connection's own multicast region,
+  // rather than standing up a second multicast connection + bootstrap handshake. We append one
+  // granularity chunk to the requested size and bind the flag first, so it lands at a deterministic
+  // offset (0) on every rank while leaving the caller's full `bufferSize` available for later
+  // bindAllocatedMemory() calls. Flag layout: element 0 is the shared arrival counter, element 1 is
+  // this rank's generation counter. GpuBuffer zero-initializes both, the clean starting state the
+  // barrier's monotonic (never-reset) scheme requires.
+  size_t barrierReserve = nvlsRecommendedGranularity(static_cast<int>(allRanks.size()));
```
Binyang2014: I remember this will revere additional 512MB?  Will this cause any issue or we can use minimal granularity

src/core/switch_channel.cc
```diff
@@ -220,19 +220,83 @@ std::vector<char> NvlsConnection::serialize() { return pimpl_->serialize(); }
...
 
 SwitchChannel::DeviceHandle SwitchChannel::deviceHandle() const {
   SwitchChannel::DeviceHandle device;
   device.devicePtr = devicePtr_;
   device.mcPtr = mcPtr_.get();
   device.bufferSize = bufferSize_;
+  device.barrier_.mcBarrierFlag = barrierMcFlag_;
+  device.barrier_.localBarrierFlag = barrierLocalFlag_;
+  device.barrier_.barrierGen = barrierGen_;
+  device.barrier_.nRanks = barrierNRanks_;
```
Binyang2014: Maybe call barrier constructor to init the obj? Not assign all fields for barrier here

chhwang (approved):
LGTM

PR #850, opened 2026-07-21
Update CI pipeline

Binyang2014:
Force to use -j16 for build process

### DIFF
--- a/.azure-pipelines/ut.yml
+++ b/.azure-pipelines/ut.yml
@@ -71,7 +71,7 @@ jobs:
       gpuArch:          '80'
 
 - job: UnitTestH100
-  timeoutInMinutes: 40
+  timeoutInMinutes: 60
   pool:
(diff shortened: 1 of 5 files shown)


Empyreus (approved):

.azure-pipelines/ut.yml
```diff
@@ -71,7 +71,7 @@ jobs:
       gpuArch:          '80'
 
 - job: UnitTestH100
-  timeoutInMinutes: 40
+  timeoutInMinutes: 60
```
mahdiehghazim, line 74: Does increasing the timeout have any impact on whether the build passes? It should normally finish within a few minutes

Binyang2014: It's not related to build, the build issue fixed by `-j16`. Add this because our `pytest` on H100 takes long time

mahdiehghazim (approved):

PR #852, opened 2026-07-23
[Not ready] MSCCLPP EP implementation

Binyang2014:


docs/quickstart.md
```diff
@@ -110,6 +110,8 @@ Python 3.10 or later is required.
 ```bash
 # For NVIDIA platforms (specify your CUDA version)
 $ python -m pip install ".[cuda12]"
+# For NVIDIA platforms with the Expert Parallel (MoE dispatch/combine) extension
+$ python -m pip install ".[cuda12,ep]"
```
Binyang2014: We don't need ep here right?

docs/quickstart.md
```diff
@@ -127,8 +129,10 @@ Optional extras can be installed by specifying them in brackets. Available extra
 ```bash
 # Example: install with CUDA 12 and benchmark extras
 $ python -m pip install ".[cuda12,benchmark]"
+# Example: install with CUDA 12 and the EP extension
+$ python -m pip install ".[cuda12,ep]"
```
Binyang2014: Also here, no ep

include/mscclpp/gpu_data_types.hpp
```diff
@@ -63,7 +64,13 @@ using __fp8x4_e5m2 = __nv_fp8x4_e5m2;
 #define __FP8_TYPES_EXIST__
 #endif
 
+// GCC's x86 intrinsic headers may define __bfloat16 before this header is
+// included by host-only Torch extension translation units. Avoid redefining it
+// there; CUDA device compilation still uses __nv_bfloat16.
+#if !defined(MSCCLPP_AVOID_BFLOAT16_ALIAS) && !defined(__AVX512BF16INTRIN_H_INCLUDED) && \
+    !defined(__AVX512BF16VLINTRIN_H_INCLUDED)
 using __bfloat16 = __nv_bfloat16;
```
Binyang2014: Do we really need this? We dont' need torch extention

python/test/test_gpu_buffer_pool_nvls_zero.py
```diff
@@ -0,0 +1,135 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+import socket
```
Binyang2014: Why need this file?

python/csrc/core_py.cpp
```diff
@@ -63,20 +63,20 @@ void register_core(nb::module_& m) {
             void* data = reinterpret_cast<void*>(ptr);
             self->send(data, size, peer, tag);
           },
-          nb::arg("data"), nb::arg("size"), nb::arg("peer"), nb::arg("tag"))
+          nb::arg("data"), nb::arg("size"), nb::arg("peer"), nb::arg("tag"), nb::call_guard<nb::gil_scoped_release>())
```
Binyang2014, line 66: Why add gil here?

python/mscclpp/_core/comm.py
```diff
@@ -54,14 +54,21 @@ def __init__(
                 import torch
                 import torch.distributed as dist
 
+                backend = str(dist.get_backend(torch_group)).lower()
+                device = torch.device("cuda", torch.cuda.current_device()) if "nccl" in backend else torch.device("cpu")
```
Binyang2014, line 57: Why need this?

python/mscclpp/_core/comm.py
```diff
@@ -54,14 +54,21 @@ def __init__(
...
                 if rank == 0:
-                    uniq_id_global = uniq_id
                     pickled_data = pickle.dumps(uniq_id)
-                    data_tensor = torch.frombuffer(bytearray(pickled_data), dtype=torch.uint8).clone()
+                    size_tensor = torch.tensor([len(pickled_data)], dtype=torch.int64, device=device)
```
Binyang2014: Do we need to parse len first?

src/core/include/unix_socket.hpp
```diff
@@ -24,6 +24,7 @@ class UnixSocketServer {
   int registerFd(int fd);
   void unregisterFd(int fd);
   std::string getSocketPath() const;
+  ~UnixSocketServer();
```
Binyang2014: I think main branch already solve the problem?

CMakeLists.txt
```diff
@@ -51,6 +51,7 @@ option(MSCCLPP_BUILD_TESTS "Build tests" OFF)
 option(MSCCLPP_BUILD_PYTHON_BINDINGS "Build Python bindings" ON)
 option(MSCCLPP_BUILD_EXT_NCCL "Build NCCL interfaces" ON)
 option(MSCCLPP_BUILD_EXT_COLLECTIVES "Build collective algorithms" ON)
+option(MSCCLPP_BUILD_EXT_EP "Build Expert-Parallel (MoE dispatch/combine) extension" OFF)
```
Binyang2014: We can change default to on

test/mscclpp-test/CMakeLists.txt
```diff
@@ -21,3 +21,22 @@ add_mscclpp_test_executable(sendrecv_test_perf sendrecv_test.cu)
...
+            endif()
+        endif()
+    endforeach()
+    add_executable(tma_pipeline_perf tma_pipeline_perf.cu)
+    target_link_libraries(tma_pipeline_perf PRIVATE ${GPU_LIBRARIES})
+    target_include_directories(tma_pipeline_perf ${TEST_INC_COMMON})
+    set_target_properties(tma_pipeline_perf PROPERTIES
+        RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin/mscclpp-test"
+        CUDA_ARCHITECTURES "${_mscclpp_ep_test_gpu_archs}")
+endif()
```
Binyang2014: Why need these lines where is tma test?

src/ext/ep/include/api.cuh
```diff
@@ -0,0 +1,198 @@
...
+
+// ===========================================================================
+// High-throughput kernels.
+// ===========================================================================
+namespace high_throughput {
+
+void getDispatchLayout(const int64_t* topkIdx, int* numTokensPerRank, int* numTokensPerExpert, bool* isTokenInRank,
+                       int numTokens, int numTopk, int numRanks, int numExperts, cudaStream_t stream);
+
+void barrier(int** taskFifoPtrs, int head, int rank, int numRanks, cudaStream_t stream);
```
Binyang2014: Why need barrier in API? It's should be a device function?

src/ext/ep/include/api.cuh
```diff
@@ -0,0 +1,198 @@
...
+  /// Fixed-stride [num_ranks, max_tokens_per_rank, hidden], grouped by source rank.
+  RANK_MAJOR
+};
+
+// ===========================================================================
+// High-throughput kernels.
+// ===========================================================================
+namespace high_throughput {
+
+void getDispatchLayout(const int64_t* topkIdx, int* numTokensPerRank, int* numTokensPerExpert, bool* isTokenInRank,
```
Binyang2014: Why return layout? The layout should return by dispatch function? Not same as LL part?

src/ext/ep/include/api.cuh
```diff
@@ -0,0 +1,198 @@
...
+                       int numTokens, int numTopk, int numRanks, int numExperts, cudaStream_t stream);
+
+void barrier(int** taskFifoPtrs, int head, int rank, int numRanks, cudaStream_t stream);
+
+void notifyDispatch(const int* numTokensPerRank, int* mappedRecvCounter, int numRanks, const int* numTokensPerExpert,
+                    int* mappedRecvExpertCounters, int numExperts, int numTokens, const bool* isTokenInRank,
+                    int* channelPrefixMatrix, int* rankPrefixMatrix, int expertAlignment, void** bufferPtrs,
+                    int** taskFifoPtrs, int head, int rank, cudaStream_t stream, int numChannels);
+
+void cachedNotifyDispatch(const int* rankPrefixMatrix, void** bufferPtrs, int** taskFifoPtrs, int head, int rank,
```
Binyang2014: What's these two functions? Different with dispatch/combine?

src/ext/ep/include/api.cuh
```diff
@@ -0,0 +1,198 @@
...
+  int maxSharedMemoryPerBlock_;
+  /// Number of streaming multiprocessors on the device.
+  int numSms_;
+  /// CUDA device ID associated with this communicator.
+  int deviceId_;
+  /// Current rank ID.
+  int rank_;
+  /// Total number of ranks.
+  int numRanks_;
+};
```
Binyang2014: This is common struct can be shared cross ll and ht?

src/ext/ep/include/api.cuh
```diff
@@ -0,0 +1,198 @@
...
+
+/// Return the optimized low-latency workspace size.
+/// @param[in] numRanks Total number of ranks.
+/// @param[in] numExperts Total number of experts.
+/// @param[in] maxTokensPerRank Maximum local token capacity.
+/// @param[in] numTopk Number of routed experts per token.
+/// @return Required workspace bytes.
+size_t workspaceSize(int numRanks, int numExperts, int maxTokensPerRank, int numTopk);
```
Binyang2014: Why need this? For allocate memory?

src/ext/ep/include/constants.cuh
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) Microsoft Corporation.
```
Binyang2014: Pls review this file, I think we don't need this anymore. For type we already use mscclpp internal datatype

src/ext/ep/include/device_helpers.cuh
```diff
@@ -0,0 +1,497 @@
...
+__device__ __forceinline__ int64_t ld_volatile_global(const int64_t *ptr) {
+  int64_t ret;
+  asm volatile("ld.volatile.global.s64 %0, [%1];" : "=l"(ret) : "l"(ptr));
+  return ret;
+}
+
+__device__ __forceinline__ int64_t ld_volatile_global(const uint64_t *ptr) {
+  int64_t ret;
+  asm volatile("ld.volatile.global.u64 %0, [%1];" : "=l"(ret) : "l"(ptr));
+  return ret;
```
Binyang2014: Do we need these lines? We already have mscclpp atomic, mscclpp device acuqire/release function (mscclpp/include/mscclpp/concurrency_device.hpp). Do we need these low-level API?

src/ext/ep/include/device_helpers.cuh
```diff
@@ -0,0 +1,497 @@
...
+__device__ __forceinline__ uint8_t ld_nc_global(const uint8_t *ptr) {
+  uint16_t ret;
+  // NOTES: we must use `uint16_t` as inline ASM does not support 8-bit constraint letter (`h` below means unsigned
+  // 16-bit)
+  asm volatile(LD_NC_FUNC ".u8 %0, [%1];" : "=h"(ret) : "l"(ptr));
+  return static_cast<uint8_t>(ret);
+}
+
+template <>
+__device__ __forceinline__ int ld_nc_global(const int *ptr) {
```
Binyang2014: Still needs this?

src/ext/ep/common/device_helpers.cuh
```diff
@@ -0,0 +1,497 @@
...
+__device__ __forceinline__ void st_na_release(const uint64_t *ptr, uint64_t val) {
+  asm volatile("st.release.gpu.global.L1::no_allocate.b64 [%0], %1;" : : "l"(ptr), "l"(val));
+}
+
+__device__ __forceinline__ void st_na_release(const int64_t *ptr, int64_t val) {
+  asm volatile("st.release.gpu.global.L1::no_allocate.b64 [%0], %1;" : : "l"(ptr), "l"(val));
+}
+
+// `st.global.L1::no_allocate` will be translated into `ST.E.NA.[width]` in SASS
+#ifndef DISABLE_AGGRESSIVE_PTX_INSTRS
```
Binyang2014, line 29: Pls double check if needs these PTX, better to use mscclpp internal function

src/ext/ep/include/launch.cuh
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#pragma once
+
+#include "constants.cuh"
+
+#ifndef SETUP_LAUNCH_CONFIG
```
Binyang2014: Still need this?

src/ext/ep/runtime_base.hpp
```diff
@@ -0,0 +1,62 @@
...
+  // failures, so the base destructor must permit exceptions too.
+  virtual ~MoERuntime() noexcept(false) {}
+
+  MoERuntime(const MoERuntime&) = delete;
+  MoERuntime& operator=(const MoERuntime&) = delete;
+
+  /// Backend this runtime implements.
+  virtual MoEMode mode() const = 0;
+
+  bool isAvailable() const { return available_; }
```
Binyang2014: Why need to check isAvalible?

src/ext/ep/runtime_base.hpp
```diff
@@ -0,0 +1,62 @@
...
+  bool isInternodeAvailable() const { return isAvailable() && numRanks_ > numNvlRanks_; }
```
Binyang2014: Is this correct? Do we need this API? If we need why not let ht/ll runtime return this?

src/ext/ep/moe_runtime.cc
```diff
@@ -0,0 +1,22 @@
...
+namespace mscclpp {
+namespace ep {
+
+std::shared_ptr<MoERuntime> createMoERuntime(mscclpp::Communicator& communicator, MoEMode mode, int maxTokensPerRank,
+                                             int hidden, int numExperts, int numTopk, int64_t maxHiddenBytes,
+                                             int numSms) {
+  if (mode == MoEMode::LOW_LATENCY) {
+    return std::make_shared<MoELowLatencyRuntime>(communicator, maxTokensPerRank, hidden, numExperts, numTopk);
+  }
+  return std::make_shared<MoEHighThroughputRuntime>(communicator, maxHiddenBytes, high_throughput::Config(numSms));
```
Binyang2014: Why high_throuput APi need config? Not clean as lowLantecyRuntime?

src/ext/ep/ht_runtime.hpp
```diff
@@ -0,0 +1,91 @@
...
+#include <mscclpp/core.hpp>
+#include <vector>
+
+#include "high-throughput/config.cuh"
+#include "runtime_base.hpp"
+
+namespace mscclpp {
+namespace ep {
+
+class MoEHighThroughputRuntime : public MoERuntime {
```
Binyang2014: This APIs are not clean as lowLatencyMode. If we want to async why not set async in dispatch function and return a handle to wait? Why need notify. If notify is for other purpose pls change to another name

python/csrc/core_py.cpp
Binyang2014: These Bootstrap send/recv calls are blocking TCP operations. Releasing the GIL prevents Python watchdog/progress threads from being stalled while peers rendezvous; the C++ calls do not invoke Python callbacks. Keeping this intentionally.

python/mscclpp/_core/comm.py
Binyang2014: NCCL process groups reject CPU tensors, so bootstrap data must use the current CUDA device for NCCL and CPU for Gloo. The updated code also uses the subgroup global root and broadcast_object_list, avoiding the previous variable-length parsing and subgroup src=0 issue.

src/ext/ep/include/api.cuh
Binyang2014: This struct should remain LL-only. invalidTokenExpertId, maxTokensPerRank, outputLayout, and dispatchDataType are LL-specific; HT uses dynamic receive sizing, token-major output, and its own transport state. Sharing it would create a common struct with invalid fields for HT.

src/ext/ep/runtime_base.hpp
Binyang2014: isAvailable is needed because a runtime may be constructed on an unsupported rank count or topology and report unavailable without partially initializing resources. The Python high-level API checks this immediately and surfaces the failure before dispatch.

src/ext/ep/runtime_base.hpp
Binyang2014: The topology calculation and availability contract are identical for LL and HT, so the base owns them once. Concrete runtimes set available_ after their resources are initialized; callers query the common interface without downcasting.

src/ext/ep/moe_runtime.cc
Binyang2014: LL and HT need different construction parameters: LL needs token/model capacity, while HT needs max hidden bytes and an SM budget. The public Python MoECommunicatorConfig hides this raw factory detail and selects the concrete runtime by MoEMode.

src/ext/ep/ht_runtime.hpp
Binyang2014: notify is not an async completion API. It is the mandatory sizing/exchange phase: ranks exchange routing counts, publish mapped receive counters, and the host obtains numRecvTokens before it can expose the exact-size receive view. The internal launchers were renamed to computeDispatchCounts/exchangeDispatchCounts/publishCachedRankPrefix and documented; the existing Python method name is retained for compatibility.

PR #855, opened 2026-07-27
Add multi-node NVL/NVLS algorithm support

Binyang2014:
## Summary

- Add first-class multi-node NVLink/NVLS support for fabrics such as GB200 NVL72.
- Generalize native collective topology handling from ranks-per-node to ranks-per-IPC-domain.
- Make NVLS allreduce algorithms and launch sizing MNNVL-aware.
- Add MNNVL tuning updates and FP8 NVLS accumulator support.
- Keep the GpuBufferPool API, implementation, bindings, and tests out of this PR.

## Validation

- `./tools/lint.sh`
- Release build with `MSCCLPP_BUILD_TESTS=ON`
- `./build/bin/unit_tests`
- `mpirun -np 2 ./build/bin/mp_unit_tests --filter=-*Ib*`


include/mscclpp/switch_channel_device.hpp
```diff
@@ -84,7 +83,9 @@ struct SwitchChannelDeviceHandle {
           : "=r"(val.words[0]), "=r"(val.words[1]), "=r"(val.words[2]), "=r"(val.words[3])
           : "l"(ptr)
           : "memory");
-    } else if constexpr (std::is_same_v<VectorType, f8_e4m3x4>) {
+    }
+#if (defined(__CUDA_ARCH_SPECIFIC__) || defined(__CUDA_ARCH_FAMILY_SPECIFIC__)) && (__CUDA_ARCH__ >= 1000)
+    else if constexpr (std::is_same_v<VectorType, f8_e4m3x4>) {
```
Binyang2014: Do we need this macro or we already check somewhere? Will raise some runtime error if not support?

src/ext/collectives/allreduce/allreduce_fullmesh.cu
```diff
@@ -9,12 +9,23 @@
...
+// Per-context cache of input-side MemoryChannels keyed by input pointer.
+// Lifetime is tied to AlgorithmCtx, so entries are released when the ctx is
+// evicted from the framework's context cache (avoids unbounded growth across
+// allreduce calls that pass different input buffers).
+using InputChannelsCache =
+    std::unordered_map<const void*,
+                       std::pair<std::vector<MemoryChannel>, std::shared_ptr<DeviceHandle<MemoryChannel>>>>;
+constexpr const char* kInputChannelsExtraKey = "inputChannels";
+}  // namespace
+
```
Binyang2014, line 23: Will we release these handles when we reset context?

include/mscclpp/switch_channel_device.hpp
Binyang2014: Verified: the macro is required, and it is not redundant with the other checks.

Without the guard, any target below `sm_100a` fails to build. Reproduced on this H100 box:

```
ptxas ... error : Feature '.e4m3x4' not supported on .target 'sm_90'
ptxas fatal   : Ptx assembly aborted due to errors
```

The other checks are not independent of it, they are all derived from the same condition:
- `common.hpp` / `execution_kernel.hpp` use the inverse condition to short-circuit FP8 in an `if constexpr ... else`, which is what prevents instantiating the compiled-out branch (otherwise it hits the `static_assert`).
- `fp8NvlsSupportProbeKernel` in `collective_utils.cu` uses the same condition to report the capability at runtime.

On unsupported hardware it does not crash: `isFp8NvlsSupported()` gates the NVLS algorithms, which return `CommInvalidArgument` with a `WARN`. The device-side `MSCCLPP_ASSERT_DEVICE` is only a backstop.

Since the condition was duplicated in 5 places with 3 different spellings, I centralized it as `MSCCLPP_DEVICE_FP8_MULTIMEM_SUPPORTED` in `device.hpp` so the compile-time guards and the runtime probe can no longer drift apart.

src/ext/collectives/allreduce/allreduce_fullmesh.cu
Binyang2014: Yes, they are released. `AlgorithmCtx::extras` is `unordered_map<string, shared_ptr<void>>`, and the cache is inserted as `make_shared<InputChannelsCache>()`, so the `shared_ptr<void>` keeps the concrete deleter. `NativeAlgorithm::reset()` does `contexts_.clear()`, which destroys the context, its `extras`, and therefore every cached `MemoryChannel` and device handle.

That said, the comment I had written was wrong on the important part: it claimed entries are released when the ctx is "evicted from the framework's context cache" and that this "avoids unbounded growth". There is no per-entry eviction, `contexts_` only grows until an explicit `reset()`. So the growth is not avoided, it just moved from `contexts_` into the per-ctx cache. I corrected the comment to describe the actual lifetime.

Separately, while checking this I found a real issue in the same function that is worth deciding on: `generateAllreduceContextKey` now uses `tag++` for the non-symmetric path, so every call produces a unique key. That means a cache miss and a brand-new context on every allreduce, including a `setupRemoteMemories` handshake, and those contexts are never released. It also makes the new input-channel cache dead code in that path. Want me to fix that in this PR?

include/mscclpp/switch_channel_device.hpp
Binyang2014: Follow-up on macro placement and style (`87f6f15a`):

- Moved `MSCCLPP_DEVICE_FP8_MULTIMEM_SUPPORTED` from `device.hpp` to `gpu_data_types.hpp`, next to `__FP8_TYPES_EXIST__` / `__FP8_E4M3_IS_FNUZ__`. It has to stay in a public header because `switch_channel_device.hpp` itself uses it; if it moved to `src/include/`, the public header would silently see it as undefined and compile FP8 multimem out even on supported hardware.
- Switched from a `0`/`1` value macro to plain define/undefine, matching CUDA's own capability macros (`__CUDA_BF16_TYPES_EXIST__`, `__CUDA_FP16_TYPES_EXIST__`) and all 8 existing capability macros in this repo. It also reads consistently where it is combined with `__FP8_TYPES_EXIST__`.

For reference, `cuda_fp8.h:73` gates FP8 with essentially the same condition:
```c
#if ( (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 1000) && (defined __CUDA_ARCH_FAMILY_SPECIFIC__) )
```

Verified `sm_90` still compiles FP8 out and `sm_100a` still enables it; lint, build, 34 unit tests and 57 two-rank MPI tests pass.

src/ext/collectives/allreduce/allreduce_fullmesh.cu
Binyang2014: Correction to my previous reply: please disregard the `tag++` concern I raised. It is intentional for the non-symmetric-memory path, where the buffer addresses are not stable enough to serve as a context key, so a fresh context per call is by design and `reset()` is the intended release point. This also matches the existing usage already on `main` in `allgather_fullmesh_2.cu` and `allreduce_rsag_zero_copy.cu`. No change needed here.

The original question stands as answered: the cached input channels and device handles are owned by `AlgorithmCtx::extras` and are destroyed with the context on `reset()`. Only the inaccurate comment was fixed.

src/ext/collectives/allgather/allgather_fullmesh_2.cu
```diff
@@ -162,12 +162,12 @@ CommResult AllgatherFullmesh2::allgatherKernelFunc(const std::shared_ptr<void> c
   size_t channelOutOffset = *static_cast<size_t*>(ctx->extras["channel_out_offset"].get());
   if ((char*)input == (char*)output + rank * inputSize) {
     allgatherFullmesh2<false><<<numBlocksAndThreads.first, numBlocksAndThreads.second, 0, stream>>>(
-        (void*)input, ctx->memoryChannelDeviceHandles.get(), channelOutOffset, ctx->rank, ctx->workSize,
-        ctx->nRanksPerNode, nElem);
+        (void*)input, ctx->memoryChannelDeviceHandles.get(), channelOutOffset, ctx->rank, ctx->worldSize,
+        ctx->nRanksPerIpcDomain, nElem);
```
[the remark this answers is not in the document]
Binyang2014: Good catch, this was a real regression and is fixed in `34f7d230`.

`ctx->nRanksPerNode` was never assigned anywhere in `src/ext/collectives/` after this PR switched context init to `nRanksPerIpcDomain`, and `AlgorithmCtx` fields are plain `int` with no default initializer, so both remaining readers were using an uninitialized value. On `main` these call sites had a matching `ctx->nRanksPerNode = ...` assignment, so this was introduced here.

Fixed both readers:
- `allgather_fullmesh_2.cu:143` (warp-count validation)
- `allreduce_packet.cu:247` (block-count validation, same root cause, mentioned in the AllreducePacket comment)

Both now use `ctx->nRanksPerIpcDomain`, consistent with the kernel launch arguments in the same functions. Also removed the now-unused `AlgorithmCtx::nRanksPerNode` field so it cannot be misused again. The `nRanksPerNode` in `Algorithm::Constraint` and `bootstrap()->getNranksPerNode()` are different entities and are unchanged; the two examples that assign `ctx->nRanksPerNode` use their own local `struct Context` and do not include `collective_utils.hpp`.

Verified with lint, a full build including examples, 34 unit tests and 57 two-rank MPI tests.

src/ext/collectives/collective_utils.cu
```diff
@@ -184,6 +186,8 @@ std::vector<mscclpp::SwitchChannel> setupNvlsChannels(std::vector<std::shared_pt
     mscclpp::SwitchChannel switchChannel = nvlsConnection->bindAllocatedMemory((CUdeviceptr)buffer, bufferSize);
     channels.push_back(switchChannel);
   }
+  // Synchronize to make sure all ranks have their NVLS channels set up before any rank starts using them.
+  comm->bootstrap()->barrier();
```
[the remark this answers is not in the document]
Binyang2014: I looked into this and do not think the deadlock is reachable, so I left the barrier as is.

All five callers of `setupNvlsChannels()` are inside `initAllreduceContext`, and for those algorithms the context key does not depend on per-rank state:

- `allreduce_nvls_block_pipeline.cu:216` and `allreduce_nvls_warp_pipeline.cu:181` return a constant `AlgorithmCtxKey{nullptr, nullptr, 0, 0, 0}`.
- `allreduce_nvls_packet.cu` calls it from `initialize()`, which every rank runs.
- `allreduce_nvls_zero_copy.cu` keys on the buffer base pointer and size, which are symmetric across ranks for this algorithm (it already requires symmetric memory).

So ranks hit or miss the context cache in lockstep, and the "some ranks hit while others miss" precondition does not occur. The barrier is needed because a rank must not start using its multicast handles before all ranks have bound theirs.

That said, the general concern is fair as a latent invariant: it only holds as long as NVLS context keys stay rank-independent. If we later add an NVLS algorithm with a rank-dependent key, this would need revisiting.

src/ext/collectives/allreduce/allreduce_packet.cu
```diff
@@ -231,9 +225,19 @@ CommResult AllreducePacket::allreduceKernelFunc(const std::shared_ptr<void> ctx_
...
-    blockAndThreadNum = getDefaultBlockNumAndThreadNum(inputSize, ctx->workSize, ctx->nRanksPerNode, dtype);
+    blockAndThreadNum = getDefaultBlockNumAndThreadNum(inputSize, ctx->nRanksPerIpcDomain, ctx->worldSize, dtype);
+  } else {
+    const int nPeers = ctx->nRanksPerIpcDomain - 1;
+    if (blockAndThreadNum.first < nPeers) {
+      return CommResult::CommInvalidArgument;
+    }
   }
   if (blockAndThreadNum.first > maxBlockNum_) {
     WARN(ALGO, "Requested block number ", blockAndThreadNum.first, " exceeds the maximum supported block number ",
```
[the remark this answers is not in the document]
Binyang2014: Partly addressed in `34f7d230`.

The peer-count part was a real bug: `ctx->nRanksPerNode` is never assigned and `AlgorithmCtx` has no default initializers, so line 247 was reading an uninitialized value. It now uses `ctx->nRanksPerIpcDomain`, and the unused field has been removed.

I did not change the default block heuristic. The overflow is arithmetically real (`(worldSize - 1) * 8` is 568 for a 72-rank domain, well above `maxBlockNum_ = 112`), but it is not reachable today: `nccl.cc:264` dispatches on `request.nRanksPerNode != request.worldSize`, so any MNNVL configuration goes to `selectMultiNodeAlgorithm`, which currently returns `nullptr` and falls back to NCCL/RCCL. The current code also fails loudly with a `WARN` plus `CommInvalidArgument` rather than silently, and this heuristic is untouched by this PR. Clamping to `maxBlockNum_` and aligning to a multiple of `nPeers` would be a behavior change that is better made together with actual multi-node algorithm selection. Happy to do it here if you prefer.

src/ext/collectives/allreduce/allreduce_fullmesh.cu
```diff
@@ -209,39 +233,36 @@ CommResult AllreduceFullmesh::allreduceKernelFunc(
...
+    return AlgorithmCtxKey{nullptr, (void*)recvBasePtr, 0, recvBytes, tag++};
   }
-
-  size_t sendBytes, recvBytes;
-  CUdeviceptr sendBasePtr, recvBasePtr;
-  MSCCLPP_CUTHROW(cuMemGetAddressRange(&sendBasePtr, &sendBytes, (CUdeviceptr)input));
-  MSCCLPP_CUTHROW(cuMemGetAddressRange(&recvBasePtr, &recvBytes, (CUdeviceptr)output));
-  return AlgorithmCtxKey{(void*)sendBasePtr, (void*)recvBasePtr, sendBytes, recvBytes, 0};
+  return AlgorithmCtxKey{nullptr, (void*)recvBasePtr, 0, recvBytes, 0};
 }
```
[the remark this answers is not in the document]
Binyang2014: This one is intentional, no change made. The `tag++` on the non-symmetric path is by design: without symmetric memory the buffer addresses are not stable enough to serve as a context key, so a fresh context per call is expected and `NativeAlgorithm::reset()` is the intended release point. The same pattern already exists on `main` in `allgather_fullmesh_2.cu:222` and `allreduce_rsag_zero_copy.cu:189`.

src/ext/collectives/collective_utils.cu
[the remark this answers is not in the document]
Binyang2014: no need to fix, for NVLS need all ranks involved

python/mscclpp_benchmark/bench_collective.py
```diff
@@ -199,7 +199,6 @@ def _candidate_specs(collective: str, *, symmetric_memory: bool = False) -> tupl
             "default_allreduce_nvls_packet",
             max_message_size=512 * 1024,
             max_nblocks=16,
-            supported_skus=("H100", "GB300"),
```
caiocbr, line 202: Does this algorithm work for AMD?

src/ext/collectives/allreduce/allreduce_allpair_packet.cu
```diff
@@ -14,14 +14,11 @@ namespace collective {
 
 template <ReduceOp OpType, typename T, typename AccumT = T>
 __global__ void allreduceAllPairs(T* buff, T* scratch, T* resultBuff, DeviceHandle<MemoryChannel>* memoryChannels,
-                                  size_t channelDataOffset, size_t scratchBufferSize, int rank, int nRanksPerNode,
+                                  size_t channelDataOffset, size_t scratchBufferSize, int rank, int nRanksPerIpcDomain,
                                   int worldSize, size_t nelems, uint32_t numScratchBuff, void* flags,
                                   uint32_t flagSize) {
-  // This version of allreduce only works for single nodes
-  if (worldSize != nRanksPerNode) return;
```
caiocbr, line 21: Maybe you can keep this and just update the if condition to ```worldSize != nRanksPerIpcDomain```

python/mscclpp_benchmark/bench_collective.py
Binyang2014: No, we have `requires_nvls=True` to avoid this

src/ext/collectives/allreduce/allreduce_allpair_packet.cu
```diff
@@ -71,25 +68,25 @@ __global__ void allreduceAllPairs(T* buff, T* scratch, T* resultBuff, DeviceHand
   }
 }
 
-inline std::pair<int, int> getDefaultBlockNumAndThreadNum(size_t inputSize, int worldSize) {
```
caiocbr, line 74: In this case the world size should be as same as nRanksPerIpcDomain, no?

Binyang2014: Correct, and that is now enforced on the host side. `allreduceKernelFunc` rejects the call with `CommInvalidArgument` unless `worldSize == nRanksPerIpcDomain` (line 108), so by the time the kernel launches the two are guaranteed equal.

`worldSize` is still passed to the kernel only because it is part of the shared `AllreduceFunc` signature used by all allreduce adapters. Also maybe we will add cross rack algo in future

src/ext/collectives/allreduce/allreduce_allpair_packet.cu
Binyang2014: The check was not dropped, it moved to the host side with the equivalent condition. `allreduceKernelFunc` now does this before launching (`allreduce_allpair_packet.cu:108`):

```cpp
if (algoCtx->worldSize != algoCtx->nRanksPerIpcDomain) {
  WARN(ALGO, "AllreduceAllpairPacket requires worldSize to match nRanksPerIpcDomain, ...");
  return CommResult::CommInvalidArgument;
}
```

`allreduceAllPairs` has exactly one launch path (`allreduceKernelFunc` -> `dispatch<AllpairAdapter>` -> the kernel), so the host-side gate covers every invocation.

Keeping the device-side `return` as well would be strictly worse than the host check: it silently returns success while producing wrong results, whereas the host check reports `CommInvalidArgument` with a `WARN` and never launches the kernel. It would also cost a redundant per-thread branch. So I would prefer to leave it as is, but happy to restore it as a defensive assert if you feel strongly.

caiocbr (approved):
Approved. The current implementation looks good to merge. I suggest addressing the following non-blocking improvements in follow-up PRs:

Revisit InputChannelsCache in the full-mesh all-reduce algorithm. Creating and caching full MemoryChannels may not be necessary for this data-transfer path. Since the input-side channels are used only to copy data and not for synchronization, we should investigate using the underlying memory handles directly. This could simplify resource ownership and avoid an unbounded per-input-pointer cache.

Consolidate handleMultiLoadReduceStore. There appears to be duplicated multimem load-reduce-store logic in the collective and executor implementations. We should consider moving the shared low-level operation into a common reusable helper while keeping context-specific offset calculation, type handling, and synchronization in the respective callers.

PR #836, opened 2026-07-15
ep(bench): unified in-process Python EP benchmark (mscclpp vs NCCL-EP)

seagater:
## Summary
Adds `run_ep_bench_python.py` — an in-process low-latency EP benchmark that drives
**both** the mscclpp EP (`MoECommunicator`) and NVIDIA NCCL-EP Python APIs through
one shared paired dispatch→combine loop, so the two backends are timed with
byte-for-byte identical methodology (matching `mscclpp_ep_bench.cu` output).

## Changes
- New `test/python/ep/run_ep_bench_python.py` (renamed from `ep_bench_unified.py`),
  MPI/mpi4py bootstrap, `--backend {mscclpp,nccl,both}`, optional in-process CUPTI
  kernel timing.
- `run_ep_bench.py`: dropped the redundant Python `mscclpp` (`ep_bench_ll.py`)
  backend; deleted `ep_bench_ll.py` (superseded by the unified script).
- Updated the mscclpp LL API usage for the current (IPC-capable) runtime
  (`num_rdma_bytes=0`; removed `get_low_latency_rdma_size_hint`).
- Docstring: clarified mscclpp LL supports both CUDA-IPC (NVLink) and RDMA/IB, and
  added single-node and validated 2-node (shared MNNVL fabric) launch examples with
  placeholder IPs/iface.

Both backends run at 1 and 2 nodes over the shared NVLink fabric.

PR #858, opened 2026-07-30
ep(bench): unified in-process Python EP benchmark — review-comment follow-ups

seagater:
## Summary
Follow-up changes on top of the merged unified Python EP benchmark (#836),
addressing the outstanding review comments and refreshing the CUDA-graph
methodology. Benchmarks all four EP backends — mscclpp EP, NVIDIA NCCL-EP,
DeepEP V2, FlashInfer — through one shared in-process harness with identical
dispatch→combine timing.

## Changes (by review comment)
- **Move the kernel-name parse into the backend files** — each backend owns
  `parse_kineto_kernels(...)`; the harness holds zero per-library kernel knowledge
  (shared `sum_matching_kernel_us()` in ep_bench_common).
- **remove sync here** — dropped the redundant warmup `stream.synchronize()`.
- **Move cudaGraph logic to run_ep_bench_python / unified timing** — capture is
  owned by the harness (`_capture_paired_graph`); each `setup_*` returns a uniform
  `{dispatch, combine, teardown, barrier, graph}` dict and one shared loop times all
  four backends. Per-library specifics stay in the backend closures.
- **run multiple iterations inside the cuda graph** — added `--iters-per-graph N`
  (sglang bench_moe_ep.py pattern): N dispatch→combine iterations per graph, reported
  per iteration.

## Other fixes
- DeepEP CUDA-graph gate keyed on transport (EP_DISABLE_GIN), not node count —
  verified graph capture at 1/2/4 nodes on GB200 NVL72.
- Force the paired kineto pass whenever the harness captures a single graph
  (fixes mscclpp "captured 0" under --cuda-graph).
- Removed the C++ LL bench, shell driver, and CUPTI timer.
- Clarified EP_KINETO_SEPARATE=0 is the required (not "legacy") CUDA-graph mode.

## Validation
All four backends: correct kernel-only dispatch/combine in eager + CUDA-graph,
genuine capture (no eager fallback), consistent per-iteration numbers at
--iters-per-graph 1 and 10, on GB200 NVL72.


test/python/ep/run_ep_bench_python.py
```diff
@@ -175,9 +172,14 @@ def parse_args() -> argparse.Namespace:
     )
     p.add_argument(
         "--graph-group-size",
+        "--iters-per-graph",
+        dest="graph_group_size",
         type=int,
-        default=1,
-        help="capture this many independently routed dispatch->combine pairs in one CUDA Graph",
+        default=10,
```
Binyang2014: Maybe increase to 50 by default?

test/python/ep/run_ep_bench_python.py
```diff
@@ -214,10 +208,10 @@ def parse_args() -> argparse.Namespace:
...
         raise SystemExit("--graph-group-size must be positive")
-    if args.graph_group_size > 1 and not args.cuda_graph:
-        raise SystemExit("--graph-group-size > 1 requires --cuda-graph")
-    if args.graph_group_size > 1 and args.backend not in ("mscclpp", "flashinfer"):
-        raise SystemExit("grouped CUDA Graph timing currently supports mscclpp or flashinfer")
+    if not args.cuda_graph:
+        # Grouping only applies to graph capture; treat as 1 for eager runs so the
+        # non-1 default does not error a plain (non-graph) benchmark.
+        args.graph_group_size = 1
     if args.dispatch_dtype == "fp8_e4m3" and args.backend in ("nccl", "all"):
```
Binyang2014, line 211: I think nccl support fp8 right?

test/python/ep/run_ep_bench_python.py
```diff
@@ -214,10 +208,10 @@ def parse_args() -> argparse.Namespace:
         raise SystemExit("--num-warmup must be non-negative and --num-iters must be positive")
     if args.graph_group_size <= 0:
```
Binyang2014: Different with iteration_per_group?

test/python/ep/run_ep_bench_python.py
```diff
@@ -231,7 +225,15 @@ def _flush_l2_cache():
 
 
 def _kineto_kernel_us(
-    dispatch_fn, combine_fn, comm, num_tests, flush_l2=True, use_barrier=True, barrier=None, mid_barrier=None
+    dispatch_fn,
```
Binyang2014, line 227: Change to another name? kineto is confused here...

test/python/ep/run_ep_bench_python.py
```diff
@@ -345,7 +349,57 @@ def _run_pass(op_fn):
...
-    return _parse(ka, "dispatch"), _parse(ka, "combine")
+    return parse_kernels(ka)
+
+
+# ============================================================================
+# CUDA-graph capture (owned by the harness, not the backends). Every EP backend
+# replays dispatch+combine as ONE combined graph, so this single helper captures
+# the paired op from any backend; the backend only supplies its capture-safe ops.
+# ============================================================================
+def _capture_paired_graph(dispatch_op, combine_op, prime=True, pre_replay=None, on_fail=None, iters_per_graph=1):
```
Binyang2014: Why need pre_replay here?

test/python/ep/run_ep_bench_python.py
```diff
@@ -357,9 +411,10 @@ def run_backend(
     inputs,
     dispatch_fn,
     combine_fn,
-    cupti=None,
     nccl_barrier=None,
     bench_barrier=None,
+    parse_kernels=None,
+    iters_per_graph=1,
```
Binyang2014: Why hard code to 1 here?

test/python/ep/run_ep_bench_python.py
seagater: This is a special handle for the FlashInfer. "FlashInfer's dispatch/combine each need an MPI host barrier to align ranks, but an MPI barrier cannot live inside a CUDA graph -- so the captured ops are barrier-free and the harness runs the barrier BEFORE each replay (pre_replay)." in ep_bench_flashinfer.py‎.

test/python/ep/run_ep_bench_python.py
seagater: Changed to torch_profiler_kernel_us.

test/python/ep/run_ep_bench_python.py
seagater: Changed default value to 50.

test/python/ep/run_ep_bench_python.py
seagater: Changed to graph_group_size.

test/python/ep/run_ep_bench_python.py
seagater: Changed to graph_group_size,  1 = no grouping (one iteration captured).

test/python/ep/run_ep_bench_python.py
seagater: Yes. Updated the descript. NCCL-EP path in this benchmark is BF16-only.

test/python/ep/run_ep_bench_python.py
Binyang2014: Why it require this? I think dispatch/combine is the sync point. No need additional sync before dispatch

test/python/ep/ep_bench_deepep.py
```diff
@@ -99,75 +110,71 @@ def setup_deepep(args, comm, rank, num_ranks, inputs):
...
-    # internode scale-out path DeepEP's symmetric-memory kernels crash under graph
-    # capture (CUDA 719 in symmetric.hpp), so restrict capture to a single node.
-    local_world = int(os.environ.get("MSCCLPP_EP_LOCAL_WORLD_SIZE", "0") or "0")
-    deepep_can_graph = args.cuda_graph and (local_world <= 0 or num_ranks <= local_world)
+    # DeepEP CUDA-graph capture is limited by TRANSPORT, not node count. On the
+    # all-NVLink / MNNVL path (EP_DISABLE_GIN=1, a single NVL72 domain) the
+    # symmetric-memory kernels ARE graph-capturable at any node count -- verified
+    # capturing at 1/2/4 nodes on a GB200 NVL72. Only the RDMA/IB scale-out path
+    # (GIN enabled, multi-rack) crashes under graph capture (CUDA 719 in
+    # symmetric.hpp), so disable capture only when GIN is active.
```
Binyang2014: This sounds like a DeepEP bug, this is by design?

test/python/ep/ep_bench_deepep.py
```diff
@@ -99,75 +110,71 @@ def setup_deepep(args, comm, rank, num_ranks, inputs):
...
-        # NVLink has negligible combine recv-spin skew, so the paired pass matches the
-        # separate pass here.)
-        os.environ["EP_KINETO_SEPARATE"] = "0"
         if rank == 0:
             print("[cfg] deepep cuda_graph=True (single graph, cached dispatch, do_cpu_sync=False)", flush=True)
         # The non-cached dispatch above did a CPU sync to size the layout; that is
-        # illegal inside a CUDA graph. Replay the CACHED dispatch instead: pass the
-        # primed handle (topk_idx reused from it), which forces do_cpu_sync=False and
-        # skips the host-side count read, leaving a pure on-stream kernel launch.
+        # illegal inside a CUDA graph. Use the CACHED dispatch instead: pass the
```
Binyang2014, line 143: For non-cached dispatch, I think it will use allgather to get layout, it will break the cuda graph?

test/python/ep/ep_bench_flashinfer.py
```diff
@@ -158,43 +143,23 @@ def _combine():
...
-            moe.dispatch(token_selected_experts, [hidden_payload], num_tokens)
-            return None
 
-        def combine_fn(_dout):
-            comm.Barrier()
-            moe.combine(combine_payload, num_tokens)
+        graph_spec = {
+            "dispatch": _dispatch,
+            "combine": _combine,
+            "pre_replay": comm.Barrier,
```
Binyang2014: Do we really need this?

test/python/ep/run_ep_bench_python.py
```diff
@@ -175,9 +173,14 @@ def parse_args() -> argparse.Namespace:
     )
     p.add_argument(
         "--graph-group-size",
```
Binyang2014: Maybe remove the --graph-group-size here, just keep --iters-per-graph

seagater: Resolved.

test/python/ep/ep_bench_deepep.py
seagater: Non-cached dispatch keeps [do_cpu_sync=True] and that is what breaks CUDA-graph capture — not the layout exchange (there's no allgather in the dispatch path; the inter-rank exchange is on-stream inside the GIN kernel). With [do_cpu_sync=True] the runtime:

1. does a blocking device→host readback of the exact received-token count, and
2. allocates the dispatch outputs to that runtime-determined (dynamic) size (recv_src_metadata.shape[0] in [deepseek-ai\DeepEP\tree\main\deep_ep\buffers\elastic.py]

Both are illegal under graph capture: capture records only async on-stream work (a blocking D2H sync aborts it), and a graph requires fixed, pre-known output shapes/addresses (data-dependent sizing isn't allowed).

test/python/ep/ep_bench_deepep.py
seagater: Update the description. Seems the limitation is related to the do_cpu_sync=True.

Andyli1007:
Unscribe
________________________________

PR #859, opened 2026-07-30
Handle empty NVML fabric cluster UUID

Binyang2014:
## Summary

- detect an empty NVML fabric cluster UUID
- fall back when NVML reports an all-zero UUID instead of deriving an invalid IPC-domain hash

## Testing

- Not run (NVML fabric topology required)


seagater (approved):
Looks good to me.

PR #863, opened 2026-08-07
Add NVLS switch-channel cross-rank barrier api

Empyreus:
Adds a device-side group barrier built directly on the NVLS switch (NVLink SHARP multicast). A barrier becomes a single  signal() / wait()  pair backed by one  multimem  atomic on a shared arrival counter that the switch reflects into every rank — replacing the O(n²) mesh of memory-channel semaphores previously needed to synchronize a multicast group.


src/core/switch_channel.cc
```diff
@@ -220,19 +220,82 @@ std::vector<char> NvlsConnection::serialize() { return pimpl_->serialize(); }
...
+#if (CUDA_NVLS_API_AVAILABLE)
+// Query the multicast granularity for `numDevices` without allocating, so the barrier reserve can be
+// sized before the connection's multicast object is created. Mirrors the min/max handling used when
+// the connection later computes its own granularity.
+size_t nvlsMinimumGranularity(int numDevices) {
+  CUmulticastObjectProp prop = {};
+  prop.numDevices = numDevices;
+  prop.size = 0;
+  size_t granFd = 0, granFabric = 0;
+  prop.handleTypes = CU_MEM_HANDLE_TYPE_POSIX_FILE_DESCRIPTOR;
```
Binyang2014: Could you use CU_MEM_HANDLE_TYPE_POSIX_FILE_DESCRIPTOR| CU_MEM_HANDLE_TYPE_FABRIC here?

src/core/switch_channel.cc
```diff
@@ -249,18 +312,24 @@ MSCCLPP_API_CPP std::shared_ptr<NvlsConnection> connectNvlsCollective(std::share
...
+  // barrier's monotonic (never-reset) scheme requires.
+  size_t barrierReserve = nvlsMinimumGranularity(static_cast<int>(allRanks.size()));
+  auto conn = makeNvlsConnection(bootstrap, allRanks, rank, rootRank, isRoot, bufferSize + barrierReserve, 0);
+
+  auto barrierBuffer = std::make_shared<GpuBuffer<uint64_t>>(2);
+  auto barrierChannel = std::make_shared<SwitchChannel>(
+      conn->bindAllocatedMemory(CUdeviceptr(barrierBuffer->data()), barrierBuffer->bytes()));
+  conn->attachBarrier(barrierBuffer->memory(), barrierChannel, static_cast<int>(allRanks.size()));
+#else
+  auto conn = makeNvlsConnection(bootstrap, allRanks, rank, rootRank, isRoot, bufferSize, 0);
```
Binyang2014: Maybe just throw error here?

src/core/switch_channel.cc
```diff
@@ -249,18 +312,24 @@ MSCCLPP_API_CPP std::shared_ptr<NvlsConnection> connectNvlsCollective(std::share
...
+  // bindAllocatedMemory() calls. Flag layout: element 0 is the shared arrival counter, element 1 is
+  // this rank's generation counter. GpuBuffer zero-initializes both, the clean starting state the
+  // barrier's monotonic (never-reset) scheme requires.
+  size_t barrierReserve = nvlsMinimumGranularity(static_cast<int>(allRanks.size()));
+  auto conn = makeNvlsConnection(bootstrap, allRanks, rank, rootRank, isRoot, bufferSize + barrierReserve, 0);
+
+  auto barrierBuffer = std::make_shared<GpuBuffer<uint64_t>>(2);
+  auto barrierChannel = std::make_shared<SwitchChannel>(
+      conn->bindAllocatedMemory(CUdeviceptr(barrierBuffer->data()), barrierBuffer->bytes()));
+  conn->attachBarrier(barrierBuffer->memory(), barrierChannel, static_cast<int>(allRanks.size()));
```
Binyang2014: Just want to make sure here we will retains the buffer reference in NvlsConnection right? If so why not allocate buffer inside the attachBarrier.

Binyang2014 (approved):

src/core/switch_channel.cc
Empyreus: I think leaving them separate will be better here. That way we can handle failure graciously and make the readability better to understand we are getting the minimum from these options.

Binyang2014 (approved):

src/core/switch_channel.cc
Binyang2014: How about use this function: `getMulticastGranularity` in include/mscclpp/gpu_utils.hpp

Binyang2014 (approved):

PR #864, opened 2026-08-07
Proxy FIFO latency improvement

chhwang:
## Summary

The FIFO now uses the reserved bit in `ProxyTrigger::snd` as a lap-parity commit bit instead of treating `fst != 0` as readiness.

- `push()` derives parity from the monotonic head position and FIFO size.
- CUDA publishes both words with one `st.global.release.sys.v2.u64`.
- ROCm writes `fst` relaxed, then commits with a release store to `snd`.
- `poll(ProxyTrigger&)` returns readiness separately, acquire-loads `snd`, validates parity, then reads `fst`.
- `pop()` advances `tail` without clearing the slot; the next lap's parity makes stale data invalid.
- FIFO size is now a positive power of two, so slot and lap calculations use a mask and shift.

This removes the payload sentinel and its workarounds: zero-valued triggers now round-trip, zero-byte trigger payloads are legal, and `signal()`/`flush()` no longer pass a dummy `bytes = 1`. It also fixes the custom-trigger hazard called out at the old `proxy.cc` TODO: readiness no longer depends on user payload.

New tests cover a fully zero trigger, exact lap-boundary wraparound, and invalid FIFO sizes. `FifoTest.Fifo` now starts at payload 0 instead of avoiding the sentinel.

## Performance

Same-node before/after runs, Release builds, three `FifoTest.Fifo` samples unless noted:

| GPU | `origin/main` | enhanced FIFO | Improvement |
|---|---:|---:|---:|
| H200, CUDA 12.9 | 2.083 us/iter | 1.973 us/iter | **5.3%** |
| A100-SXM4-80GB, CUDA 12.9 | 3.907 us/iter | 3.559 us/iter | **8.9%** |
| MI300X, ROCm 7.2 | 1.184 us/iter | 1.128 us/iter | **4.7%** |

H200 `PortChannelOneToOneTest.PingPongPerf` also improved within run-to-run variance: 19.38 us/iter baseline vs 19.02 us/iter enhanced. Packet ping-pong was unchanged at about 8.1 us/iter.

The A100 `__threadfence_system()` + relaxed-store special case is removed. Controlled sm_80 measurements show that plain `st.global.release.sys.v2.u64` is now faster:

| A100 code | fence + relaxed | release | Improvement |
|---|---:|---:|---:|
| `origin/main` FIFO | 3.907 us/iter | 3.687 us/iter | **5.6%** |
| enhanced FIFO | 3.687 us/iter | 3.559 us/iter | **3.5%** |

The throughput gain comes primarily from removing the host write that cleared `fst` in every `pop()`.

## Accuracy

- H200: `unit_tests` 37/37; `mp_unit_tests` 29/29.
- MI300X: 26 unit tests passed and 9 skipped, versus 23 passed and 9 skipped on baseline; the three added passes are the new FIFO tests.
- MI300X multiprocess results match baseline: both builds have the same four unrelated environment failures on the node without usable IB.

The custom allgather pattern remains compatible: handlers still receive clean trigger payloads, bit 63 of `snd` remains reserved, and raw triggers may now use `fst == 0` safely.

### DIFF
--- a/include/mscclpp/fifo.hpp
+++ b/include/mscclpp/fifo.hpp
@@ -16,15 +16,21 @@ constexpr size_t DEFAULT_FIFO_SIZE = 512;
 class Fifo {
  public:
   /// Constructor.
-  /// @param size Number of entries (default: DEFAULT_FIFO_SIZE).
+  /// @param size Number of entries. Must be a power of two (default: DEFAULT_FIFO_SIZE).
+  /// @throws Error with ErrorCode::InvalidUsage if size is not a positive power of two.
   Fifo(int size = DEFAULT_FIFO_SIZE);
 
   /// Destructor.
   ~Fifo();
 
-  /// Poll and get the trigger at the head.
-  /// @return ProxyTrigger at the head of the FIFO.
-  ProxyTrigger poll();
+  /// Poll for the trigger at the head.
+  ///
+  /// A trigger carries no reserved payload value, so readiness is reported separately rather than
+  /// encoded in the trigger itself.
+  ///
+  /// @param trigger Set to the trigger at the head if one is ready. Untouched otherwise.
+  /// @return True if a trigger was ready and written to @p trigger.
+  bool poll(ProxyTrigger& trigger);
 
   /// Remove the head trigger.
   void pop();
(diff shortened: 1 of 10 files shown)


Binyang2014 (commented):
Do we need to update  `python/test/test_mscclpp.py:529` for fifo.poll interface

Binyang2014 (approved):

Binyang2014 (approved):

PR #866, opened 2026-08-11
Introduce reduce-scatter algo

Binyang2014:
This pull request introduces a new hierarchical multi-node `ReduceScatter` algorithm for the low-latency packet protocol, improves thread block group support in the packet transfer API, and adds an `Executor::reset()` method to release cached execution contexts. It also refines memory registration logic for IB transport and enhances test coverage for executor resets.


python/mscclpp/default_algos/reducescatter_multi_nodes.py
```diff
@@ -0,0 +1,194 @@
...
+                        local_reduced_chunk,
+                        tb_group=thread_block_group,
+                    )
+                    remote_node_slot = src_node_id if src_node_id < owner_node_id else src_node_id - 1
+                    inter_node_channels[(owner_rank, src_rank)].read_put_packets(
+                        scratch_buffers[owner_rank][
+                            remote_receive_offset + remote_node_slot : remote_receive_offset + remote_node_slot + 1
+                        ],
+                        scratch_buffers[src_rank][local_packet_slot : local_packet_slot + 1],
+                        tb_group=thread_block_group,
```
caiocbr, line 157: Do we really need to use multiple thread blocks here?

python/mscclpp/default_algos/reducescatter_multi_nodes.py
```diff
@@ -0,0 +1,194 @@
...
+                        if src_rank != owner_rank:
+                            intra_node_channels[(owner_rank, src_rank)].put_packets(
+                                scratch_buffers[owner_rank][local_owner_offset : local_owner_offset + 1],
+                                local_reduced_chunk,
+                                tb_group=thread_block_group,
+                            )
+                        continue
+
+                    local_packet_slot = local_send_offset + chunk_offset
+                    rank.copy_packets(
```
caiocbr, line 143: I think there’s a fusion function when we copy first and then put packets. If we change the order and place copy packets before put packets, it enables some fuse instruction optimization.

python/mscclpp/default_algos/reducescatter_multi_nodes.py
```diff
@@ -0,0 +1,194 @@
...
+                        scratch_buffers[src_rank][local_packet_slot : local_packet_slot + 1],
+                        tb_group=thread_block_group,
+                    )
+
+        if num_nodes == 1:
+            return prog
+
+        # Every rank receives one standard shard. The owner-node handoff and
+        # direct IB transfers from the other nodes can progress concurrently.
+        for owner_rank in range(total_gpus):
```
caiocbr, line 164: I think we could use all the thread blocks for this work. Since the port channel operation will be handled by the CPU, all the threads should be free by the time we reach this step. My suggestion is to use a larger thread block group with all the thread blocks created for this operation. So, instead of using ```thread_block_group = thread_block_groups[local_reduce_offset + chunk_offset]```, I would create a large tbg and use ```thread_block_group = global_tbg```. I think this could improve performance, so it seems worth testing.

python/mscclpp/default_algos/reducescatter_multi_nodes.py
Binyang2014: I tested using only `thread_block_group.tb_list[0]` for `read_put_packets`, with the final reduction restored to its original group. Correctness passed at 2 KiB and 8 MiB, but the full TP64 benchmark regressed 8 MiB RSAG graph latency from 162.15 us to 183.06 us (+12.9%). I kept the multi-block partitioning.

python/mscclpp/default_algos/reducescatter_multi_nodes.py
Binyang2014: The remote path is already ordered as `reduce -> copy_packets -> read_put_packets`. With instruction fusion enabled, the serialized plan emits fused `reduce_copy_packet` (`recpkt`) before `read_put_packet` (`rppkt`) on all 64 remote thread blocks. I added a short comment documenting why this ordering must be preserved in 6b4ec50b.

python/mscclpp/default_algos/reducescatter_multi_nodes.py
Binyang2014: I tested a global group containing all logical thread blocks for the final unpack/reduce. TP64 hung while tuning the 2 KiB ReduceScatter candidate and all eight nodes timed out. Since this was a 1-TBG candidate, the single-TB PortChannel form was equivalent to the original group; the global final-reduction group was the differentiating change. I reverted it.

caiocbr (approved):
LGTM

PR #874, opened 2026-08-13
Refactor EP runtime architecture

Binyang2014:
Unify latency and overlap under one runtime, organize kernels by dispatch and combine algorithms, and expose one opaque dispatch handle.


python/mscclpp/ep/backend.py
```diff
@@ -0,0 +1,104 @@
...
+
+from typing import Optional
+
+import torch
+
+from ._cpp import DispatchLayout, MoEMode
+from .latency import _LatencyMethods
+from .overlap import _OverlapMethods
+from .runtime import Runtime
+from .types import DispatchHandle, DispatchOutput, MoECommunicatorConfig, QuantConfig
```
Binyang2014: user full path to import?

python/mscclpp/ep/types.py
```diff
@@ -63,7 +63,7 @@ class MoECommunicatorConfig:
     low_latency_combine_mode: CombineMode = CombineMode.RANK_LOCAL_REDUCE
```
Binyang2014: Can we rename this? We don't use low_latency now

python/mscclpp/ep/backend.py
```diff
@@ -0,0 +1,104 @@
...
+            self.runtime = Runtime(
+                config.comm,
+                MoEMode.LATENCY,
+                max_tokens_per_rank=config.max_tokens_per_rank,
+                hidden=config.hidden_size,
+                num_experts=config.num_experts,
+                num_topk=config.topk,
+                output_layout=output_layout,
+            )
+            self._init_latency(config, output_layout)
```
Binyang2014: _init_latency this name not good

python/mscclpp/ep/backend.py
```diff
@@ -0,0 +1,104 @@
...
+        else:
+            max_hidden_bytes = config.hidden_size * torch.empty((), dtype=torch.bfloat16).element_size()
+            self.runtime = Runtime(
+                config.comm,
+                MoEMode.OVERLAP,
+                max_hidden_bytes=max_hidden_bytes,
+                num_sms=config.num_sms,
+            )
+            self._init_overlap(config, output_layout)
```
Binyang2014: THis name not good

python/mscclpp/ep/backend.py
```diff
@@ -0,0 +1,104 @@
...
+            self.expert_output_buffer = None
```
Binyang2014: Why need this? We have unified output_buffer dispatch_output_buffer?

python/mscclpp/ep/communicator.py
```diff
@@ -152,7 +135,7 @@ def get_expert_output_buffer(self) -> torch.Tensor:
         """
         buffer = getattr(self._backend, "expert_output_buffer", None)
         if buffer is None:
-            raise RuntimeError("expert output buffer is only available for RANK_MAJOR low-latency mode")
+            raise RuntimeError("expert output buffer is only available for RANK_MAJOR latency mode")
         return buffer
```
Binyang2014: WHy named get_expert_output_buffer, we need to rename get dispatch_output_buffer. As it will be used for expert major/rank major output buffer

python/mscclpp/ep/latency.py
```diff
@@ -107,54 +105,20 @@ def _tensor_from_pointer(
...
-
-    def is_internode_available(self) -> bool:
-        return self.cpp_runtime.is_internode_available()
-
-
-class LowLatencyBackend:
-    """Backend implementation for ``MoEMode.LOW_LATENCY``."""
-
-    def __init__(self, config: MoECommunicatorConfig, output_layout: DispatchLayout) -> None:
+    def _init_latency(self, config: MoECommunicatorConfig, output_layout: DispatchLayout) -> None:
```
Binyang2014: WHy not jut __init__? init_lantecy feel strange

python/mscclpp/ep/latency.py
```diff
@@ -377,40 +324,38 @@ def dispatch(
                 ),
             )
         else:
-            raise ValueError(f"unsupported low-latency output layout: {self.output_layout}")
+            raise ValueError(f"unsupported latency output layout: {self.output_layout}")
         return dispatch_out, handle
 
-    def combine(
+    def _combine_latency(
```
Binyang2014: Can not just name as combine as already in latency.py

python/mscclpp/ep/latency.py
```diff
@@ -538,10 +485,12 @@ def _validate_dispatch_inputs(self, input, topk_ids, weights, quant, output_buff
         if tuple(output_buffer.shape) != expected_shape:
             raise ValueError(f"output_buffer shape must be {expected_shape}")
 
-    def _validate_combine_inputs(self, expert_output, handle, out) -> None:
-        if not isinstance(handle, (ExpertMajorDispatchHandle, RankMajorDispatchHandle)):
-            raise ValueError("DispatchHandle does not contain low-latency combine context")
-        context = handle.combine_context
+    def _validate_latency_combine_inputs(self, expert_output, handle, out) -> None:
```
Binyang2014: Maybe no need latency word here?

python/mscclpp/ep/backend.py
Binyang2014: Updated backend imports to use the full mscclpp.ep package paths.

python/mscclpp/ep/types.py
Binyang2014: Renamed the public fields to latency_num_blocks and combine_mode; no low_latency names remain.

python/mscclpp/ep/backend.py
Binyang2014: Replaced _init_latency with the mode-local _Latency.__init__.

python/mscclpp/ep/backend.py
Binyang2014: Replaced _init_overlap with the mode-local _Overlap.__init__.

python/mscclpp/ep/backend.py
Binyang2014: Removed the overlap-mode expert_output_buffer placeholder. Fixed-buffer dispatch storage now exists only on the latency implementation.

python/mscclpp/ep/communicator.py
Binyang2014: Added get_dispatch_output_buffer() for both expert-major and rank-major dispatch destinations. Rank-major still needs a second registered buffer for the MLP result that peers read during combine, so that distinct buffer is now named get_combine_input_buffer() rather than expert output.

python/mscclpp/ep/latency.py
Binyang2014: Changed the latency implementation to use __init__ directly.

python/mscclpp/ep/latency.py
Binyang2014: Renamed the mode-local method to combine; Backend explicitly dispatches to _Latency.combine or _Overlap.combine.

python/mscclpp/ep/latency.py
Binyang2014: Removed the redundant latency qualifier; the method is now _validate_combine_inputs inside latency.py.

python/mscclpp/ep/backend.py
Binyang2014: Confuse here? Call init directly? It's statis init? No need to create instance?

python/mscclpp/ep/backend.py
Binyang2014: Same here? Call init directly is confused

Binyang2014:
Addressed the latest review in 2090bd52:

- Backend now uses one standard `super().__init__(config, output_layout)` path; there are no direct class `__init__` calls and no separate mode instances.
- Removed `get_combine_input_buffer()`. Rank-major dispatch returns the runtime-owned MLP/combine destination as `dispatch_out.combine_input_buffer`, tying it to the dispatch result and active layout.

Binyang2014:
Updated backend hierarchy in db0cb303: `Backend` is now an abstract parent, `create_backend()` selects `LatencyBackend` or `OverlapBackend` from `MoEMode`, and each concrete backend owns the same unified C++ `Runtime` type. The previous multiple-inheritance/MRO initialization is removed.

Binyang2014:
Updated in 5d752e36: all internal Python imports under `mscclpp.ep` now use full absolute package paths.

Binyang2014:
Updated in 61eb952e: `create_backend()` and concrete backend constructors now accept only `MoECommunicatorConfig`; each backend reads and resolves `config.output_layout` directly. The duplicate layout plumbing is removed.

Binyang2014:
Updated in e7fb5860: runtime-owned generic, BF16, and FP8 CUDA pointer tensor views are now centralized in `mscclpp.ep.utils`; `latency.py` only selects layout/dtype and owns mode behavior.

Binyang2014:
Updated in f4377dd8: removed the latency-only dispatch output getter. Every mode now exposes dispatch output uniformly through `dispatch_out.tokens`; runtime allocation remains internal.

Binyang2014:
Updated in 6e44bb75: dispatch datatype resolution and scale metadata helpers are now common `mscclpp.ep.utils` functions. `latency.py` retains only latency-mode support policy and execution.

Binyang2014:
Updated in a67f7942: BF16 and FP8 pointer wrappers are merged into one dtype-driven `tensor_from_pointer()` helper, also used for INT32/FP32 metadata.

Binyang2014:
Updated in 5125a93c: removed the redundant `bf16_view()` wrapper; overlap mode now calls the generic dtype-driven `tensor_from_pointer()` directly.

Binyang2014:
Updated in 4faa42cf: removed `LOW_LATENCY` and `HIGH_THROUGHPUT` compatibility aliases. MSCCL++ now exposes only `MoEMode.LATENCY` and `MoEMode.OVERLAP`.

Binyang2014:
Updated in 3c9ef8b1: removed mode suffixes throughout the EP surface. Fixed-buffer algorithms use qualifier-first names (`latencyExpertMajor`, `latencyRankMajor`, etc.); token-major APIs use layout-first names (`tokenMajorPrepare`, `tokenMajorDispatch`, `tokenMajorReduce`).

Binyang2014:
Updated in 5b8ae8aa: `api.cuh` now documents each expert-major, rank-major, and token-major dispatch/combine contract separately. `DeviceContext` also carries its persistent `devicePtr_`, so every API accepts one context argument instead of separate host/device context parameters.

Binyang2014:
Updated in 21f74b9e: flattened kernel sources directly under `dispatch/` and `combine/`; backend/runtime selection now owns mode policy. Fixed-buffer APIs use `expertMajorDispatch`/`rankMajorDispatch`, while token-major remains `tokenMajorPrepare`/`tokenMajorDispatch`/`tokenMajorReduce`.

Binyang2014:
Updated in 654e7079: split the generic combine API into layout-specific entry points: `expertMajorRankLocalReduce`, `rankMajorGatherReduce`, and `expertMajorDirectSend`.

Binyang2014:
Updated in f0b8759e: all combine kernels and entry points now use a `Combine` suffix, matching dispatch naming (`expertMajorLocalReduceCombine`, `rankMajorGatherReduceCombine`, `expertMajorDirectSendCombine`, `tokenMajorReduceCombine`).

Binyang2014:
Updated in 50246806: replaced `latency_num_blocks` and `num_sms` tuning with one optional `num_blocks` field (defaults: 130 latency, 20 overlap). Test-only NCCL monitoring overrides were also removed; hardware SM-count fields remain internal.

caiocbr (approved):
LGTM

PR #875, opened 2026-08-14
Fix low-latency EP epoch race with host-assigned dispatch epochs

caiocbr:
## Summary

Assign low-latency dispatch epochs once in `MoELowLatencyRuntime` and pass the epoch to dispatch and combine through `Workload`.

Previously, each CUDA block derived the epoch from device workspace state while block 0 updated that state without a grid-wide ordering guarantee. Late blocks could therefore observe a different epoch and wait indefinitely on readiness flags.

## Changes

- Add a host-owned dispatch epoch counter to `MoELowLatencyRuntime`.
- Increment the epoch once before each dispatch.
- Pass the same epoch to dispatch and its subsequent combine operation.
- Use the workload epoch for rank-major synchronization.
- Remove device-side epoch generation, publication, and workspace storage.

This gives every block in a launch one immutable epoch value.

## Validation

- Passed eager rank-major tests at capacities 1 and 2.
- Passed eager expert-major tests at capacity 2.
- Passed bit-exact 2-rank and 4-rank multi-operation CUDA Graph replays.
- Passed all 10 SGLang dispatcher tests.
- Completed six 32-rank serving rounds with 40/40 responses and no scheduler watchdog timeouts.

### DIFF
--- a/src/ext/ep/low_latency/combine.cu
+++ b/src/ext/ep/low_latency/combine.cu
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
+#include <mscclpp/bulk_device.hpp>
 #include <mscclpp/gpu_data_types.hpp>
 
 #include "api.cuh"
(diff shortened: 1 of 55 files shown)


src/ext/ep/low_latency/combine.cu
```diff
@@ -286,7 +286,7 @@ MSCCLPP_DEVICE_INLINE void exchangeCombineReady(const TransportView& transport,
 }
 
 MSCCLPP_DEVICE_INLINE void synchronizeRankMajorCombine(const TransportView& transport, int nRanks,
-                                                       WorkspaceView& workspaceView) {
+                                                       uint32_t combineEpoch, WorkspaceView& workspaceView) {
```
Binyang2014: Since dispatch/combine use same value for epoch, we can rename the var to epoch. Not specific to dispatch/combien

Binyang2014 (approved):

PR #877, opened 2026-08-18
add allgather mnnvl test to ci pipeline

Empyreus:


Binyang2014 (approved):
LGTM, this algo works for both in-place and out-of-place right? Did you try to use switchChannel  semaphore for the synchronization

Empyreus:
> LGTM, this algo works for both in-place and out-of-place right? Did you try to use switchChannel semaphore for the synchronization

I haven't tried with those options. I can add some tests if we would like.

.azure-pipelines/mnnvl-test.yml
```diff
@@ -126,9 +126,15 @@ jobs:
   - template: templates/run-remote-task.yml
     parameters:
       name: RunCollectivebenchmarks
-      displayName: Run Collective Benchmarks
+      displayName: Run Collective Benchmarks
```
Binyang2014: Remove the space at the end?

Binyang2014 (approved):

Binyang2014 (approved):

PR #881, opened 2026-08-21
feat(allgather): prepared canonical row-major contexts

chhwang:
## Summary
- add typed, explicit pre-capture native contexts for all-gather
- make capture/replay lookup-only and fail closed on key mismatch
- add direct canonical row-major logits mapping without transpose staging
- preserve legacy rank-major mode
- make replay-visible device counters optional/null by default

## Validation
- fast C++/CUDA/nanobind compile gates
- arm64 wheel SHA-256 `9444497664ea978c2b67c3d544275587c603e90427fa0c9150026b938edfa168`
- 4-GPU zero/small/all eight GLM graph buckets
- bitwise canonical row-major oracle and adversarial legacy-layout rejection
- CUDA graph capture/replay, pointer/size/dtype/layout misses, release/reprepare, peer-failure cleanup
- TP16 stock-vs-candidate bitwise BF16 equality on 16 ranks/all buckets

## Stack note
The first commit materializes the accepted Version B native overlay tree `7dede6fe...` over the retained EP base because accepted commit `6abd2bb...` was delivered as a deterministic packaging overlay rather than a public commit object.

## Performance
Functional baseline only. End-to-end results are neutral within run/order noise; p99 remains noisy. This is the foundation for later communication KOP work, not a claimed speedup.


chhwang:
Closing because the TP logits functionality was reimplemented and qualified in IE-native [ShrinkWrap-Inference-Engine PR44](https://github.com/microsoft/ShrinkWrap-Inference-Engine/pull/44). MSCCL++ remains unchanged by design. This does not claim that PR881 code was merged.

PR #886, opened 2026-08-25
Add stream operators for TransportFlags and ErrorCode

ooooo-create:


ooooo-create:

src/core/errors.cc
```diff
@@ -15,6 +15,8 @@ std::string errorToString(enum ErrorCode error) {
       return "SystemError";
     case ErrorCode::InternalError:
       return "InternalError";
+    case ErrorCode::RemoteError:
+      return "RemoteError";
```
Binyang2014, line 19: Why we add this errorCode, we use it in somewhere?

ooooo-create: https://github.com/microsoft/mscclpp/blob/main/include%2Fmscclpp%2Ferrors.hpp#L12-L20
Hi, I see it in the definition of ErrorCode, it missed that before, so I add it.

Binyang2014 (approved):

PR #895, opened 2026-09-03
[MSCCLPP-EP] Add C++ latency expert-parallel runtime

Binyang2014:
## Summary
- add the public C++ `MoERuntime` API and latency dispatch/combine request types
- add the current low-latency expert-major and rank-major CUDA dispatch/combine kernels
- support BF16 and FP8 dispatch plus rank-local-reduce and direct-send combine modes
- build the implementation as a standalone `mscclpp_ep` shared library with no Python dependency
- add 8-GPU SM90+ C++ initialization, correctness, and CUDA-graph performance tests

## Scope
- empty `ThroughputDispatchRequest` and `ThroughputCombineRequest` types reserve names for the follow-up implementation
- throughput request fields, `prepare`, `notify`, host runtime, and CUDA kernels are intentionally excluded
- all Python bindings, wrappers, benchmarks, and tests are intentionally excluded for follow-up PRs

## C++ test coverage
- runtime initialization and unsupported throughput-mode validation
- expert-major BF16 and FP8 with rank-local-reduce and direct-send combine
- rank-major BF16 with rank-local-reduce and direct-send combine
- CUDA-graph D+C performance for expert-major and rank-major, 32 tokens/rank, hidden 7168, 8 experts, top-8
- tests skip unless launched with exactly eight local GPUs at SM90 or newer

## Validation
- `./tools/lint.sh`
- C++-only `mscclpp_ep` builds for `MSCCLPP_GPU_ARCHS=90` and `native`
- Python-enabled configure builds both `mscclpp_ep` and `mscclpp_py`
- external C++ consumer compiles and links against the installed public API
- install contains `libmscclpp_ep.so` and the public EP headers
- `mpirun -np 8 ./build-ep-tests/bin/mp_unit_tests --filter=MoERuntimeTest`: 4/4 passed
- graph D+C: expert-major 51.08 us, rank-major 40.95 us
## Feature/EP regression comparison
- identical 8-GPU C++ harness, hidden 7168, 32 tokens/rank, BF16, rank-local-reduce, 50 D+C pairs/graph, 100 replays
- split branch averages: expert-major 50.47 us, rank-major 40.75 us
- `feature/ep` averages: expert-major 50.58 us, rank-major 41.21 us
- six dispatch/combine correctness combinations pass on both branches


src/ext/ep/latency.cc
```diff
@@ -0,0 +1,286 @@
...
+  const DispatchLayout dispatchLayout = request.dispatchLayout;
+  const DispatchDataType dispatchDataType = request.dispatchDataType;
+  const CombineMode mode = request.combineMode;
+  const int numBlocks = request.numBlocks;
+  const cudaStream_t stream = request.stream;
+
+  auto& context = *latencyContext_;
+  EP_HOST_ASSERT(context.available_);
+  EP_HOST_ASSERT(context.deviceContext_.devicePtr_ != nullptr);
+  EP_HOST_ASSERT(maxTokensPerRank > 0 && maxTokensPerRank <= context.maxTokensPerRank_);
```
mahdiehghazim, line 241: Should we also validate numTokens <= maxTokensPerRank here, consistent with launchLatencyDispatch()?

src/ext/ep/latency.cc
```diff
@@ -0,0 +1,286 @@
...
+  EP_HOST_ASSERT(numExperts % context.numRanks_ == 0);
+  EP_HOST_ASSERT(numBlocks > 0 && numBlocks <= MaxWorkerBlocks);
+  EP_HOST_ASSERT(dispatchLayout == context.outputLayout_);
+  EP_HOST_ASSERT(mode == context.combineMode_);
+
+  LatencyStorageLayout allocationLayout(context.symmetricBuffer_, context.maxTokensPerRank_, hidden, context.numRanks_,
+                                        numExperts, numTopk, context.outputLayout_, context.combineMode_);
```
mahdiehghazim, line 248: Are hidden, numExperts, and numTopk intended to vary between requests? I noticed they're specified both when constructing the runtime and in each request, and the request values don't appear to be required to match the runtime values

src/ext/ep/latency.cc
```diff
@@ -0,0 +1,286 @@
...
+  EP_HOST_ASSERT(allocationLayout.totalBytes_ <= static_cast<size_t>(context.symmetricBufferBytes_));
+  void* combineRecvBuffer = allocationLayout.combineRecvBuffer_;
+  void* dispatchRecvBuffer = allocationLayout.dispatchRecvBuffer_;
+  if (dispatchLayout == DispatchLayout::RANK_MAJOR) {
+    EP_HOST_ASSERT(input == allocationLayout.combineRecvBuffer_);
+  }
+
+  const Workload workload{.epoch_ = context.epoch_,
```
mahdiehghazim, line 256: Does combine need to validate that its workload parameters match the preceding dispatch? Since it reuses the dispatch routing metadata/epoch, I'm wondering what happens if numTokens or other workload parameters differ between the two calls.

test/mp_unit/ep_tests.cu
```diff
@@ -0,0 +1,540 @@
...
+#include "mp_unit_tests.hpp"
+
+namespace {
+
+using Bf16 = typename mscclpp::bf16x2::ElementType;
+using Fp8E4M3 = typename mscclpp::f8_e4m3x2::ElementType;
+
+constexpr int NumRanks = 8;
+constexpr int NumExperts = 8;
+constexpr int NumTopk = 8;
```
mahdiehghazim, line 27: Could we add a couple more routing configurations here. we can also do it in a separate PR.

1. A case with `NumExperts > NumTopk` (for example, more experts per rank / a larger global expert count), since the current `NumExperts == NumTopk == 8` means every token selects every expert.

2. A nonuniform routing case where different experts receive different numbers of tokens, including potentially some experts receiving no tokens. This would help exercise the dispatch/combine paths under more realistic expert-load distributions.

src/ext/ep/latency.cc
```diff
@@ -0,0 +1,286 @@
...
+  EP_HOST_ASSERT(outputLayout == DispatchLayout::EXPERT_MAJOR || outputLayout == DispatchLayout::RANK_MAJOR);
+
+  MSCCLPP_CUDATHROW(cudaGetDevice(&deviceId_));
+  EP_HOST_ASSERT(numRanks_ % numNvlRanks == 0);
+  EP_HOST_ASSERT(numRanks_ % numRanksPerIpcDomain_ == 0);
+  available_ = numRanksPerIpcDomain_ >= numRanks_;
+}
+
+LatencyRuntimeContext::~LatencyRuntimeContext() noexcept(false) {
+  MSCCLPP_CUDATHROW(cudaDeviceSynchronize());
```
mahdiehghazim, line 48: Is the device-wide synchronization intentional here? `cudaDeviceSynchronize()` also waits for unrelated GPU work. Could we synchronize only the EP work before freeing the runtime resources?

src/ext/ep/latency.cc
```diff
@@ -0,0 +1,286 @@
...
+      outputLayout_(outputLayout),
+      combineMode_(combineMode),
+      symmetricBufferBytes_(static_cast<int64_t>(
+          latencyStorageSize(maxTokensPerRank, hidden, numRanks_, numExperts, numTopk, outputLayout, combineMode))),
+      workspaceBytes_(workspaceSize(numRanks_, numExperts, maxTokensPerRank, numTopk)),
+      communicator_(communicator) {
+  EP_HOST_ASSERT(symmetricBufferBytes_ % BufferAlignmentBytes == 0);
+  EP_HOST_ASSERT(maxTokensPerRank > 0);
+  EP_HOST_ASSERT(numExperts > 0 && numExperts % numRanks_ == 0);
+  EP_HOST_ASSERT(numTopk > 0 && numTopk <= MaxNumTopk);
```
mahdiehghazim, line 38: Should we validate `hidden` against the supported latency hidden sizes here? The runtime allocates its buffers during construction, but unsupported hidden sizes are only rejected later when dispatch selects the kernel.

include/mscclpp/ext/ep/types.hpp
```diff
@@ -0,0 +1,167 @@
...
+  /// Hidden dimension.
+  int hidden;
+  /// Number of routed experts per token.
+  int numTopk;
+  /// Active per-rank token capacity.
+  int maxTokensPerRank;
+  /// Global expert count.
+  int numExperts;
+  /// Expert ID used for invalid rank-major entries.
+  int invalidTokenExpertId;
```
mahdiehghazim, line 87: Could we document the expected value/range for invalidTokenExpertId?

test/mp_unit/ep_tests.cu
Binyang2014: Use random seed to generate the expert id for this

src/ext/ep/latency.cc
Binyang2014: Update the API, let dispatch return a handle and combine use this handle as the input

src/ext/ep/latency.cc
Binyang2014: Thanks, fix this issue. Remove the duplicated fields in dispatch/combine reuqests

include/mscclpp/ext/ep/types.hpp
```diff
@@ -0,0 +1,161 @@
...
+  /// Combined token output.
+  void* output;
+  /// Local expert output.
+  const void* input;
+  /// Handle returned by the matching dispatch.
+  DispatchHandle handle;
+  /// Combine grid block count.
+  int numBlocks;
+  /// CUDA stream used for the operation.
+  cudaStream_t stream;
```
mahdiehghazim, line 143: Can dispatch and combine use different CUDA streams? If so, does the runtime establish the dependency between them internally, or is the caller expected to synchronize the streams before combine? It may be useful to document the ordering requirement here.

include/mscclpp/ext/ep/types.hpp
```diff
@@ -0,0 +1,161 @@
...
+  explicit DispatchRequest(ThroughputDispatchRequest request) : value_(std::move(request)) {}
+
+ private:
+  friend class MoERuntime;
+  std::variant<LatencyDispatchRequest, ThroughputDispatchRequest> value_;
+};
+
+/// Opaque metadata returned by a successful dispatch.
+///
+/// A handle borrows the dispatch's routing buffers and runtime resources.
```
mahdiehghazim, line 117: Could we clarify the lifetime requirement for the borrowed routing buffers here? Do they need to remain valid until the matching combine has completed on the GPU?

include/mscclpp/ext/ep/types.hpp
Binyang2014: Caller is responsible for this. I think it's fine to keep it as it is. Same for the other op

include/mscclpp/ext/ep/types.hpp
Binyang2014: addressed in the new commit

mahdiehghazim (approved):

include/mscclpp/ext/ep/moe_runtime.hpp
```diff
@@ -0,0 +1,130 @@
...
+ private:
+  void requireMode(MoEMode expected) const;
+  DispatchHandle launchLatencyDispatch(const LatencyDispatchRequest& request);
+  void launchLatencyCombine(const LatencyCombineRequest& request);
+
+  std::shared_ptr<mscclpp::Bootstrap> bootstrap_;
+  MoEMode mode_;
+  int rank_;
+  int numRanks_;
+  int numNvlRanks_;
```
caiocbr: Could we rename `numNvlRanks_` to `numRanksPerNode_`? It is derived from `getNranksPerNode()`, so the current name implies an NVLink topology property that is not actually established here. We could potentially remove this field instead, since the current latency implementation uses `numRanksPerIpcDomain_` for its communication-domain and availability decisions.

include/mscclpp/ext/ep/moe_runtime.hpp
```diff
@@ -0,0 +1,130 @@
...
+  /// @throws EPException If @p request is invalid or its handle is empty, expired,
+  /// stale, or belongs to another runtime.
+  void combine(const CombineRequest& request);
+
+ private:
+  void requireMode(MoEMode expected) const;
+  DispatchHandle launchLatencyDispatch(const LatencyDispatchRequest& request);
+  void launchLatencyCombine(const LatencyCombineRequest& request);
+
+  std::shared_ptr<mscclpp::Bootstrap> bootstrap_;
```
caiocbr: Since we pass the communicator through the constructor, wouldn’t it make sense to retain it directly? I realize that most of the time we essentially use only the bootstrap, but in other codebases we generally keep the communicator rather than retaining only the bootstrap.

include/mscclpp/ext/ep/types.hpp
```diff
@@ -0,0 +1,164 @@
...
+  BF16,
+  /// FP8 E4M3 payload with one floating-point scale per 128 hidden elements.
+  FP8_E4M3
+};
+
+/// Arguments for latency-mode dispatch.
+///
+/// The caller must keep all buffers referenced by this struct (inputs and
+/// outputs) valid until the corresponding combine operation has finished.
+struct LatencyDispatchRequest {
```
caiocbr, line 59: Would it be clearer and safer to split this into layout-specific request types (for example, `ExpertMajorDispatchRequest` and `RankMajorDispatchRequest`) and make `LatencyDispatchRequest` a variant or shared/common portion? The current struct permits invalid combinations and relies on runtime assertions even though several fields are mutually exclusive:

- Expert-major only: `outputScales` (FP8 only), `outputSrcInfo`, and `outputLayoutRange`.
- Rank-major only: `outputTopkIdx`, `outputTopkWeights`, and `invalidTokenExpertId`.
- Both layouts: `output`, `outputCount`, `input`, `topkIdx`, `topkWeights`, `numTokens`, `maxTokensPerRank`, `dispatchDataType`, `numBlocks`, and `stream`; however, `outputCount` has layout-dependent indexing, and FP8 `dispatchDataType` is currently supported only by expert-major.

Separate request types would make these requirements explicit in the type system, document the different output shapes, and avoid exposing fields that callers must set to `nullptr` for the selected layout.

include/mscclpp/ext/ep/types.hpp
```diff
@@ -0,0 +1,164 @@
...
+  /// Here numExperts is the runtime's configured global expert count.
+  /// Must be outside [0, numExperts): use a negative int (e.g., -1) or an
+  /// int >= numExperts (e.g., numExperts). All ranks must use the same sentinel.
+  /// This is an output sentinel; input topkIdx entries must still be valid
+  /// global expert IDs or negative values for dropped routes.
+  int invalidTokenExpertId = -1;
+  /// Requested dispatch payload format.
+  DispatchDataType dispatchDataType;
+  /// Dispatch grid block count.
+  int numBlocks;
```
caiocbr: Does `numBlocks` need to be part of every dispatch request? This looks like an implementation-specific tuning parameter whose valid and optimal value depends on the selected algorithm, device occupancy, rank count, and shared-memory use. Could the runtime/algorithm choose a standard value by default (for example, treating `0` as automatic), while retaining an optional override for benchmarking or advanced tuning? That would keep ordinary callers from needing to understand the kernel's control-block and all-block-residency constraints.

include/mscclpp/ext/ep/types.hpp
```diff
@@ -0,0 +1,164 @@
...
+/// Arguments for latency-mode combine.
+struct LatencyCombineRequest {
+  /// Combined token output.
+  void* output;
+  /// Local expert output.
+  const void* input;
+  /// Handle returned by the matching dispatch.
+  DispatchHandle handle;
+  /// Combine grid block count.
+  int numBlocks;
```
caiocbr: Similarly, should combine block count default to an algorithm-selected value rather than being mandatory in every request? A default such as `0` meaning automatic would let the selected combine implementation derive a safe value from occupancy, layout, and device properties, while still allowing an explicit override for performance tuning.

src/ext/ep/common/latency.cuh
```diff
@@ -0,0 +1,227 @@
...
+#include <mscclpp/memory_channel_device.hpp>
+
+#include "config.hpp"
+#include "device_helpers.cuh"
+#include "kernels.hpp"
+
+namespace mscclpp {
+namespace ep {
+
+constexpr int DispatchNWarps = 16;
```
caiocbr, line 19: Could you clarify why dispatch uses 16 warps (512 threads) rather than the maximum 32 warps (1024 threads) per block? Is this an intentional performance choice based on benchmarking—for example, to balance register/shared-memory pressure, block residency, or TMA throughput? Using fewer threads can certainly perform better here, but it would be helpful to document the rationale and which workload or GPU configurations were used to select 16.

src/ext/ep/include/exception.hpp
```diff
@@ -0,0 +1,48 @@
...
+#include <exception>
+#include <mscclpp/assert_device.hpp>
+#include <mscclpp/gpu_utils.hpp>
+#include <string>
+
+#ifndef EP_STATIC_ASSERT
+#define EP_STATIC_ASSERT(cond, reason) static_assert(cond, reason)
+#endif
+
+class EPException : public std::exception {
```
caiocbr: Do we need an EP-specific exception type and this separate exception header? MSCCL++ already has the public `mscclpp::Error` hierarchy with typed `ErrorCode` values, along with `MSCCLPP_CUDATHROW`, `MSCCLPP_CUTHROW`, and `MSCCLPP_ASSERT_DEVICE`. It may be more consistent to use `mscclpp::Error(..., ErrorCode::InvalidUsage)` for invalid requests and `ErrorCode::InternalError` for failed internal invariants, keeping only minimal EP helper macros if needed. This would also avoid the current public-API mismatch where public methods document `EPException`, but that type is defined only in this private `src` header and therefore cannot be caught by users through the installed public headers.

src/ext/ep/latency.cc
```diff
@@ -0,0 +1,303 @@
...
+                                        numExperts, numTopk, context.outputLayout_, context.combineMode_);
+  EP_HOST_ASSERT(allocationLayout.totalBytes_ <= static_cast<size_t>(context.symmetricBufferBytes_));
+  void* dispatchRecvBuffer = allocationLayout.dispatchRecvBuffer_;
+  if (dispatchLayout == DispatchLayout::RANK_MAJOR) {
+    EP_HOST_ASSERT(output == allocationLayout.dispatchOutputBuffer_);
+    EP_HOST_ASSERT(outputTopkIdx == allocationLayout.rankMajorTopkIdsBuffer_);
+    EP_HOST_ASSERT(outputTopkWeights == allocationLayout.rankMajorTopkWeightsBuffer_);
+  }
+
+  const Workload workload{.epoch_ = context.epoch_ + 1,
```
caiocbr: Should we update the context epoch here—or perhaps use an atomic add—to prevent two kernels from being called with the same epoch if one don't update before the other arrives?

include/mscclpp/ext/ep/types.hpp
```diff
@@ -0,0 +1,164 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+#ifndef MSCCLPP_EXT_EP_TYPES_HPP_
+#define MSCCLPP_EXT_EP_TYPES_HPP_
+
+#include <cuda_runtime.h>
```
chhwang: Please use `mscclpp/gpu.hpp` instead.

src/ext/ep/include/exception.hpp
Binyang2014: I prefer EP specific exception here. More clear to show to the error scope. Also we treat ep as an extension here, I don't want to add new log subsys to mscclpp core. Maybe we can inherit mscclpp exception here

include/mscclpp/ext/ep/types.hpp
Binyang2014: I think we don't need split this into several layouts. OutputScale is not limited to expert-major, it could be used for rank major also. Let keep as it is for now.

src/ext/ep/latency.cc
Binyang2014: Our function is not thread safe. It's users' responsibility to make it saft if there are concurrent function calls

include/mscclpp/ext/ep/types.hpp
Binyang2014: Resolved

src/ext/ep/common/latency.cuh
Binyang2014: For TAM load we just need few warp. Other warps just for quantization and some copy instruction. 512 threads per blocks is already good for most of case based on the benchmarking results

caiocbr (approved):
Looks good to me. One non-blocking architectural consideration for future work: the current mode-centric runtime context may become difficult to scale as more algorithms are added for the same mode and layout, since algorithm-specific initialization, buffer planning, resource ownership, context management, and dispatch/combine selection would accumulate in a generic context. A design similar in spirit to NativeAlgorithm could keep MoERuntime as the public facade while delegating to a selected MoEAlgorithm that owns the complete compatible dispatch/combine protocol, including construction, collective initialization, resources, context, execution, and cleanup. Common transport pieces could still be composed and shared across implementations. Beyond this seems we still have room for some small optimization, in the thread block allocation in the LL dispatch and combine algorithms for Rank Major we are using 1 thread block only for signaling when the tokens are send.

PR #898, opened 2026-09-04
Port token-major latency EP path

mahdiehghazim:
Enable latency-mode TOKEN_MAJOR dispatch/combine support and wire the unified Python EP benchmark to exercise the new layout.

### DIFF
--- a/python/mscclpp/ep/latency.py
+++ b/python/mscclpp/ep/latency.py
@@ -422,10 +422,14 @@ class LatencyRuntime(Runtime):
         )
         if type(resolved) is not int or not 0 < resolved <= mode_context.max_tokens_per_rank:
             raise ValueError("runtime_max_tokens_per_rank must be positive and not exceed max_tokens_per_rank")
-        if mode_context.output_layout not in (
-            DispatchLayout.RANK_MAJOR,
-            DispatchLayout.TOKEN_MAJOR,
-        ) and resolved != mode_context.max_tokens_per_rank:
+        if (
+            mode_context.output_layout
+            not in (
+                DispatchLayout.RANK_MAJOR,
+                DispatchLayout.TOKEN_MAJOR,
+            )
+            and resolved != mode_context.max_tokens_per_rank
+        ):
             raise ValueError("runtime_max_tokens_per_rank is only supported by rank-major dispatch")
         return resolved
 
@@ -607,7 +611,9 @@ class LatencyRuntime(Runtime):
         if handle.output_info.layout.kind in (DispatchLayout.RANK_MAJOR, DispatchLayout.TOKEN_MAJOR):
             assert mode_context.combine_input_buffer is not None
             if expert_output.data_ptr() != mode_context.combine_input_buffer.data_ptr():
-                raise ValueError(f"{handle.output_info.layout.kind} combine requires the runtime-owned combine input buffer")
+                raise ValueError(
+                    f"{handle.output_info.layout.kind} combine requires the runtime-owned combine input buffer"
+                )
         if out is not None:
             expected_out_shape = (context.num_tokens, mode_context.hidden_size)
             if tuple(out.shape) != expected_out_shape or out.dtype != expected_dtype or not out.is_contiguous():
--- a/python/mscclpp/ep/latency.py
+++ b/python/mscclpp/ep/latency.py
@@ -20,12 +20,15 @@ from mscclpp.ep.types import (
     QuantConfig,
     _ExpertMajorCombineContext,
     _RankMajorCombineContext,
+    _TokenMajorCombineContext,
 )
 from mscclpp.ep.utils import (
     DevicePointerArray,
+    combine_tensor_dtype,
     cuda_stream_ptr,
     dispatch_scale_block_size,
     dispatch_scale_dtype,
+    dispatch_tensor_dtype,
     resolve_expert_placement,
     resolve_dispatch_data_type,
     resolve_num_blocks,
@@ -73,6 +76,7 @@ class LatencyContext(Context):
         if self.output_layout not in (
             DispatchLayout.EXPERT_MAJOR,
             DispatchLayout.RANK_MAJOR,
+            DispatchLayout.TOKEN_MAJOR,
         ):
             raise NotImplementedError("unsupported latency output layout")
         if self.num_experts % self.world_size != 0:
@@ -97,6 +101,11 @@ class LatencyContext(Context):
                 raise ValueError("RANK_MAJOR output requires a supported combine mode")
             if self.enable_overlap:
                 raise NotImplementedError("RANK_MAJOR output does not support overlapping calls yet")
+        if self.output_layout == DispatchLayout.TOKEN_MAJOR:
+            if self.combine_mode != CombineMode.RANK_LOCAL_REDUCE:
+                raise ValueError("TOKEN_MAJOR output requires RANK_LOCAL_REDUCE combine")
+            if self.enable_overlap:
+                raise NotImplementedError("TOKEN_MAJOR output does not support overlapping calls yet")
 
         self.num_local_experts, self.local_expert_start = resolve_expert_placement(
             num_experts=self.num_experts,
@@ -109,6 +118,8 @@ class LatencyContext(Context):
         self.dispatch_data_type = resolve_dispatch_data_type(config.quant)
         if self.output_layout == DispatchLayout.RANK_MAJOR and self.dispatch_data_type != DispatchDataType.BF16:
             raise NotImplementedError("RANK_MAJOR output currently supports BF16 dispatch only")
+        if self.output_layout == DispatchLayout.TOKEN_MAJOR and self.dispatch_data_type != DispatchDataType.BF16:
+            raise NotImplementedError("TOKEN_MAJOR output currently supports BF16 dispatch only")
 
         self._dispatch_scales: Optional[torch.Tensor] = None
         self._dispatch_src_info: Optional[torch.Tensor] = None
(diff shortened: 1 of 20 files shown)


Binyang2014:
@copilot+gpt-5.6-sol:max, pls review the pr

Binyang2014 (commented):
**Agent-generated review (GitHub Copilot).**

These findings were independently rechecked against the current PR head and the surrounding device/host contracts.

python/mscclpp/ep/latency.py
```diff
@@ -445,8 +501,9 @@ def _validate_dispatch(self, input, topk_ids, weights, quant, output_buffer, act
             )
         if input.dim() != 2 or not input.is_contiguous():
             raise ValueError("input must be a contiguous [num_tokens, hidden_size] tensor")
-        if input.device.type != "cuda" or input.dtype != torch.bfloat16:
-            raise ValueError("latency dispatch input must be a CUDA BF16 tensor")
+        expected_input_dtype = dispatch_tensor_dtype(mode_context.dispatch_data_type)
```
Binyang2014, line 504: **Agent-generated review comment (GitHub Copilot).**

dispatch_data_type describes the wire/output storage format, not the source tensor format. The FP8 sender still casts inputTokens to mscclpp::bf16x8 and quantizes it in dispatchSendFp8. This check therefore rejects the existing valid BF16+FP8 configuration, while an accepted FP8 tensor would be interpreted as BF16 and read past each row. Please keep the dispatch input BF16 until a distinct FP16/FP8 source path is implemented.

src/ext/ep/dispatch/common.cuh
```diff
@@ -153,6 +180,158 @@ MSCCLPP_DEVICE_INLINE void dispatchSendRankMajorBf16(void* output, int* outputTo
...
+        laneId < nTopk ? (topkWeights == nullptr ? 1.0f : topkWeights[tokenIdx * nTopk + laneId]) : 0.0f;
+    const size_t destIndex = (static_cast<size_t>(transport.rank_) * maxTokensPerRank + tokenIdx) * nTopk + laneId;
+
+    if (laneId == 0) bulkBarrier->wait(sendBulkPhase);
+    __syncwarp();
+    mscclpp::bulkFence();
+
+    if (dstRank >= 0) {
+      issueTokenMajorTokenStore<Hidden>(output, transport, dstRank, destIndex, stagedToken);
+      sendTokenMajorMetadata(transport, outputTopkIdx, outputTopkWeights, dstRank, destIndex, expert, weight);
```
Binyang2014, line 284: **Agent-generated review comment (GitHub Copilot).**

Could we invalidate metadata slots that are not sent to this rank? destIndex is a fixed sparse (source, token, top-k) slot, but only the selected destination receives an ID/weight write. Other ranks therefore expose zero-initialized values on the first epoch (expert 0 is valid on rank 0) or previous-epoch values through dispatch_out.topk_ids and weights. The in-tree combine reconstructs validity from the source handle and avoids these slots, but an actual MLP consuming the exposed metadata can process stale routes. Please write invalidTokenExpertId/zero for non-local and padding slots, or clear the full metadata region with correct ordering before publication.

src/ext/ep/include/config.hpp
```diff
@@ -184,25 +193,34 @@ struct LatencyStorageLayout {
...
-    const size_t dispatchRecvBufferBytes =
-        std::max({dispatchBufferBytes, rankMajorDispatchBufferBytes, dispatchOutputBytes_});
-    const size_t combineRecvBufferBytes = rankMajorDirectSend    ? rankMajorDirectSendCombineInputBytes
-                                          : rankMajorLocalReduce ? 0
-                                                                 : dispatchOutputBytes_;
+    const size_t tokenMajorDispatchBufferBytes = tokenMajorTokenOffsetBytes + tokenMajorDispatchOutputBytes;
+    dispatchOutputBytes_ = rankMajor    ? rankMajorDispatchOutputBytes
+                           : tokenMajor ? tokenMajorDispatchOutputBytes
+                                        : expertMajorDispatchOutputBytes;
+    const size_t dispatchRecvBufferBytes = std::max(
```
Binyang2014, line 204: **Agent-generated review comment (GitHub Copilot).**

Please include 	okenMajorDispatchBufferBytes only when TOKEN_MAJOR is selected. Putting it unconditionally in this maximum makes existing rank-major configurations reserve 
umTopk copies of their prior token capacity; with top-k 8 this is approximately an 8x symmetric-buffer regression and can make initialization OOM. Expert-major configurations can also grow whenever this term exceeds their existing output allocation.

src/ext/ep/combine/common.cuh
```diff
@@ -661,12 +847,14 @@ inline void combineHiddenMode(void* output, const void* expertOutput, const int6
   auto combineFunc = KernelSelector::template get<Hidden, DispatchType, ScaleBlockSize, Layout>();
   const size_t sharedBytes = combineSharedBytes<Hidden, Mode, Layout>(nLocalExperts, workload.numTopk_);
   const bool useRankMajorTma = Layout == DispatchLayout::RANK_MAJOR && workload.numTopk_ <= RankMajorTmaMaxNTopk;
-  EP_HOST_ASSERT(!useRankMajorTma || numBlocks > 1);
+  const bool useTokenMajorTma = Layout == DispatchLayout::TOKEN_MAJOR && workload.numTopk_ <= RankMajorTmaMaxNTopk;
+  const int launchBlocks = numBlocks + ((useRankMajorTma || useTokenMajorTma) ? 1 : 0);
```
Binyang2014, line 851: **Agent-generated review comment (GitHub Copilot).**


um_blocks is documented as the total combine grid size, and the existing rank-major TMA path previously launched exactly that many blocks. Adding a control block here silently changes that contract for rank-major as well as token-major. It also makes a previously valid occupancy-limit configuration (
esidentBlocks == numBlocks) fail the new assertion. Please keep block 0 within 
umBlocks, or update the public contract, defaults, and validation consistently.

src/ext/ep/common/latency.cuh
```diff
@@ -82,7 +88,8 @@ template <DispatchDataType DataType>
 using DispatchPayloadView = PayloadView<DispatchElementType<DataType>, DispatchScaleType<DataType>>;
 
 MSCCLPP_HOST_DEVICE_INLINE constexpr bool isSupportedDispatchDataType(DispatchDataType dataType) {
-  return dataType == DispatchDataType::BF16 || dataType == DispatchDataType::FP8_E4M3;
+  return dataType == DispatchDataType::BF16 || dataType == DispatchDataType::FP16 ||
```
Binyang2014, line 91: **Agent-generated review comment (GitHub Copilot).**

FP16 is now reported as supported and exposed through the public/Python enums, but the latency path is not wired for it. Python allocates no scale buffer for FP16, so dispatchAlgorithm first fails its non-BF16 outputScales assertion; even with a buffer, the dispatch and combine switches have no FP16 cases. Please remove FP16 exposure from this PR or implement the complete send, receive, and combine path before advertising it.

src/ext/ep/dispatch/common.cuh
```diff
@@ -153,6 +180,158 @@ MSCCLPP_DEVICE_INLINE void dispatchSendRankMajorBf16(void* output, int* outputTo
...
+    if (laneId == 0) {
+      bulkBarrier->arriveAndExpect(static_cast<uint32_t>(HiddenBytes));
+      mscclpp::bulkLoad(stagedToken, inputData, static_cast<uint32_t>(HiddenBytes), *bulkBarrier);
+    }
+
+    const int expert = laneId < nTopk ? static_cast<int>(topkIndices[tokenIdx * nTopk + laneId]) : -1;
+    const int dstRank = expert >= 0 ? expert / nLocalExperts : -1;
+    const float weight =
+        laneId < nTopk ? (topkWeights == nullptr ? 1.0f : topkWeights[tokenIdx * nTopk + laneId]) : 0.0f;
+    const size_t destIndex = (static_cast<size_t>(transport.rank_) * maxTokensPerRank + tokenIdx) * nTopk + laneId;
```
Binyang2014, line 307: **Agent-generated comment (GitHub Copilot), added at the reviewer's request.**

Please rename this layout to distinguish it from the existing `TOKEN_MAJOR` layout. The index here describes `[source_rank, original_token_idx, topk_slot, hidden]`: the outermost partition is still source rank, so this is a fixed-slot, top-k-expanded rank-major variant. Compared with the existing latency `RANK_MAJOR`, it preserves original token positions instead of compacting received tokens and stores one copy per expert route instead of one per destination rank.

Throughput already uses `TOKEN_MAJOR` for compact rows deduplicated per destination rank, so reusing that name hides a materially different storage and MLP contract. A distinct name such as `RANK_MAJOR_TOPK_EXPANDED`, with matching API/kernel names and layout documentation, would make that difference explicit. This rename should apply to the new latency layout, not the existing throughput layout.

src/ext/ep/dispatch/common.cuh
```diff
@@ -153,6 +180,158 @@ MSCCLPP_DEVICE_INLINE void dispatchSendRankMajorBf16(void* output, int* outputTo
...
+                                                  float* outputTopkWeights, int destinationRank, size_t destIndex,
+                                                  int expert, float weight) {
+  auto* destinationTopkIdx = reinterpret_cast<int*>(transport.mappedBuffer(outputTopkIdx, destinationRank));
+  auto* destinationTopkWeights = reinterpret_cast<float*>(transport.mappedBuffer(outputTopkWeights, destinationRank));
+  destinationTopkIdx[destIndex] = expert;
+  destinationTopkWeights[destIndex] = weight;
+}
+
+template <int Hidden>
+MSCCLPP_DEVICE_INLINE void dispatchSendTokenMajorBf16(void* output, int* outputTopkIdx, float* outputTopkWeights,
```
Binyang2014: **Agent-generated comment (GitHub Copilot).**

I'd reuse the existing rank-major dispatch framework here: `dispatchSendRankMajorBf16` already has the same warp assignment, shared-memory setup, token loop, and TMA load/barrier sequence, and the new store/count/notify helpers also closely duplicate their rank-major counterparts. Please factor out the shared mechanics and use compile-time layout differences for destination-rank deduplication, row indexing, and metadata handling, so synchronization fixes don't need to be maintained in two copies. Keep sparse-slot invalidation and the per-token versus aggregated completion-publication ordering explicit; those are not interchangeable with the existing path.

python/mscclpp/ep/latency.py
mahdiehghazim: Done

src/ext/ep/dispatch/common.cuh
mahdiehghazim: Done

src/ext/ep/include/config.hpp
mahdiehghazim: Done

src/ext/ep/combine/common.cuh
mahdiehghazim: Done

src/ext/ep/common/latency.cuh
mahdiehghazim: Done

src/ext/ep/dispatch/common.cuh
mahdiehghazim: Done

src/ext/ep/dispatch/common.cuh
mahdiehghazim: Done

PR #900, opened 2026-09-04
Expose flat token-major latency metadata

mahdiehghazim:
Expose TOKEN_MAJOR top-k metadata as flat row-aligned tensors and make token-major dispatch publish deterministic invalid metadata for non-local lanes. This lets Doug KI token-major compute consume dispatch_out.topk_ids and dispatch_out.weights directly by token row.

### DIFF
--- a/python/mscclpp/ep/README.md
+++ b/python/mscclpp/ep/README.md
@@ -591,6 +591,14 @@ dispatch_out.topk_ids   # [world_size * max_tokens_per_rank, K]
 dispatch_out.weights    # [world_size * max_tokens_per_rank, K]
 ```
 
+Latency `DispatchLayout.TOKEN_MAJOR` exposes one row per routed token/top-k lane:
+
+```python
+dispatch_out.tokens     # [world_size * max_tokens_per_rank * K, H]
+dispatch_out.topk_ids   # [world_size * max_tokens_per_rank * K]
+dispatch_out.weights    # [world_size * max_tokens_per_rank * K]
+```
+
 Tokens, top-k IDs, and weights are sent directly into runtime-owned registered
 final receive buffers and exposed as zero-copy Torch tensors.
 Unused rows in every source-rank block use `invalid_token_expert_id` and zero
--- a/python/mscclpp/ep/README.md
+++ b/python/mscclpp/ep/README.md
@@ -598,12 +598,14 @@ Latency `DispatchLayout.RANK_MAJOR_TOPK_EXPANDED` exposes one row per routed tok
 
 ```python
 dispatch_out.tokens     # [world_size * max_tokens_per_rank * K, H]
-dispatch_out.topk_ids   # [world_size * max_tokens_per_rank * K]
-dispatch_out.weights    # [world_size * max_tokens_per_rank * K]
+dispatch_out.topk_ids   # [world_size * max_tokens_per_rank, K]
+dispatch_out.weights    # [world_size * max_tokens_per_rank, K]
 ```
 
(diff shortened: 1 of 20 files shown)


python/mscclpp/ep/README.md
```diff
@@ -594,6 +594,14 @@ dispatch_out.topk_ids   # [world_size * max_tokens_per_rank, K]
 dispatch_out.weights    # [world_size * max_tokens_per_rank, K]
 ```
 
+Latency `DispatchLayout.RANK_MAJOR_TOPK_EXPANDED` exposes one row per routed token/top-k lane:
+
+```python
+dispatch_out.tokens     # [world_size * max_tokens_per_rank * K, H]
+dispatch_out.topk_ids   # [world_size * max_tokens_per_rank * K]
```
Binyang2014: change `dispatch_out.topk_ids` and `dispatch_out.weights` to `[world_size * max_tokens_per_rank, K]`?

python/mscclpp/ep/latency.py
```diff
@@ -387,7 +387,11 @@ def _bind_buffers(self) -> None:
         if context.output_layout not in (DispatchLayout.RANK_MAJOR, DispatchLayout.RANK_MAJOR_TOPK_EXPANDED):
             return
 
-        metadata_shape = (context.world_size * context.max_tokens_per_rank, context.topk)
+        metadata_shape = (
+            (context.world_size * context.max_tokens_per_rank * context.topk,)
+            if context.output_layout == DispatchLayout.RANK_MAJOR_TOPK_EXPANDED
+            else (context.world_size * context.max_tokens_per_rank, context.topk)
+        )
```
Binyang2014: `(context.world_size * context.max_tokens_per_rank * context.topk,)` and `(context.world_size * context.max_tokens_per_rank, context.topk)` are almost same. Can we use `(context.world_size * context.max_tokens_per_rank, context.topk)` for both format?

src/ext/ep/dispatch/common.cuh
```diff
@@ -256,10 +256,15 @@ MSCCLPP_DEVICE_INLINE void dispatchRankMajorTopkExpandedNotify(const TransportVi
   WorkspaceView workspaceView(workspace, nRanks, nExperts);
   auto* rankTokenCounts = sharedMem;
   countRankMajorTopkExpandedRoutes(rankTokenCounts, topkIndices, nTokens, nTopk, nRanks, nExperts);
+  auto* metadataCompletionCounts = sharedMem + nRanks;
+  for (int dstRank = static_cast<int>(threadIdx.x); dstRank < nRanks; dstRank += static_cast<int>(blockDim.x)) {
+    metadataCompletionCounts[dstRank] = nTokens * nTopk;
+  }
```
Binyang2014, line 263: Maybe we can move these lines before countRankMajorTopkExpandedRoutes and remove the __syncthreads() at line 263?

Binyang2014: Confused about this part, the dstRank no needs `nToken x nTopK` data right? As topK will send to different dst rank

mahdiehghazim: For this layout the token payload is sparse per dstRank, but the metadata is dense.

Each destination rank gets metadata for every [source token, top-k lane] slot. If that lane does not route to the destination rank, we still write invalid_token_expert_id and zero weight for that slot. So the metadata completion count is nTokens * nTopk for every dstRank, while the actual payload count remains rankTokenCounts[dstRank].

I added a comment about this in the code to clarify.

PR #901, opened 2026-09-04
Reuse CUDA Fabric mappings to prevent resource exhaustion

caiocbr:


src/core/gpu_ipc_mem.cc
```diff
@@ -62,6 +63,85 @@ struct PosixFdEntry {
...
+  }
+
+  void* data(size_t offset) const { return reinterpret_cast<void*>(base_ + offset); }
+
+ private:
+  std::shared_ptr<GpuIpcMem> owner_;
+  CUdeviceptr base_ = 0;
+  size_t size_ = 0;
+  bool reserved_ = false;
+  bool mapped_ = false;
```
Binyang2014: Do we need two flags? Or one flag is enough

src/core/gpu_ipc_mem.cc
```diff
@@ -465,44 +545,82 @@ std::shared_ptr<void> GpuIpcMem::map() {
...
 
-  CUmemAccessDesc accessDesc = {};
-  accessDesc.location.type = CU_MEM_LOCATION_TYPE_DEVICE;
-  accessDesc.location.id = deviceId;
-  accessDesc.flags = CU_MEM_ACCESS_FLAGS_PROT_READWRITE;
-  MSCCLPP_CUTHROW(cuMemSetAccess(base, handle_.baseSize, &accessDesc, 1));
+#if !defined(MSCCLPP_DEVICE_HIP)
+  if (type_ == GpuIpcMemHandle::Type::Fabric) {
+    CUcontext context;
+    MSCCLPP_CUTHROW(cuCtxGetCurrent(&context));
```
Binyang2014: Why need context here?

src/ext/collectives/allreduce/allreduce_rsag_zero_copy.cu
```diff
@@ -139,8 +139,12 @@ struct AllreduceRsAgZeroCopyAdapter {
       allreduceRsAgZeroCopy<16, OpType, T, AccumT>
           <<<nBlocks, nThreadsPerBlock, 0, stream>>>((T*)input, (T*)scratch, (T*)output, (ChannelType*)memoryChannels,
                                                      switchChannel, remoteMemories, rank, worldSize, nelems);
+    } else if (nRanksPerIpcDomain == 32) {
```
Binyang2014, line 142: **Generated by GitHub Copilot:** The new 32-rank path can deadlock for small inputs. When `nInt4PerRank < gridDim.x`, zero-work blocks return at line 70, but the remaining block calls `globalSyncer.sync(gridDim.x)`, which waits for every launched block. With the automatic 64-block launch and 32 ranks, this affects positive payloads up to 32,256 bytes. Please cap `nBlocks` to the available per-rank `int4` work before launch (and handle zero-length input) so every launched block reaches the barrier and block 0 remains active for the final `signal()`/`wait()`. Changing only the sync count is insufficient because that final inter-rank synchronization is executed by block 0.

src/core/gpu_ipc_mem.cc
```diff
@@ -465,44 +545,82 @@ std::shared_ptr<void> GpuIpcMem::map() {
...
+      }
+      fabricMappingCache.erase(it);
+    }
+    for (auto cacheIt = fabricMappingCache.begin(); cacheIt != fabricMappingCache.end();) {
+      if (cacheIt->second.expired()) {
+        cacheIt = fabricMappingCache.erase(cacheIt);
+      } else {
+        ++cacheIt;
+      }
+    }
```
Binyang2014: `lock()` already returns an empty `shared_ptr` for an expired entry, and the code above erases that key, so this loop is not needed for correctness. I would remove the O(cache size) scan unless eagerly bounding stale metadata for unrelated keys is intentional; lazy per-key cleanup should be sufficient.

src/ext/collectives/allreduce/allreduce_rsag_zero_copy.cu
caiocbr: We usually handle this by setting the message range, which is what I did in the sglang integration. Do you want me to add a check in the algorithm, or should I keep the limitation in message size? This algorithm works better for message sizes above 1M; below that, allreduce pkt is better, so I think this should be fine.

Binyang2014: I think better to cap nBlock in algo side.

Binyang2014 (approved):
LGTM, pls address the nBlock issue first

Binyang2014 (approved):

Binyang2014 (approved):

caiocbr:

PR #903, opened 2026-09-10
[MSCCLPP-EP] C++ throughput expert-parallel runtime

Binyang2014:
## Summary

Adds throughput-mode expert-parallel dispatch/combine to the C++ `MoERuntime`, alongside the existing latency path. This PR targets `binyli/ep-latency-cpp`.

## Implementation

- **GPU-only preparation:** `prepare(PrepareRequest)` returns an opaque, reusable `PrepareHandle`. Counts, stable receive offsets, and per-token destination maps stay on the GPU. Dispatch prepares automatically when `prepareHandle` is empty, or reuses an explicit preparation while routing remains unchanged.
- **Unified token-based dispatch:** the existing count/scan kernel builds rank-ordered `{rank, offset}` entries, deduplicating destinations without an extra kernel. Dispatch uses this map for every supported rank count, loads each input vector once, and fans it out to the selected peers. There is no `RouteByTopk` specialization or lane-ID-equals-rank-ID requirement.
- **Pipelined combine:** reuses the same routing map to gather BF16 expert results through a two-stage TMA pipeline, accumulate in FP32, and write BF16 output using MSCCL++ vector types.
- **Configuration-sized storage:** one registered symmetric allocation per rank plus a private workspace. Dispatch output and combine input share the receive buffer. `TOKEN_MAJOR` rows are dense, without gaps between source-rank batches; `RANK_MAJOR` retains explicit source-rank strides. Dispatch supports BF16 and FP8 E4M3, with optional expert-ID, weight, and scale outputs.
- **Shared launch and synchronization helpers:** reuse IPC mappings and channel semaphores, with cooperative grid synchronization and common shared-memory/occupancy setup for latency and throughput.

## API contract and scope

- CUDA SM90 or newer; 1-64 ranks within one CUDA IPC domain; top-k in `[1, 8]`. This does not add a cross-IPC-domain/RDMA transport.
- Operations enqueue asynchronously on one caller-provided stream, including preparation, dispatch, expert computation, and combine. Cross-stream overlap is not supported, and handles are not completion events. The caller must serialize host calls and complete local/peer GPU use before destroying the runtime.
- Token input/output buffers must be 16-byte aligned. Active per-rank capacity may be smaller than the configured allocation capacity.

## Validation and performance

The existing `ThroughputCorrectness` and `ThroughputPerformance` tests passed on eight H100 GPUs. Correctness includes both layouts and dispatch formats, automatic/explicit preparation, reuse and routing rebuilds, scan boundaries, duplicate/invalid routes, and non-power-of-two experts per rank. Preparation is covered by correctness rather than a separate test.

Performance measures **eager dispatch + combine**, using one initialized runtime across 1K/2K/4K tokens per rank, hidden size 7168, 128 experts, top-k 8, D/C block budgets of 24/32 and 130/128, BF16/FP8 dispatch, and automatic/cached preparation. Allocation, input conversion, and expert compute are outside timing; normal combine input staging is included.

The latest route-map simplification (`3d7964d9` versus `6c8d9cc3`, both within this PR) has the following measured D+C latency changes across the 24 profiles, comparing three baseline runs with two updated runs. Positive values mean slower:

| Preparation | Per-profile change range | Median change |
|---|---:|---:|
| Automatic | +0.38% to +1.31% | +0.69% |
| Cached | -0.38% to +0.82% | -0.06% |

Physical 32/64-GPU qualification has not been performed.


src/ext/ep/include/config.hpp
```diff
@@ -30,6 +32,19 @@ MSCCLPP_HOST_DEVICE_INLINE constexpr dtype_t configAlign(dtype_t a, dtype_t b) {
 using Bf16 = typename mscclpp::bf16x2::ElementType;
 using Fp8E4M3 = typename mscclpp::f8_e4m3x2::ElementType;
 
+MSCCLPP_HOST_DEVICE_INLINE constexpr int dispatchElementBytes(DispatchDataType dispatchDataType) {
+  return dispatchDataType == DispatchDataType::BF16 ? static_cast<int>(sizeof(Bf16))
+                                                    : static_cast<int>(sizeof(Fp8E4M3));
+}
+
+MSCCLPP_HOST_DEVICE_INLINE constexpr int dispatchScaleBlockSize(DispatchDataType dispatchDataType) {
```
Binyang2014: Change the name for this? nelemsPerScale?

src/ext/ep/include/config.hpp
```diff
@@ -213,6 +228,81 @@ inline size_t latencyStorageSize(int maxTokensPerRank, int hidden, int numRanks,
...
+  static constexpr int RecvPoolMaxTokens = 65536;
+  static constexpr int64_t RecvPoolMaxHiddenBytes = 16384;
+  static constexpr int64_t RecvPoolMetaBytes =
+      ((MaxTopk * (sizeof(int) + sizeof(float)) + MaxScales * sizeof(float) + BufferAlignmentBytes - 1) /
+       BufferAlignmentBytes) *
+      BufferAlignmentBytes;
+
+  // Control and receive-pool storage are separate peer-visible allocations.
+  size_t controlBufferBytes_;
+  size_t recvPoolBytes_;
```
Binyang2014: It's recv pool? Not just dipatch Output?

src/ext/ep/include/config.hpp
```diff
@@ -213,6 +228,81 @@ inline size_t latencyStorageSize(int maxTokensPerRank, int hidden, int numRanks,
...
+  size_t recvPoolHiddenBytes() const { return recvPoolMetadataOffset_ - recvPoolHeaderBytes_; }
+};
+
+struct ThroughputWorkspaceLayout {
+  size_t totalBytes_;
+  int* numTokensPerRank_ = nullptr;
+  int* numTokensPerExpert_ = nullptr;
+  bool* isTokenInRank_ = nullptr;
+  int* rankPrefixMatrix_ = nullptr;
+  int* channelPrefixMatrix_ = nullptr;
```
Binyang2014: Why need this?

src/ext/ep/include/config.hpp
```diff
@@ -213,6 +228,81 @@ inline size_t latencyStorageSize(int maxTokensPerRank, int hidden, int numRanks,
...
+  }
+
+  size_t recvPoolHiddenBytes() const { return recvPoolMetadataOffset_ - recvPoolHeaderBytes_; }
+};
+
+struct ThroughputWorkspaceLayout {
+  size_t totalBytes_;
+  int* numTokensPerRank_ = nullptr;
+  int* numTokensPerExpert_ = nullptr;
+  bool* isTokenInRank_ = nullptr;
```
Binyang2014: It's a bitmap?

src/ext/ep/include/device_context.hpp
```diff
@@ -14,10 +14,18 @@ struct DeviceContext {
   void* localBufferBase_;
   /// Peer-mapped control or symmetric-buffer bases.
   void* const* peerBufferBases_;
+  /// Optional peer-mapped payload-pool bases.
+  void* const* peerPayloadBases_;
```
Binyang2014: Different with peerBufferBase?

src/ext/ep/include/device_context.hpp
```diff
@@ -14,10 +14,18 @@ struct DeviceContext {
...
   /// Peer synchronization channels.
   mscclpp::BaseMemoryChannelDeviceHandle* channels_;
   /// Optional algorithm workspace.
   void* workspace_;
+  /// Optional token/rank receive indices used by throughput combine.
+  int* combineRecvIdx_;
+  /// Optional mapped receive count used by throughput preparation.
+  int* mappedRecvCounter_;
+  /// Optional mapped per-expert receive counts used by throughput preparation.
+  int* mappedRecvExpertCounters_;
```
Binyang2014: Can this hide in workspace?

src/ext/ep/include/moe_runtime_context.hpp
```diff
@@ -80,8 +97,55 @@ struct LatencyRuntimeContext {
...
+  DispatchLayout outputLayout_;
+  size_t controlBufferBytes_ = 0;
+  size_t symmetricBufferBytes_ = 0;
+  size_t recvPoolBytes_ = 0;
+  size_t workspaceBytes_ = 0;
+  bool physicalControlBuffer_ = false;
+  mscclpp::Communicator& communicator_;
+  uint32_t epoch_ = 0;
+  void* symmetricBuffer_ = nullptr;
+  void* recvPool_ = nullptr;
```
Binyang2014: Why have bother symmetricBuffer and recvPool

src/ext/ep/include/config.hpp
```diff
@@ -213,6 +228,81 @@ inline size_t latencyStorageSize(int maxTokensPerRank, int hidden, int numRanks,
...
+      ((MaxTopk * (sizeof(int) + sizeof(float)) + MaxScales * sizeof(float) + BufferAlignmentBytes - 1) /
+       BufferAlignmentBytes) *
+      BufferAlignmentBytes;
+
+  // Control and receive-pool storage are separate peer-visible allocations.
+  size_t controlBufferBytes_;
+  size_t recvPoolBytes_;
+  size_t recvPoolHeaderBytes_;
+  size_t recvPoolMetadataOffset_;
+  void* dispatchOutputBuffer_ = nullptr;
```
Binyang2014: Can we use something similar to laytency payload for this? Or it's different?

src/ext/ep/latency.cc
```diff
@@ -242,12 +253,17 @@ void MoERuntime::launchLatencyCombine(const LatencyCombineRequest& request) {
     EP_THROW("Stale dispatch handle: a newer dispatch has replaced its metadata");
   }
 
+  const auto* latencyMetadata = std::get_if<DispatchHandle::Impl::LatencyMetadata>(&handle.metadata_);
```
Binyang2014: Just rename metadata?

src/ext/ep/latency.cc
```diff
@@ -149,13 +149,24 @@ void* MoERuntime::outputTopkWeightsBuffer() const {
...
 
 void* MoERuntime::combineInputBuffer() const {
-  requireMode(MoEMode::LATENCY);
-  const auto& context = *latencyContext_;
-  EP_HOST_ASSERT(context.outputLayout_ == DispatchLayout::RANK_MAJOR);
-  EP_HOST_ASSERT(context.symmetricBuffer_ != nullptr);
-  return LatencyStorageLayout(context.symmetricBuffer_, context.maxTokensPerRank_, context.hidden_, context.numRanks_,
-                              context.numExperts_, context.numTopk_, context.outputLayout_, context.combineMode_)
-      .combineRecvBuffer_;
+  switch (mode_) {
```
Binyang2014: Move this function to moe_runtime.cc?

src/ext/ep/include/kernels.hpp
```diff
@@ -9,11 +9,21 @@
...
 
+#include "config.hpp"
 #include "device_context.hpp"
 
 namespace mscclpp {
 namespace ep {
 
+inline constexpr int ThroughputCountThreads = 128;
+inline constexpr int DispatchNWarps = 16;
+inline constexpr int DispatchMinNWarpsPerGroup = 8;
```
Binyang2014: If it just for throughput, pls add thoguhput prefix

src/ext/ep/include/kernels.hpp
```diff
@@ -45,6 +55,55 @@ struct Workload {
...
+                           const DeviceContext& context, cudaStream_t stream);
+
+// Collective preparation: exchange counts and determine source-rank receive ranges.
+void throughputExchangeCounts(const ThroughputWorkspaceLayout& workspace, const Workload& workload,
+                              const DeviceContext& context, cudaStream_t stream);
+
+// Wait until peers have finished consuming the previous payload before overwriting it.
+void throughputSynchronizePeers(const DeviceContext& context, cudaStream_t stream);
+
+int maxCooperativeThroughputDispatchBlocks(DispatchLayout layout, const DeviceContext& context);
```
Binyang2014, line 90: Why do we need this? Can use adjust sm?

src/ext/ep/include/moe_runtime_context.hpp
```diff
@@ -20,27 +21,57 @@
 namespace mscclpp {
 namespace ep {
 
+struct Workload;
+
+struct PrepareHandle::Impl {
+  Impl(std::weak_ptr<ThroughputRuntimeContext> owner, uint64_t epoch, const PrepareRequest& request)
+      : owner_(std::move(owner)),
+        epoch_(epoch),
```
Binyang2014: Why need epoch?

src/ext/ep/include/moe_runtime_context.hpp
```diff
@@ -80,8 +111,45 @@ struct LatencyRuntimeContext {
...
+  size_t workspaceBytes_ = 0;
+  mscclpp::Communicator& communicator_;
+  uint32_t epoch_ = 0;
+  uint64_t prepareEpoch_ = 0;
+  cudaEvent_t prepareEvent_ = nullptr;
+  void* symmetricBuffer_ = nullptr;
+  void* workspace_ = nullptr;
+  std::vector<void*> peerMappedBufferBases_;
+  std::vector<mscclpp::RegisteredMemory> peerBufferMemories_;
+  void** peerMappedBufferBasesGpu_ = nullptr;
```
Binyang2014, line 150: What's this? Does latency mode has similar structure or use other object for this?

src/ext/ep/dispatch/throughput_dispatch.cu
```diff
@@ -0,0 +1,290 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+//
+// Portions adapted from DeepEP (https://github.com/deepseek-ai/DeepEP),
```
Binyang2014: Still similar as deepEP?

src/ext/ep/dispatch/throughput_dispatch.cu
```diff
@@ -0,0 +1,290 @@
...
+  __syncthreads();
+
+  const int numExpertsPerRank = workload.numExperts_ / numRanks;
+  if (threadId < numRanks) {
+    auto* peerRankCounts = reinterpret_cast<int*>(context->peerBufferBases_[threadId]);
+    auto* peerExpertCounts = peerRankCounts + numRanks * numRanks;
+    for (int dstRank = 0; dstRank < numRanks; ++dstRank) {
+      peerRankCounts[context->rank_ * numRanks + dstRank] = workspace.numTokensPerRank_[dstRank];
+    }
+#pragma unroll
```
Binyang2014: Does this work?

src/ext/ep/dispatch/throughput_dispatch.cu
```diff
@@ -0,0 +1,290 @@
...
+      float secondScale = 0.0f;
+      if (cacheScales) {
+        if (laneId < numScales) firstScale = __ldg(inputScales + static_cast<int64_t>(token) * numScales + laneId);
+        if (laneId + WARP_SIZE < numScales)
+          secondScale = __ldg(inputScales + static_cast<int64_t>(token) * numScales + laneId + WARP_SIZE);
+      }
+      unsigned destinations = destinationMask;
+      while (destinations != 0) {
+        const int dstRank = __ffs(static_cast<int>(destinations)) - 1;
+        const int outputIndex = __shfl_sync(0xffffffffu, laneSlot, dstRank);
```
Binyang2014: We have some utils for these in device_hlpders? Do we still need this? Or you can warp this to a device function?

src/ext/ep/dispatch/throughput_dispatch.cu
```diff
@@ -0,0 +1,290 @@
...
+          payload.topKValues(dstBuffer, outputIndex)[laneId] = localExpert >= 0 ? weight : 0.0f;
+        }
+        if constexpr (DataType == DispatchDataType::FP8_E4M3) {
+          auto* metadataScales = payload.scaleFactors(dstBuffer, outputIndex);
+          if (cacheScales) {
+            if (laneId < numScales) metadataScales[laneId] = firstScale;
+            if (laneId + WARP_SIZE < numScales) metadataScales[laneId + WARP_SIZE] = secondScale;
+          } else {
+            for (int scale = laneId; scale < numScales; scale += WARP_SIZE) {
+              metadataScales[scale] = __ldg(inputScales + static_cast<int64_t>(token) * numScales + scale);
```
Binyang2014: Do we really need __ldg?

src/ext/ep/combine/throughput_reduce_combine.cu
```diff
@@ -0,0 +1,253 @@
...
+        launch = useWideKernel ? launchThroughputCombine<8, COMBINE_TMA_WARPS_WIDE>
+                               : launchThroughputCombine<8, COMBINE_TMA_WARPS_NARROW>;
+      } else if (numTopk <= 12) {
+        launch = launchThroughputCombine<12, 9>;
+      } else {
+        launch = launchThroughputCombine<16, 7>;
+      }
+      break;
+    default:
+      EP_HOST_ASSERT(false && "Unsupported ranks");
```
Binyang2014: We need to support 32 ranks? Can we do now?

Binyang2014: For topK, just support up to 8 for now

src/ext/ep/include/config.hpp
Binyang2014: Renamed this to `dispatchElementsPerScale()`. The result is an element count (128 for FP8), not a byte size; BF16 does not use scales. The callers now use the same name. This is included in [75f248b1](https://github.com/microsoft/mscclpp/commit/75f248b1f5a93f710bef9c24f3fbd5243b3702f8).

src/ext/ep/include/config.hpp
Binyang2014: Agreed that this is not dispatch-only storage. `recvBuffer_` is the shared payload region used as dispatch output and then as combine input, so both public buffer accessors return it. There is no separate global receive pool now: `ThroughputStorageLayout` derives the fixed allocation from the configured capacity, ranks, hidden size, experts, and top-k. Included in [75f248b1](https://github.com/microsoft/mscclpp/commit/75f248b1f5a93f710bef9c24f3fbd5243b3702f8).

src/ext/ep/include/config.hpp
Binyang2014: The channel-prefix matrix was needed by the old channel-partitioned sender, but it is no longer needed by token-centric scheduling. Removed it in [75f248b1](https://github.com/microsoft/mscclpp/commit/75f248b1f5a93f710bef9c24f3fbd5243b3702f8). A destination slot is now derived from the local `rankOffsets_` base plus the stable `recvTokenOffsets_` entry; dispatch and combine use the same `recvTokenIndex()` helper.

src/ext/ep/include/config.hpp
Binyang2014: The original field was an unpacked bool matrix, not a packed bitmap. It has since been removed entirely in [75f248b1](https://github.com/microsoft/mscclpp/commit/75f248b1f5a93f710bef9c24f3fbd5243b3702f8): each token/rank pair already has a stable receive offset, with `-1` meaning not routed. That carries membership without a second mask allocation.

src/ext/ep/include/device_context.hpp
Binyang2014: Removed the separate peer-payload pointer array. Both modes now use `TransportView` to translate a local address into the corresponding peer mapping of the single symmetric allocation. Control data and payload regions share that mapping; they do not need separate peer-base tables. Included in [75f248b1](https://github.com/microsoft/mscclpp/commit/75f248b1f5a93f710bef9c24f3fbd5243b3702f8).

src/ext/ep/include/device_context.hpp
Binyang2014: Moved the algorithm scratch into `ThroughputWorkspaceLayout`. Receive placement is represented by `rankOffsets_` and `recvTokenOffsets_`, alongside the GPU count arrays; the duplicate combine-index allocation and host-mapped counters are gone. `DeviceContext` now contains common device/transport resources only. Included in [75f248b1](https://github.com/microsoft/mscclpp/commit/75f248b1f5a93f710bef9c24f3fbd5243b3702f8).

src/ext/ep/include/moe_runtime_context.hpp
Binyang2014: Consolidated them into one configuration-sized physical symmetric allocation, registered once locally and shared with peers. `ThroughputStorageLayout` describes its control and payload regions; private routing scratch remains in the separate local workspace. There is no separately allocated `recvPool_` anymore. Included in [75f248b1](https://github.com/microsoft/mscclpp/commit/75f248b1f5a93f710bef9c24f3fbd5243b3702f8).

src/ext/ep/include/config.hpp
Binyang2014: Added `ThroughputPayloadView` with typed accessors, following the latency view pattern. The layout intentionally differs: throughput keeps dense token rows in one slab and fixed-stride metadata in another, so GEMM can consume the token rows directly. Latency keeps packed per-token payloads. Both modes reuse `TransportView`, `Workload`, and the launch-configuration cache. Included in [75f248b1](https://github.com/microsoft/mscclpp/commit/75f248b1f5a93f710bef9c24f3fbd5243b3702f8).

src/ext/ep/latency.cc
Binyang2014: Renamed the local pointer to `metadata` in [68dd209a](https://github.com/microsoft/mscclpp/commit/68dd209a76344749c8528e3bcb3bd149ad0eddd5). The function already establishes latency mode, so the extra local-name prefix was redundant. The variant type check, ownership check, and stale-handle check are unchanged.

src/ext/ep/latency.cc
Binyang2014: Moved `MoERuntime::combineInputBuffer()` to `moe_runtime.cc`, next to `dispatchOutputBuffer()`, in [68dd209a](https://github.com/microsoft/mscclpp/commit/68dd209a76344749c8528e3bcb3bd149ad0eddd5). It switches over both runtime modes, so it belongs in the shared runtime implementation. Buffer selection and validation are unchanged.

src/ext/ep/include/kernels.hpp
Binyang2014: These were actually latency scheduling constants, with `DispatchNWarps` also reused by throughput. In [68dd209a](https://github.com/microsoft/mscclpp/commit/68dd209a76344749c8528e3bcb3bd149ad0eddd5), the latency constants and its warp-group heuristic are back in `common/latency.cuh`; throughput now has file-local `ThroughputDispatchWarps`/`ThroughputDispatchThreads`. The existing geometry values are unchanged.

src/ext/ep/include/kernels.hpp
Binyang2014: This is a cooperative-grid residency check, rather than an SM-count tuning heuristic. `this_grid().sync()` requires the whole grid to be resident; the legal block count depends on registers/shared memory and blocks per SM, not only the SM count. The helper reuses `configureKernel()` and takes the lower BF16/FP8 limit because a preparation can be reused with either format. Keeping the preflight check also rejects an invalid request before preparation overwrites workspace or invalidates handles. Callers can reduce `numBlocks`, but the occupancy check is still needed. Added that explanation in [68dd209a](https://github.com/microsoft/mscclpp/commit/68dd209a76344749c8528e3bcb3bd149ad0eddd5).

src/ext/ep/include/moe_runtime_context.hpp
Binyang2014: A preparation handle borrows the runtime's routing workspace; it does not own a snapshot. After `prepare(A)` followed by `prepare(B)`, the pointers and owner are still the same, but A's routing has been overwritten. The generation lets dispatch reject that stale handle. It is separate from the dispatch epoch because several dispatches can reuse one preparation without invalidating it. This is a host-side lifetime check, not GPU synchronization or count readback. Added a field comment in [68dd209a](https://github.com/microsoft/mscclpp/commit/68dd209a76344749c8528e3bcb3bd149ad0eddd5).

src/ext/ep/include/moe_runtime_context.hpp
Binyang2014: Yes, latency has the same `peerMappedBufferBasesGpu_` member and initialization pattern. `peerMappedBufferBases_` is the host pointer table used during setup; the GPU copy is passed through `DeviceContext::peerBufferBases_` so kernels can resolve peer addresses. `peerBufferMemories_` retains the registrations/mappings. This allocates a device pointer table, not another payload buffer. Added a clarifying comment in [68dd209a](https://github.com/microsoft/mscclpp/commit/68dd209a76344749c8528e3bcb3bd149ad0eddd5).

src/ext/ep/dispatch/throughput_dispatch.cu
Binyang2014: The header records the original adaptation, not equivalence with current DeepEP. The sender has since been reworked into token-centric warp-group fanout using GPU-generated stable offsets and MSCCL++ transport/synchronization primitives. The routing-count/peer-count-exchange approach retains that lineage, so the attribution is kept even though the current implementation and public API have diverged.

src/ext/ep/dispatch/throughput_dispatch.cu
Binyang2014: `numExpertsPerRank` is runtime-sized now. The loop is valid, but an unqualified `#pragma unroll` is only a compiler hint and does not guarantee full unrolling of that runtime bound. Removed the misleading directive in [68dd209a](https://github.com/microsoft/mscclpp/commit/68dd209a76344749c8528e3bcb3bd149ad0eddd5); the normal loop and its surrounding publication barriers remain.

src/ext/ep/dispatch/throughput_dispatch.cu
Binyang2014: Reused the existing `warpBroadcast()` helper for the receive-slot value in [68dd209a](https://github.com/microsoft/mscclpp/commit/68dd209a76344749c8528e3bcb3bd149ad0eddd5), matching the peer-row pointer broadcast in the data path. No duplicate shuffle helper is needed. The simple `__ffs` set-bit iteration remains, and all participating lanes still reach the broadcasts uniformly, including partial-row copies.

src/ext/ep/dispatch/throughput_dispatch.cu
Binyang2014: No explicit `__ldg` is required for these scale reads. Changed both the cached-scale loads and the larger-scale fallback to ordinary typed loads in [68dd209a](https://github.com/microsoft/mscclpp/commit/68dd209a76344749c8528e3bcb3bd149ad0eddd5). Scale readiness comes from stream/graph ordering, not a read-only-cache intrinsic. The separate input-vector/top-k load paths are otherwise unchanged.

src/ext/ep/combine/throughput_reduce_combine.cu
Binyang2014: Implemented the 32-rank configuration path and capped top-k at 8 in [68dd209a](https://github.com/microsoft/mscclpp/commit/68dd209a76344749c8528e3bcb3bd149ad0eddd5). Runtime/storage/launch validation now agree on ranks 2/4/8/16/32 and top-k 1..8. Combine selects 2/4/8 contributor slots from `min(numTopk, numRanks)`, so 32 ranks do not require a 32-contributor shared-memory allocation. The token-centric sender already uses a complete warp for routing/metadata.

This still requires every rank to be in the same CUDA IPC domain; it does not add an RDMA path. Added 32-destination routing coverage (including rank 31 and scan-tile boundaries), storage coverage, and eight-GPU dispatch/combine cases for top-k 1/2/3/4/5/8 in both layouts. A physical 32-rank end-to-end run is not available on this eight-GPU machine, so I am leaving this thread open for that confirmation.

include/mscclpp/ext/ep/moe_runtime.hpp
```diff
@@ -14,26 +14,31 @@ namespace mscclpp {
...
-/// One runtime owns the communication buffers and synchronization state for the
-/// selected mode. This library initially implements LATENCY with fixed-capacity
-/// expert-major or rank-major layouts. The THROUGHPUT API is reserved for a
-/// follow-up implementation and is rejected at runtime.
-/// Operations are asynchronous with respect to the host and execute on the
-/// CUDA stream supplied by each request.
+/// Both modes own fixed-capacity symmetric storage sized from the runtime
+/// configuration. LATENCY exposes expert-major or rank-major rows. THROUGHPUT
+/// exposes compact token-major or fixed-stride rank-major rows within that storage.
+/// Preparation, dispatch, and combine execute asynchronously on the request's
```
Binyang2014: Pls just let say all operation should in same stream or gurantee the gpu execution order

include/mscclpp/ext/ep/moe_runtime.hpp
```diff
@@ -77,35 +82,69 @@ class MoERuntime {
   /// Return the runtime-owned combine input buffer.
   void* combineInputBuffer() const;
 
+  /// Collectively prepare throughput routing without moving token payloads.
+  ///
+  /// Computes stable local token offsets and exchanges peer counts to determine
+  /// receive ranges entirely on the GPU, without a host copy or wait.
+  /// Dispatch can consume the handle on another stream using a device-side
+  /// event dependency. Preparation is supported only in THROUGHPUT mode and
+  /// can be captured together with dispatch in a CUDA graph.
```
Binyang2014: Just say dispatch will consume, user need to make sure the handle ready

src/ext/ep/include/config.hpp
```diff
@@ -12,10 +12,15 @@
...
 
+#include "exception.hpp"
+
 namespace mscclpp {
 namespace ep {
 
 inline constexpr size_t BufferAlignmentBytes = 128;
+inline constexpr int MaxNumTopk = 8;
+
+inline constexpr bool isSupportedThroughputRanks(int numRanks) { return numRanks > 0 && numRanks <= 64; }
```
Binyang2014: Same for latency? So remove throughput from the name?

src/ext/ep/include/config.hpp
```diff
@@ -213,6 +236,128 @@ inline size_t latencyStorageSize(int maxTokensPerRank, int hidden, int numRanks,
   return configAlign<size_t>(numBytes, BufferAlignmentBytes);
 }
 
+// Unlike latency's packed per-token payload, throughput keeps dense token rows
+// and fixed-stride metadata in separate slabs so GEMM can use the rows directly.
+struct ThroughputPayloadView {
```
Binyang2014, line 258: Can u add some comments like this to show the layout
//
//   [data: DataType[hidden]]
//   [optional scales: ScaleType[hidden / format scale block size]]
//   [topKIndices: int[topK]]
//   [topKValues: float[topK]]
//   [srcTokenGlobalIdx: int]
//

src/ext/ep/include/kernels.hpp
```diff
@@ -45,6 +47,57 @@ struct Workload {
...
+    MSCCLPP_CUDATHROW(cudaFuncSetAttribute(kernel, cudaFuncAttributeMaxDynamicSharedMemorySize,
+                                           static_cast<int>(dynamicSharedBytes)));
+    int blocksPerSm;
+    MSCCLPP_CUDATHROW(
+        cudaOccupancyMaxActiveBlocksPerMultiprocessor(&blocksPerSm, kernel, nThreads, dynamicSharedBytes));
+    cache.deviceId_ = context.deviceId_;
+    cache.dynamicSharedBytes_ = dynamicSharedBytes;
+    cache.residentBlocks_ = blocksPerSm * context.numSms_;
+  }
+  return cache.residentBlocks_;
```
Binyang2014, line 74: Really need this? Or like latency mode just launch with some config

src/ext/ep/throughput.cc
```diff
@@ -0,0 +1,308 @@
...
+  if (!available_) return;
+  symmetricBufferBytes_ = storageLayout().totalBytes_;
+  workspaceBytes_ = throughputWorkspaceSize(maxTokensPerRank_, numRanks_, numExperts_);
+}
+
+ThroughputRuntimeContext::~ThroughputRuntimeContext() noexcept(false) {
+  if (deviceContext_.devicePtr_ == nullptr) return;
+
+  CudaDeviceGuard deviceGuard(deviceContext_.deviceId_);
+  MSCCLPP_CUDATHROW(cudaDeviceSynchronize());
```
Binyang2014: No need sync, just like latency mode, use guranntee this

src/ext/ep/throughput.cc
```diff
@@ -0,0 +1,308 @@
...
+  MSCCLPP_CUDATHROW(cudaStreamIsCapturing(request.stream, &captureStatus));
+  const ThroughputWorkspaceLayout workspaceLayout(context.workspace_, context.maxTokensPerRank_, context.numRanks_,
+                                                  context.numExperts_);
+  EP_HOST_ASSERT(workspaceLayout.totalBytes_ <= context.workspaceBytes_);
+  auto metadata = std::make_shared<PrepareHandle::Impl>(throughputContext_, context.prepareEpoch_ + 1, request);
+
+  // Preparation replaces shared routing metadata, but reusing a preparation only
+  // replaces dispatch results. Track those lifetimes independently.
+  ++context.prepareEpoch_;
+  ++context.epoch_;
```
Binyang2014: Why nee tow epoch here?

src/ext/ep/dispatch/throughput_dispatch.cu
Binyang2014: I think we can remove, the kernel already different

src/ext/ep/dispatch/throughput_dispatch.cu
```diff
@@ -0,0 +1,326 @@
...
+      const int expert = routedExpert >= 0 && routedExpert < workload.numExperts_ ? static_cast<int>(routedExpert) : -1;
+      laneRank = expert >= 0 ? (powerOfTwoExperts ? expert >> expertRankShift : expert / expertsPerRank) : -1;
+      const bool firstLaneForRank = isFirstLaneForRank(laneRank, laneId);
+      laneSlot = laneRank >= 0 && firstLaneForRank ? workspace.recvTokenIndex(token, laneRank, numRanks) : -1;
+      laneBuffer = laneSlot >= 0 ? transport.mappedBuffer(recvBuffer, laneRank) : nullptr;
+    } else {
+      laneSlot = laneId < numRanks ? workspace.recvTokenIndex(token, laneId, numRanks) : -1;
+    }
+    // For larger peer sets, mask bits identify top-k lanes rather than rank IDs.
+    const unsigned destinationMask = __ballot_sync(0xffffffffu, laneSlot >= 0);
```
Binyang2014: What's this function __ballot_sync? Can we move to device helper and give a good name for this?

src/ext/ep/combine/throughput_reduce_combine.cu
```diff
@@ -0,0 +1,233 @@
...
+  EP_DEVICE_ASSERT(input != nullptr || receivedTokens == 0);
+
+  auto* localTokens = payload.data<int4>(recvBuffer);
+  if (receivedTokens > 0 && input != localTokens) {
+    const int rows = rankMajor ? numRanks * maxTokensPerRank : receivedTokens;
+    copyThroughputRows(localTokens, input, rows, hiddenInt4 * sizeof(int4), workspace.recvCounts_, maxTokensPerRank,
+                       rankMajor, blockIdx.x * blockDim.x + threadIdx.x, gridDim.x * blockDim.x);
+    // Join staging stores before the system-release peer publication below.
+    cooperative_groups::this_grid().sync();
+  }
```
Binyang2014, line 60: What is doing here? Need a device sync?

src/ext/ep/combine/throughput_reduce_combine.cu
```diff
@@ -0,0 +1,234 @@
...
+  const int maxTokensPerRank = workload.maxTokensPerRank_;
+  const bool rankMajor = workload.outputLayout_ == DispatchLayout::RANK_MAJOR;
+  const TransportView transport(context);
+  const int numRanks = context->numRanks_;
+  // MaxContributors is min(numTopk, numRanks) rounded up to 2, 4, or 8 slots.
+  EP_DEVICE_ASSERT(MaxContributors >= (numTopk < numRanks ? numTopk : numRanks));
+  constexpr int ChunkInt4 = COMBINE_TMA_CHUNK_INT4;
+  constexpr int NumStages = COMBINE_TMA_STAGES;
+  constexpr int ChunkBytes = ChunkInt4 * static_cast<int>(sizeof(int4));
+  constexpr int Bf16PerInt4 = sizeof(int4) / sizeof(nv_bfloat16);
```
Binyang2014: Why not use BF16 dataytpe from mscclpp

test/mp_unit/ep_tests.cu
```diff
@@ -558,3 +1507,13 @@ PERF_TEST(MoERuntimeTest, GraphRankMajorDispatchCombine32Tokens) {
   runGraphPerformance(*communicator, gEnv->rank, dispatchBlocks_, combineBlocks_,
                       mscclpp::ep::DispatchLayout::RANK_MAJOR, "rank-major 32 tokens/rank graph D+weighted staging+C");
 }
+
+PERF_TEST(MoERuntimeTest, GraphThroughputDispatchOnly) {
```
Binyang2014: For throughput no graph, eager mode but memory register just once

test/mp_unit/ep_tests.cu
```diff
@@ -549,6 +1218,286 @@ TEST(MoERuntimeTest, DispatchCombineCorrectness) {
...
+      runThroughputCorrectnessCase(*communicator, gEnv->rank, 24, combineBlocks_, layout,
+                                   {.preparationMode = ThroughputPreparation::Captured,
+                                    .numTokens = tokens,
+                                    .hidden = 136,
+                                    .runtimeCapacity = tokens + 31});
+    }
+  }
+}
+
+TEST(MoERuntimeTest, ThroughputPreparationValidation) {
```
Binyang2014: Why need this? we already have correctness check

src/ext/ep/dispatch/throughput_dispatch.cu
Binyang2014: Removed the generic adaptation header from the rewritten dispatch implementation in [8bcb2b9e](https://github.com/microsoft/mscclpp/commit/8bcb2b9e01a8c79a6c1648aa95bacff93a80586c). The remaining provenance note is scoped to `exchangeThroughputCountsKernel`, whose count-publication/prefix structure is retained from the original adaptation. This no longer labels the token-centric payload sender as the old DeepEP kernel.

src/ext/ep/combine/throughput_reduce_combine.cu
Binyang2014: The implementation request is now addressed by [d80bf7c2](https://github.com/microsoft/mscclpp/commit/d80bf7c24c64f8e06c2207b2426d76336d74e3eb) and [8bcb2b9e](https://github.com/microsoft/mscclpp/commit/8bcb2b9e01a8c79a6c1648aa95bacff93a80586c): both modes share `isSupportedRanks()` for every rank count from 1 through 64, and top-k remains limited to 1..8. Throughput no longer divides payload workers by rank count; larger peer sets use top-k destination lanes, peer synchronization covers the full rank set, and combine capacity is bounded by `min(numTopk, numRanks)`. All peers still need to share one CUDA IPC domain.

Closing this implementation discussion rather than leaving it tied to a particular machine size. Physical 32/64-GPU qualification has not been performed on the available eight-GPU machine; this is not a claim of such a hardware run.

include/mscclpp/ext/ep/moe_runtime.hpp
Binyang2014: Simplified the contract in [8bcb2b9e](https://github.com/microsoft/mscclpp/commit/8bcb2b9e01a8c79a6c1648aa95bacff93a80586c): all GPU work sharing a runtime, including expert computation, must use the same supplied CUDA stream. Cross-stream execution/overlap is not supported for now. Calls still enqueue asynchronously with respect to the host; no CPU wait is added to normal operations.

include/mscclpp/ext/ep/moe_runtime.hpp
Binyang2014: Updated in [8bcb2b9e](https://github.com/microsoft/mscclpp/commit/8bcb2b9e01a8c79a6c1648aa95bacff93a80586c) following the final single-stream decision: dispatch consumes the routing handle on the same stream as preparation, so stream order provides readiness. Removed the internal readiness event, capture-status queries and event record/wait logic. The handle carries routing identity/version, not a completion event.

src/ext/ep/include/config.hpp
Binyang2014: Renamed this to `isSupportedRanks()` in [8bcb2b9e](https://github.com/microsoft/mscclpp/commit/8bcb2b9e01a8c79a6c1648aa95bacff93a80586c) and reused it in latency availability/launch validation as well as throughput storage and launch checks. The shared policy is 1 <= numRanks <= 64, without a power-of-two whitelist.

src/ext/ep/include/config.hpp
Binyang2014: Added the layout diagram in [8bcb2b9e](https://github.com/microsoft/mscclpp/commit/8bcb2b9e01a8c79a6c1648aa95bacff93a80586c). It now explicitly shows TOKEN_MAJOR as dense `DataType[numRecvTokens][hidden]`, with no gaps between source-rank batches; unused allocation capacity is only after that data. Metadata has its own fixed-stride slab containing top-k IDs, weights, optional FP8 scales and padding. The BF16-sized allocation reserve keeps `metadataOffset_` fixed. There is no source-token ID in this payload because inverse routing stays in the workspace.

src/ext/ep/include/kernels.hpp
Binyang2014: Kept this shared setup and clarified its purpose in [8bcb2b9e](https://github.com/microsoft/mscclpp/commit/8bcb2b9e01a8c79a6c1648aa95bacff93a80586c). Latency dispatch and latency combine already use this same `configureKernel()` helper. It opts in to the required dynamic shared memory and computes legal grid residency from the actual kernel resources. Throughput additionally uses cooperative grid synchronization, so an arbitrary launch configuration or an SM-count limit alone cannot replace the residency check. The result is cached rather than queried on every invocation.

src/ext/ep/throughput.cc
Binyang2014: Removed the explicit `cudaDeviceSynchronize()` and bootstrap barrier from throughput destruction in [8bcb2b9e](https://github.com/microsoft/mscclpp/commit/8bcb2b9e01a8c79a6c1648aa95bacff93a80586c), matching the latency ownership contract. The unused context bootstrap reference and readiness event were removed too. Callers must finish all local and peer GPU work using the resources before destruction; ordinary CUDA deallocation remains, but the runtime no longer inserts an explicit collective teardown wait.

src/ext/ep/throughput.cc
Binyang2014: Clarified the two host-side lifetimes in [8bcb2b9e](https://github.com/microsoft/mscclpp/commit/8bcb2b9e01a8c79a6c1648aa95bacff93a80586c): `routingEpoch_` changes when routing is rebuilt, while `dispatchEpoch_` changes for each dispatch and whenever routing is rebuilt. A preparation can therefore remain valid across multiple dispatches, while an older dispatch handle is rejected after its shared results have been replaced. These are not GPU communication epochs. The throughput kernels do not consume `Workload::epoch_`; constructor arguments, snapshots and checks now consistently use the routing/dispatch names.

src/ext/ep/dispatch/throughput_dispatch.cu
Binyang2014: Added `warpLaneMask(predicate)` to `device_helpers.cuh` in [8bcb2b9e](https://github.com/microsoft/mscclpp/commit/8bcb2b9e01a8c79a6c1648aa95bacff93a80586c) and reused it in both dispatch paths and throughput combine. Bit i is set exactly when lane i's predicate is true. Its comment states the full-warp participation requirement. For the large-rank dispatch path these bits identify top-k lanes, not global rank IDs.

src/ext/ep/combine/throughput_reduce_combine.cu
Binyang2014: This stages a caller-provided expert-output tensor into the runtime-owned symmetric buffer: peers have mappings for that buffer, not for arbitrary caller allocations. Multiple blocks perform the copy, so the cooperative `this_grid().sync()` must join all writers before block 0 publishes readiness to peers. Kept the grid join and clarified the comments in [8bcb2b9e](https://github.com/microsoft/mscclpp/commit/8bcb2b9e01a8c79a6c1648aa95bacff93a80586c); no host `cudaDeviceSynchronize()` is needed or added. The copy is skipped when the input already aliases the runtime buffer.

src/ext/ep/combine/throughput_reduce_combine.cu
Binyang2014: Changed in [b1f86531](https://github.com/microsoft/mscclpp/commit/b1f86531dc4940d77aeb0735b2581f9d3bef5f83) to use the MSCCL++ packed types and conversion helpers: `bf16x8` -> `f32x8` for loading/accumulation, then `to<bf16x8>()` and `bit_cast<int4>()` at the store boundary. One int4 is 16 bytes, so it contains eight BF16 values rather than four. Accumulation remains FP32, with BF16 conversion only at the final store; the accumulator is explicitly zero-initialized.

test/mp_unit/ep_tests.cu
Binyang2014: Changed `ThroughputPerformance` to eager dispatch in [b1f86531](https://github.com/microsoft/mscclpp/commit/b1f86531dc4940d77aeb0735b2581f9d3bef5f83); it does not capture or replay CUDA graphs. One runtime is initialized with the maximum token capacity before the entire parameter matrix, so communication memory is registered once per rank and reused across all token counts, formats, block budgets and repetitions. The same CUDA stream is reused throughout. Timing uses CUDA events around the eager loop, with initialization/registration outside the measured region.

test/mp_unit/ep_tests.cu
Binyang2014: Simplified in [b1f86531](https://github.com/microsoft/mscclpp/commit/b1f86531dc4940d77aeb0735b2581f9d3bef5f83). Throughput now has just three entry points: `ThroughputCorrectness`, `ThroughputPrepare`, and `ThroughputPerformance`. Removed the standalone storage-layout, raw routing, rank-count and scheduling cases and combined representative data/metadata, format, capacity and top-k coverage into correctness. The prepare entry focuses on explicit preparation, reuse/rebuild and handle lifetime rather than a separate broad validation suite. Throughput coverage is eager; the existing latency graph tests are unchanged.

test/mp_unit/ep_tests.cu
Binyang2014: Follow-up in [1d877d64](https://github.com/microsoft/mscclpp/commit/1d877d6412daecf8f7fb30a3ca9a3464f77ce1a9): corrected the measured region to **eager dispatch + combine**, not dispatch-only. The same initialized runtime and CUDA stream are reused across the whole matrix, so communication memory is registered once. BF16 expert-result inputs are allocated and initialized before timing (also for FP8 dispatch); the reported `us/D+C` includes the two public operations, including combine's normal input staging, but excludes allocation, registration, GEMM and conversion. Automatic preparation is included in the automatic profile; cached routing is prepared outside the timed loop.

src/ext/ep/throughput.cc
```diff
@@ -0,0 +1,294 @@
...
+      outputLayout_(outputLayout),
+      communicator_(communicator) {
+  EP_HOST_ASSERT(hidden_ > 0);
+  EP_HOST_ASSERT(numExperts_ > 0 && numExperts_ % numRanks_ == 0);
+  EP_HOST_ASSERT(numTopk_ > 0 && numTopk_ <= MaxNumTopk);
+  EP_HOST_ASSERT(outputLayout_ == DispatchLayout::TOKEN_MAJOR || outputLayout_ == DispatchLayout::RANK_MAJOR);
+  EP_HOST_ASSERT(maxTokensPerRank_ > 0);
+  EP_HOST_ASSERT(static_cast<int64_t>(hidden_) * sizeof(Bf16) % sizeof(int4) == 0);
+
+  if (!isSupportedRanks(numRanks_) || numRanksPerIpcDomain_ < numRanks_) {
```
seagater, line 39: Should we also check `cudaDevAttrComputeCapabilityMajor >= 9`, and `EP_THROW` a clear message?  The dynamic shared request also exceeds `MaxSharedMemoryPerBlockOptin` in lower arch like sm_80.

src/ext/ep/combine/throughput_reduce_combine.cu
```diff
@@ -0,0 +1,229 @@
...
+                                           : ThroughputTokenRoute{-1, -1};
+    const int slot = workspace.recvTokenIndex(route);
+    const int numContributors = __popc(warpLaneMask(slot >= 0));
+    EP_DEVICE_ASSERT(numContributors <= MaxContributors);
+    for (int contributor = 0; contributor < numContributors; ++contributor) {
+      contributorRanks[contributor] = warpBroadcast(route.rank_, contributor);
+      contributorSlots[contributor] = warpBroadcast(slot, contributor);
+    }
+
+    auto* outputRow = output + static_cast<int64_t>(token) * hiddenInt4;
```
seagater, line 100: combine `output` requires 16-byte alignment, should we check the 16-byte alignment for combine `output` and dispatch `input`?

Binyang2014:
@copilot+gpt-6-astra:xhigh  pls help me resolve the conflict

include/mscclpp/ext/ep/moe_runtime.hpp
```diff
@@ -14,26 +14,30 @@ namespace mscclpp {
...
 ///
-/// One runtime owns the communication buffers and synchronization state for the
-/// selected mode. This library initially implements LATENCY with fixed-capacity
-/// expert-major or rank-major layouts. The THROUGHPUT API is reserved for a
-/// follow-up implementation and is rejected at runtime.
-/// Operations are asynchronous with respect to the host and execute on the
-/// CUDA stream supplied by each request.
+/// Both modes own fixed-capacity symmetric storage sized from the runtime
+/// configuration. LATENCY exposes expert-major or rank-major rows. THROUGHPUT
+/// exposes compact token-major or fixed-stride rank-major rows within that storage.
```
Binyang2014: Why do u say it's symmetric? Does throughput need to copy to scratch?

src/ext/ep/include/config.hpp
```diff
@@ -213,6 +236,150 @@ inline size_t latencyStorageSize(int maxTokensPerRank, int hidden, int numRanks,
...
+//   [padding to metadataSlotBytes_]
+//
+// maxRows = numRanks * configuredMaxTokensPerRank is capacity, not a per-rank stride.
+// The BF16-sized data reservation keeps metadataOffset_ fixed for FP8 too.
+// RANK_MAJOR uses explicit source-rank row ranges instead of the compact view above.
+// Inverse routing stays in the workspace; there is no source-token ID in this payload.
+struct ThroughputPayloadView {
+  int topK_;
+  size_t metadataOffset_;
+  size_t metadataSlotBytes_;
```
Binyang2014: What is slot? For quantization? May change a better name?

src/ext/ep/include/config.hpp
```diff
@@ -213,6 +236,150 @@ inline size_t latencyStorageSize(int maxTokensPerRank, int hidden, int numRanks,
...
+  int* recvCounts_ = nullptr;
+
+  ThroughputWorkspaceLayout(void* workspace, int maxTokensPerRank, int numRanks, int numExperts, int numTopk) {
+    size_t offset = 0;
+    auto place = [&](size_t bytes, size_t alignment) -> void* {
+      offset = configAlign<size_t>(offset, alignment);
+      void* ptr = workspace == nullptr ? nullptr : reinterpret_cast<uint8_t*>(workspace) + offset;
+      offset += bytes;
+      return ptr;
+    };
```
Binyang2014, line 364: What's the difference between playload layout and workspace layout

src/ext/ep/throughput.cc
Binyang2014: I think during compilation, we force to use sm>=90. So the code will no run on arch like sm_80

src/ext/ep/throughput.cc
```diff
@@ -0,0 +1,294 @@
...
+  constexpr uint64_t MaxIndex = std::numeric_limits<int>::max();
+  return hiddenBytes <= MaxIndex && maxRows * static_cast<uint64_t>(numTopk_) <= MaxIndex;
+}
+
+ThroughputStorageLayout ThroughputRuntimeContext::storageLayout() const {
+  return {symmetricBuffer_, maxTokensPerRank_, hidden_, numRanks_, numExperts_, numTopk_};
+}
+
+Workload ThroughputRuntimeContext::makeWorkload(int numTokens, int maxTokensPerRank, DispatchDataType dataType) const {
+  return {.epoch_ = 0,
```
Binyang2014: epoch start from1?

src/ext/ep/throughput.cc
```diff
@@ -0,0 +1,294 @@
...
+    }
+  } else {
+    preparation = prepare(prepareRequest);
+  }
+
+  const ThroughputWorkspaceLayout workspaceLayout(context.workspace_, context.maxTokensPerRank_, context.numRanks_,
+                                                  context.numExperts_, context.numTopk_);
+  if (reusePreparation) {
+    // Replays still need a peer handshake before overwriting the previous payload.
+    throughputSynchronizePeers(context.deviceContext_, request.stream);
```
Binyang2014, line 234: What's this

src/ext/ep/throughput.cc
```diff
@@ -0,0 +1,294 @@
...
+  }
+
+  const ThroughputStorageLayout storageLayout = context.storageLayout();
+  const Workload workload = context.makeWorkload(request.numTokens, request.maxTokensPerRank, request.dispatchDataType);
+  throughputDispatch(request.output, request.outputTopkIdx, request.outputTopkWeights,
+                     static_cast<float*>(request.outputScales), request.input, request.topkIdx, request.topkWeights,
+                     request.inputScales, workload, workspaceLayout, storageLayout.payload_, storageLayout.recvBuffer_,
+                     context.deviceContext_, request.numBlocks, request.stream);
+
+  ++context.dispatchEpoch_;
```
Binyang2014: Also I hope we can do this in combine

Binyang2014:
@copilot resolve the merge conflicts in this pull request

PR #904, opened 2026-09-11
fix sglang pipeline issues

Empyreus:


Binyang2014 (approved):

.azure-pipelines/multi-nodes-test.yml
```diff
@@ -24,6 +24,9 @@ parameters:
     10.0.0.5 mscclpp-h100-multinode-ci000000
     10.0.0.4 mscclpp-h100-multinode-ci000001
 
+variables:
+  system.debug: 'true'
```
Binyang2014: Can we pass it manually when trigger the pipeline?

Binyang2014 (approved):

PR #907, opened 2026-09-15
[MSCCLPP-EP] Python interface for the C++ expert-parallel runtime

Binyang2014:
## Summary

Stacked on #903 (`binyli/ep-throughput-cpp`). Ports the PyTorch-facing EP interface from `feature/ep` (`232c922d`) onto the current C++ latency/throughput runtime, without copying the donor's kernels or legacy receive-pool implementation.

## Changes

- Add the optional `mscclpp.ep` package with `MoECommunicator`, configuration/output types, explicit `prepare`, opaque preparation/dispatch handles, and tensor-based dispatch/combine.
- Add a standalone nanobind module linked to the existing `libmscclpp_ep`. The native boundary uses raw device pointers and does not link libtorch; PyTorch is an optional `ep` dependency.
- Support latency expert-major/rank-major and throughput token-major/rank-major layouts, BF16 and supported FP8 dispatch paths, metadata/scale outputs, external buffers, and active capacities smaller than the allocation.
- Replace the donor's notify/count polling and dispatch caches with GPU-only preparation and current native handles. Throughput returns capacity-sized tensors plus a borrowed GPU receive-row count, rather than synchronizing to construct host-sized outputs.
- Validate tensor shape, dtype, device, alignment, stream, and handle ownership before enqueueing. Retain native owners through tensor storage and handles, and record allocator-backed tensors on the caller stream.
- Stage external latency rank-major expert results in native code after handle validation, so rejected stale handles cannot overwrite the current receive buffer.
- Wire conditional native-module build/install, development copies, relocatable library lookup, and usage/lifetime documentation.

## Contract and scope

All work sharing a runtime uses one caller CUDA stream, including expert computation. There are no internal completion events, cross-stream overlap stubs, or automatic autograd integration. Runtime-owned views and handles retain storage, but callers must finish local and peer GPU use before releasing the final owner.

The underlying implementation requires CUDA SM90+ and 1-64 ranks in one IPC domain, with top-k up to 8. Physical 32/64-GPU qualification is not claimed. The donor's `previous_handle`, overlap configuration, notify matrices, and receive-pool tuning are not carried into this API.

## Validation

- 19 focused Python cases passed on eight H100 GPUs, both from the source tree and from an isolated installed wheel.
- Coverage includes both modes/layouts/formats, prepared routing reuse and invalidation, CPU-readback rejection, zero-token inputs, external outputs, sliced-view ownership, cached CUDA graph replay, and stale rank-major combine rejection.
- All three existing non-performance `MoERuntimeTest` C++ cases passed on eight GPUs.
- The SM90 wheel builds and resolves its packaged native dependencies without libtorch. Repository formatting completed.


src/ext/ep/latency.cc
```diff
@@ -269,8 +269,12 @@ void MoERuntime::launchLatencyCombine(const LatencyCombineRequest& request) {
...
   void* dispatchRecvBuffer = allocationLayout.dispatchRecvBuffer_;
-  if (dispatchLayout == DispatchLayout::RANK_MAJOR) {
-    EP_HOST_ASSERT(input == allocationLayout.combineBuffer_);
+  if (dispatchLayout == DispatchLayout::RANK_MAJOR && input != combineBuffer) {
+    EP_HOST_ASSERT(input != nullptr);
+    const size_t rows = static_cast<size_t>(context.numRanks_) * maxTokensPerRank;
+    const size_t elements = rows * hidden * (mode == CombineMode::DIRECT_SEND ? numTopk : 1);
+    MSCCLPP_CUDATHROW(cudaMemcpyAsync(combineBuffer, input, elements * sizeof(Bf16), cudaMemcpyDeviceToDevice, stream));
+    input = combineBuffer;
   }
```
Binyang2014: WHy change this?

python/mscclpp/ep/communicator.py
```diff
@@ -0,0 +1,674 @@
...
+        if config is not None and kwargs:
+            raise ValueError("pass either MoECommunicatorConfig or keyword arguments, not both")
+        if config is None:
+            config = MoECommunicatorConfig(**kwargs)
+        self._config = _resolve_config(config)
+        self._initialized = False
+        self._stream: Optional[torch.cuda.Stream] = None
+        with torch.cuda.device(self.device):
+            if torch.cuda.is_current_stream_capturing():
+                raise RuntimeError("construct and initialize MoECommunicator before CUDA graph capture")
```
Binyang2014: DO we need to check this? If in graph captre? I think user need to know that. Also really need to set decice here?

python/mscclpp/ep/communicator.py
```diff
@@ -0,0 +1,674 @@
...
+        """Collectively initialize once, outside capture; later calls are no-ops."""
+        if self._initialized:
+            return
+        with torch.cuda.device(self.device):
+            if torch.cuda.is_current_stream_capturing():
+                raise RuntimeError("call initialize() collectively before CUDA graph capture")
+            self._runtime.initialize()
+            self._initialized = True
+
+    def get_dispatch_output_buffer(
```
Binyang2014, line 165: I think we can have an annotation for initlization right?

include/mscclpp/ext/ep/types.hpp
```diff
@@ -222,7 +222,13 @@ class DispatchHandle {
...
   void* output;
-  /// Local expert output.
+  /// Local BF16 expert output.
+  ///
+  /// Rank-major external input is staged into combineInputBuffer() on the
+  /// caller stream after handle validation. It must contain [numRanks,
+  /// active capacity, hidden] elements, with an extra top-k dimension before
+  /// hidden for DIRECT_SEND. It may alias the runtime buffer exactly, but
+  /// must not overlap it partially.
   const void* input;
```
Binyang2014, line 230: Do we need this? This should be reallocated by MoERuntime right, not expose to user

include/mscclpp/ext/ep/moe_runtime.hpp
```diff
@@ -79,6 +79,16 @@ class MoERuntime {
...
+  /// Return the runtime-owned device scalar containing the throughput receive row count.
+  ///
+  /// Preparation updates this count asynchronously. Consume it on the same stream
+  /// after preparation, without a host readback. The count excludes duplicate
+  /// destinations and rank-major padding, and is overwritten by the next preparation.
+  /// The returned pointer remains valid until this runtime is destroyed. Do not write
+  /// to this scalar: it is runtime-owned routing metadata.
+  /// @return A borrowed, read-only pointer to one device-resident int.
+  /// @throws EPException If the runtime is not initialized or is not in THROUGHPUT mode.
+  const int* numRecvTokensBuffer() const;
```
Binyang2014: Nameing is confused here

src/ext/ep/throughput.cc
```diff
@@ -150,6 +150,15 @@ Workload ThroughputRuntimeContext::makeWorkload(int numTokens, int maxTokensPerR
           .dispatchDataType_ = dataType};
 }
 
+const int* MoERuntime::numRecvTokensBuffer() const {
```
Binyang2014: Why need to expose this API?

src/ext/ep/latency.cc
Binyang2014: This supports expert results produced in an ordinary PyTorch allocation while preserving the registered-buffer requirement of rank-major communication. MoERuntime already preallocates the destination; this only stages external results on the supplied stream, and an input that already aliases combineInputBuffer() takes the no-copy path.

The placement in C++ is deliberate: owner/epoch validation happens before the copy. Copying in Python first would let a rejected stale handle overwrite the current runtime buffer. test_stale_rank_major_combine_preserves_buffer covers that failure for both latency combine modes. I have retained this implementation and clarified ownership in the request documentation in 312b9084; leaving this design discussion open for your review.

python/mscclpp/ep/communicator.py
Binyang2014: Removed the Python capture-state queries from construction and initialization in 312b9084. Constructing/initializing before graph capture is now an explicit caller precondition.

I also removed the availability getter's device guard and the redundant nested device guards around stream-scoped operations: torch.cuda.stream already selects and restores the stream's device. The construction/initialization guards remain to honor an explicit config.device when another GPU is current: the latency C++ constructor records cudaGetDevice(), and initialize allocates native resources on the current device. Buffer views also stay scoped to their configured device. These guards restore the previous device rather than permanently changing it. Added coverage for this explicit-device case and for operation without Python capture queries.

python/mscclpp/ep/communicator.py
Binyang2014: Added @requires_initialized in 312b9084, following the donor's decorator pattern. get_dispatch_output_buffer(), prepare(), and dispatch() now share it instead of repeating initialize() calls. initialize() remains idempotent, and functools.wraps preserves the method metadata. combine() already requires an existing dispatch handle, so it does not need a separate lazy-initialization path.

include/mscclpp/ext/ep/types.hpp
Binyang2014: The input field is inherited from the base C++ API: it supplies the result of expert computation, not an allocation request or ownership of communication storage. MoERuntime does preallocate the rank-major registered buffer. Expert computation can write directly into combineInputBuffer(), in which case no staging copy is made; externally produced PyTorch results are copied into that same preallocated storage.

I clarified this distinction in the field documentation in 312b9084. Keeping the input argument is also necessary for the existing expert-major path. Leaving the interface-design discussion open rather than assuming you want that existing input contract removed.

include/mscclpp/ext/ep/moe_runtime.hpp
Binyang2014: Renamed this to numRecvTokensDevicePtr() in 312b9084, with num_recv_tokens_device_ptr() in the native Python binding. The declaration, definition, binding, and Python caller are updated together. The name and documentation now explicitly identify a borrowed pointer to one GPU-resident count, rather than a host integer or a newly allocated payload buffer.

src/ext/ep/throughput.cc
Binyang2014: The Python TOKEN_MAJOR result has a capacity-sized tensor, so GPU consumers need the actual number of valid received rows without copying a count to the CPU. The existing outputCount is per expert: a token routed to two local experts contributes two expert counts but occupies only one received row, so summing those counts is incorrect.

This accessor only borrows the scalar already allocated in the native workspace. It adds no allocation, count-computation kernel, or host synchronization; Python exposes it as read-only-by-contract layout.num_recv_tokens. I renamed it numRecvTokensDevicePtr() in 312b9084 to make the device-pointer semantics clear. Leaving the public-interface design question open for your review.

include/mscclpp/ext/ep/types.hpp
Binyang2014: The commment is confused. Pls say for rank major pls use buffer retured by handle. For other layout can use mlp output

python/mscclpp/ep/types.py
```diff
@@ -0,0 +1,146 @@
...
+    format: Optional[DispatchDataType] = None
+    block_scales: Optional[torch.Tensor] = None
+
+
+@dataclass
+class MoECommunicatorConfig:
+    """Fixed runtime configuration, which must agree across participating ranks.
+
+    Experts are partitioned evenly into contiguous rank-local ranges. A scalar
+    ``num_blocks=N`` resolves to ``(N, N - 2)`` for latency or ``(N, N)`` for
```
Binyang2014, line 38: What's the N, N-2 meaning

python/mscclpp/ep/utils.py
```diff
@@ -0,0 +1,150 @@
...
+def record_stream(tensors: Iterable[Optional[torch.Tensor]], stream: torch.cuda.Stream) -> None:
+    """Protect allocator-backed storage used by native asynchronous GPU work.
+
+    PyTorch's allocator ignores foreign storage. Runtime buffers instead rely on
+    their native owner and the caller's obligation to finish local and peer use.
+    This does not provide producer/consumer ordering across streams.
+    """
+    for tensor in tensors:
+        if tensor is not None:
+            tensor.record_stream(stream)
```
Binyang2014, line 107: Remove this function?

python/mscclpp/ep/utils.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+class DevicePointerArray:
```
Binyang2014, line 110: Why need this?

src/ext/ep/CMakeLists.txt
```diff
@@ -62,6 +62,7 @@ set_target_properties(mscclpp_ep PROPERTIES
     CUDA_ARCHITECTURES "${_mscclpp_ep_gpu_archs}"
     LINKER_LANGUAGE CXX
     POSITION_INDEPENDENT_CODE ON
+    INSTALL_RPATH "\$ORIGIN"
```
Binyang2014, line 65: WHy need this?

python/mscclpp/ep/types.py
Binyang2014: Clarified in 757abfa4. `num_blocks=(D, C)` now explicitly means the dispatch and combine grid sizes. For a scalar `N`, dispatch uses `N` blocks; latency reserves two dispatch control blocks, so combine uses the remaining `N - 2` worker blocks, while throughput uses `N` for both operations.

python/mscclpp/ep/communicator.py
```diff
@@ -0,0 +1,609 @@
...
+    expert computation must use the same stream as communication. This interface
+    supplies no autograd backward.
+    """
+
+    def __init__(self, config: Optional[MoECommunicatorConfig] = None, **kwargs) -> None:
+        if config is not None and kwargs:
+            raise ValueError("pass either MoECommunicatorConfig or keyword arguments, not both")
+        if config is None:
+            config = MoECommunicatorConfig(**kwargs)
+        self._config = _resolve_config(config)
```
caiocbr, line 52: Could we move validation of the configuration values into `MoECommunicatorConfig` initialization (`__post_init__` for the dataclass)? Checks such as positive dimensions, top-k limits, and compatible mode/layout settings could fail when the user creates the config, rather than waiting until it is passed to the communicator. Device/rank-dependent default resolution and runtime checks could stay here. Together with a config-only constructor, this would make the configuration contract more explicit. Is there a reason to keep all validation in `_resolve_config`? If we validate only at config construction, we'd also need to prevent later mutation or revalidate when consuming it.

python/mscclpp/ep/communicator.py
```diff
@@ -0,0 +1,609 @@
...
+    """Collective MoE dispatch/combine on one caller-owned CUDA stream.
+
+    LATENCY defaults to EXPERT_MAJOR; THROUGHPUT defaults to TOKEN_MAJOR.
+    Construction and collective initialization must precede graph capture.
+    Operations otherwise initialize lazily. Host calls must be serialized, and
+    expert computation must use the same stream as communication. This interface
+    supplies no autograd backward.
+    """
+
+    def __init__(self, config: Optional[MoECommunicatorConfig] = None, **kwargs) -> None:
```
caiocbr, line 47: What is the intended benefit of supporting `**kwargs` alongside `MoECommunicatorConfig`? For a public API, I'd prefer a single explicit constructor, `__init__(self, config: MoECommunicatorConfig)`, with callers creating the config first. Although unknown keywords are rejected when forwarded to the dataclass, this generic signature hides the supported parameters from the communicator's signature and weakens IDE completion and static checking. Is there a use case or compatibility requirement that calls for both construction styles?

python/mscclpp/ep/types.py
```diff
@@ -0,0 +1,138 @@
...
+
+from . import _cpp
+from ._cpp import CombineMode, DispatchDataType, DispatchLayout, MoEMode
+
+
+@dataclass
+class QuantConfig:
+    """Payload format and optional FP32 scales, one per 128 hidden elements.
+
+    Latency dispatch quantizes BF16 input and returns scales with logical shape
```
caiocbr, line 23: Make clear here we only support FP8, for this reason the shape is  hidden // 128.

python/mscclpp/ep/types.py
```diff
@@ -0,0 +1,138 @@
...
+    max_tokens_per_rank: int = 0
+    mode: MoEMode = MoEMode.LATENCY
+    output_layout: Optional[DispatchLayout] = None
+    invalid_token_expert_id: Optional[int] = None
+    quant: Optional[QuantConfig] = None
+    num_blocks: Optional[Union[int, Tuple[Optional[int], Optional[int]]]] = None
+    combine_mode: CombineMode = CombineMode.RANK_LOCAL_REDUCE
+
+
+@dataclass
```
caiocbr, line 62: Would it be clearer to expose separate metadata classes for each layout, such as `RankMajorLayoutInfo`, `ExpertMajorLayoutInfo`, and `TokenMajorLayoutInfo`, and use a union for the output's layout type? For example, rank-major would require `num_tokens_per_rank` without exposing the unused `num_tokens_per_expert` field. This would make the available metadata explicit and avoid combinations such as a rank-major layout with missing rank counts or unrelated expert counts. A separate token-major type would also help distinguish its per-expert route counts from the row counts in expert-major. Is there a reason to prefer one class with optional fields for all layouts?

PR #908, opened 2026-09-17
EP: integrate GPUNetIO and rank-major top-k-expanded layout

seagater:
## Summary
Integrate GPU-initiated networking into the newer feature/ep runtime and add the GPUNetIO-capable rank-major top-k-expanded layout, including topology-aware multi-HCA support and default-enabled expanded fastpaths.

- **Base:** `feature/ep-experimental` (`63fea77` at creation).
- **Head:** `qinghuazhou/gpunetio_port_channel_merge_feature_ep` (`824823a`).
- Preserve the newer `DeviceContext`, `LatencyStorageLayout`, generic dispatch handles, and `MoEMode.LATENCY` APIs rather than restoring the legacy runtime API.

## GPUNetIO
- Add the DOCA GPUNetIO/GDAKI device-initiated transport and host service for QP setup, symmetric-memory registration, and remote-key exchange. IPC/NVLink-mapped peers retain mapped-memory communication; GPUNetIO handles peers outside that domain.
- Support multiple HCAs and QPs per peer, per-HCA registrations and keys, ordered completion markers, batched writes, and cooperative/deferred queue drains. Keep WQEs CQ-signaled for queue recycling and retain system-scope publication fences.
- Auto-select all active HCAs tied for a GPU's best PCI/NUMA affinity, sorted by device name. Nearby GPUs may share the full HCA set instead of splitting it by local rank.
- `MSCCLPP_EP_GPUNETIO_HCAS` overrides `MSCCLPP_EP_GPUNETIO_HCA`; otherwise topology selection is automatic. `MSCCLPP_EP_GPUNETIO_QPS_PER_PEER` defaults to the selected HCA count. Ranks must agree on HCA/QP counts; QPs must be in [1, 64] and a positive multiple of the HCA count.
- Networking remains opt-in through the build option `MSCCLPP_USE_GPUNETIO` and runtime `MSCCLPP_EP_ENABLE_GPUNETIO=1`. Compact cross-domain rank-major uses BF16 rank-local-reduce; expert-major and compact direct-send remain IPC-only.

## Rank-Major Top-K-Expanded
- Expose `DispatchLayout.RANK_MAJOR_TOPK_EXPANDED` with logical rows `[source_rank, original_token, topk_slot, hidden]`. Python tokens are `[world_size * capacity * topk, hidden]`, with flat row-aligned int32 expert IDs and FP32 weights.
- Preserve duplicate top-k slots and fixed positions without compaction. Nonlocal, invalid, and padding metadata use the configured invalid ID and zero weight; inactive payload rows are unspecified. Per-rank counts include valid local selections, including duplicates and zero-weight selections.
- Alias dispatched tokens and the registered combine-input buffer. The consumer writes **unweighted per-slot expert results** in place; combine applies the original source routing weights exactly once, in top-k order with FP32 accumulation, before writing BF16 output. Invalid IDs and zero weights are skipped before payload reads.
- Restrict expanded mode to BF16, fixed configured capacity, top-k 1-9, serialized calls, and the `RANK_LOCAL_REDUCE` mode setting. Preserve existing compact rank-major behavior and buffer contracts.
- Use TMA scatter/gather for mapped peers, registered staging and per-HCA pushes for network peers, device-generation completion, and collective lifetime acknowledgments to protect aliased-buffer reuse across CUDA-graph replays.
- Add Python validation and benchmark selection via `--ep-layout rank_major_topk_expanded` for MSCCL++ latency mode, plus correctness coverage for padding, duplicate routes, oversized invalid IDs, zero weights/NaN inactive rows, and repeated graph pairs/replays.

## Expanded Fastpaths
- Default-enable the IPC specialization only when every peer is mapped and no GPUNetIO context is active, or the network specialization when GPUNetIO is already active outside the IPC domain. Select the protocol collectively before graph capture.
- Retain baseline kernels as opt-out paths: `MSCCLPP_EP_EXPANDED_IPC_FASTPATH=0` and `MSCCLPP_EP_EXPANDED_GPUNETIO_FASTPATH=0`. An opt-out disables the corresponding fastpath collectively; these flags do not enable networking itself.
- Optimize worker retirement, cached receive readiness, metadata submission, and sparse warp-batched network returns without removing the final peer acknowledgment or outgoing queue drains.

## Validation
Validation recorded during the port series:
- Latest clean fastpath commit: **55 CPU regressions passed**, covering API/allocation contracts, collective selection/opt-out, retirement/ACK ordering, sparse WQE ticket wrap, per-HCA keys, and metadata transfer-size fallback.
- **152 topology-selection checks passed**, including CPU-only CTest with GPUNetIO enabled and disabled.
- Full SM100a EP builds passed with GPUNetIO ON/OFF; clean committed snapshots were also compiled on the compile-only host. Expanded kernels compiled for SM90 with GPUNetIO ON/OFF.
- Changed-file formatting and diagnostics passed. Whole-repository C++ lint still reports pre-existing formatting debt.
- **Distributed GPU correctness and performance remain unverified.** CPU models/stubs and compilation do not establish GPU/NIC ordering or performance. SM90 network-combine fastpaths report register spills (TMA: 36-byte stores/68-byte loads; vector: 40-byte stores/92-byte loads), requiring runtime evaluation. SM100a expanded compilation reported no spills.

## Merge Notes
- GitHub currently reports **cannot automatically merge** with the requested base; conflict resolution is still required.
- This comparison spans **206 source commits and 87 changed files**, including merged feature/ep architecture history and vendored GPUNetIO dependencies, not only the most recent fastpath commit. The base has 12 commits absent from the head.
- Earlier local initialization/IPC diagnostic edits were deliberately excluded from the published commits and are not part of this PR. No runtime deployment or GPU jobs were performed as part of creating it.


CMakeLists.txt
```diff
@@ -55,6 +55,7 @@ option(MSCCLPP_BUILD_EXT_EP "Build Expert-Parallel (MoE dispatch/combine) extens
 option(MSCCLPP_USE_CUDA "Use NVIDIA/CUDA." OFF)
 option(MSCCLPP_USE_ROCM "Use AMD/ROCm." OFF)
 option(MSCCLPP_USE_IB "Use InfiniBand." ON)
+option(MSCCLPP_USE_GPUNETIO "Build the vendored DOCA GPUNetIO (GDAKI) GPU-initiated networking backend for PortChannel." OFF)
```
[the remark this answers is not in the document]
seagater: Resolved.

include/mscclpp/port_channel_device.hpp
```diff
@@ -59,6 +77,11 @@ struct BasePortChannelDeviceHandle {
   /// @param size The size of the transfer.
   MSCCLPP_DEVICE_INLINE void put(MemoryId dstId, uint64_t dstOffset, MemoryId srcId, uint64_t srcOffset,
                                  uint64_t size) {
+    if (backend_ == PortChannelBackend::GpuNetIo) {
+      // Peer to reach is encoded by the destination MemoryId in the symmetric model.
+      gin_->put(static_cast<int>(dstId), dstOffset, srcOffset, size);
```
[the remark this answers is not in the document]
seagater: Resolved.

include/mscclpp/port_channel_device.hpp
```diff
@@ -82,6 +105,12 @@ struct BasePortChannelDeviceHandle {
   /// @param size The size of the transfer.
   MSCCLPP_DEVICE_INLINE void putWithSignal(MemoryId dstId, uint64_t dstOffset, MemoryId srcId, uint64_t srcOffset,
                                            uint64_t size) {
+    if (backend_ == PortChannelBackend::GpuNetIo) {
+      // Signal target offset/value are supplied by the higher-level channel; the
+      // base handle fuses payload+signal on the same peer (dstId).
+      gin_->putWithSignal(static_cast<int>(dstId), dstOffset, srcOffset, size, /*signalOffset=*/0, /*signalValue=*/1);
+      return;
```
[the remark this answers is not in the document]
seagater: Resolved.

src/ext/ep/include/moe_runtime_context.hpp
```diff
@@ -55,6 +60,7 @@ struct LatencyContext {
   void** peerMappedBufferBasesGpu_ = nullptr;
   std::vector<mscclpp::BaseMemoryChannel> baseMemoryChannels_;
   std::shared_ptr<mscclpp::BaseMemoryChannelDeviceHandle> baseMemoryChannelHandles_;
+  std::shared_ptr<mscclpp::GpuNetIoService> gpuNetIoService_;
```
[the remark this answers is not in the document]
seagater: resolved by calling gpuNetIoService_.reset() to release QPs and MRs in latency.cc

src/ext/ep/include/config.hpp
```diff
@@ -171,65 +180,116 @@ struct LatencyStorageLayout {
...
+        BufferAlignmentBytes);
+    const int gpuNetIoStagingRows =
+        topkExpanded && maxTokensPerRank > GpuNetIoStagingSlots ? maxTokensPerRank : GpuNetIoStagingSlots;
+    const size_t gpuNetIoStagingBytes =
+        configAlign<size_t>(static_cast<size_t>(gpuNetIoStagingRows) * gpuNetIoSlotStride_, BufferAlignmentBytes);
+    const size_t gpuNetIoFlagsBytes = configAlign<size_t>(
+        static_cast<size_t>(numRanks) * GpuNetIoMaxQpsPerPeer * sizeof(uint64_t), BufferAlignmentBytes);
+    const size_t gpuNetIoCombineFlagsBytes = configAlign<size_t>(
+        static_cast<size_t>(numRanks) * GpuNetIoMaxQpsPerPeer * sizeof(uint64_t), BufferAlignmentBytes);
+    const size_t gpuNetIoCombineLandingBytes = configAlign<size_t>(rankMajorDispatchOutputBytes, BufferAlignmentBytes);
```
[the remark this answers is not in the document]
seagater: Resolved. config.hpp:238: staging and landing sizes are zero unless GPUNetIO is selected. Expanded IPC retains only its required small synchronization regions.
latency.cc:101: ranks collectively verify backend selection, IPC-domain size, and allocation size before allocating.

src/gpunetio/host/gpu_net_io_service.cpp
```diff
@@ -0,0 +1,522 @@
...
+    MSCCLPP_DOCA_THROW(doca_verbs_qp_modify(qp->qp, attr,
+                                            DOCA_VERBS_QP_ATTR_NEXT_STATE | DOCA_VERBS_QP_ATTR_ALLOW_REMOTE_WRITE |
+                                                DOCA_VERBS_QP_ATTR_ALLOW_REMOTE_READ | DOCA_VERBS_QP_ATTR_PKEY_INDEX |
+                                                DOCA_VERBS_QP_ATTR_PORT_NUM));
+
+    // INIT -> RTR
+    MSCCLPP_DOCA_THROW(doca_verbs_qp_attr_set_next_state(attr, DOCA_VERBS_QP_STATE_RTR));
+    MSCCLPP_DOCA_THROW(doca_verbs_qp_attr_set_rq_psn(attr, 0));
+    MSCCLPP_DOCA_THROW(doca_verbs_qp_attr_set_dest_qp_num(attr, remote.qpn));
+    MSCCLPP_DOCA_THROW(doca_verbs_qp_attr_set_path_mtu(attr, pathMtu(remote)));
```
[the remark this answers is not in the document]
seagater: Resolved. Fixed this in gpu_net_io_service.cpp:252:
Queries the selected local HCA’s active_mtu.
Validates both MTU values.
Programs path_mtu using std::min(localMtu, remote.activeMtu).

src/gpunetio/src/doca_gpunetio.cpp
```diff
@@ -0,0 +1,1211 @@
...
+  }
+
+  mentry = it->second;
+
+  if (mentry->mtype == DOCA_GPU_MEM_TYPE_GPU)
+    DOCA_VERBS_CUDA_CALL_CLEAR_ERROR(cudaFree((void *)mentry->base_addr));
+  else if (mentry->mtype == DOCA_GPU_MEM_TYPE_GPU_CPU) {
+    if (gpu_dev->support_gdrcopy)
+      doca_gpu_gdrcopy_destroy_mapping(mentry->gdr_mh, (void *)mentry->align_addr_cpu, mentry->size);
+    DOCA_VERBS_CUDA_CALL_CLEAR_ERROR(cudaFree((void *)mentry->base_addr));
```
mahdiehghazim, line 407: When `GPU_CPU` falls back because GDRCopy is unavailable, the backing storage was allocated with `calloc` and registered with `cudaHostRegister`, but `mtype` remains `GPU_CPU`, so this path calls `cudaFree` on a host pointer and then drops the bookkeeping entry. That leaves the allocation pinned and leaked. Please track the effective backing type and use `cudaHostUnregister` plus `free` for the fallback allocation.

src/gpunetio/host/gpu_net_io_service.cpp
```diff
@@ -0,0 +1,528 @@
...
+  for (int r = 0; r < s.worldSize; ++r) {
+    for (int hca = 0; hca < nHcas; ++hca) {
+      const auto& info = memAll[static_cast<size_t>(r) * nHcas + hca];
+      rkeysHost[static_cast<size_t>(hca) * s.worldSize + r] = htobe32(info.rkey);
+    }
+    baseHost[r] = static_cast<uintptr_t>(memAll[static_cast<size_t>(r) * nHcas].base);
+  }
+
+  // 7. Publish device-side arrays + context.
+  MSCCLPP_CUDA_THROW(cudaMalloc(&s.rkeysGpu, sizeof(uint32_t) * rkeysHost.size()));
```
mahdiehghazim, line 522: `cudaDeviceId` is used to identify the GPU but `setup()` never makes it current. These `cudaMalloc` calls, the allocations inside QP export, and the matching frees in `Impl::~Impl()` all operate on the calling thread's current CUDA device. Direct users of `GpuNetIoService` can therefore allocate/export the context on a different GPU from the one passed to the constructor (or tear it down after switching devices). Please guard setup and teardown with the stored device ordinal.

src/gpunetio/src/doca_gpunetio_high_level.cpp
```diff
@@ -0,0 +1,933 @@
...
+  if (status != DOCA_SUCCESS) {
+    DOCA_LOG(LOG_ERR, "Failed to create doca verbs qp");
+    goto exit_error;
+  }
+
+  status = doca_gpu_verbs_export_qp(qp_->gpu_dev, qp_->qp, qp_->nic_handler, qp_->qp_umem_gpu_ptr, qp_->cq_sq,
+                                    qp_->send_dbr_mode_ext, &qp_->qp_gverbs);
+  if (status != DOCA_SUCCESS) {
+    DOCA_LOG(LOG_ERR, "Failed to create GPU verbs QP");
+    return status;
```
mahdiehghazim: Returning directly here bypasses `exit_error`, so an export failure leaks the already-created QP, CQ, UAR, UMEM registrations, GPU buffers, and `qp_` itself while returning no handle that the caller can destroy. Route this branch through the existing cleanup label; the two corresponding group-export failures at lines 827 and 864 need the same treatment.

include/mscclpp/port_channel_device.hpp
```diff
@@ -122,6 +198,11 @@ struct BasePortChannelDeviceHandle {
   /// Push a TriggerSync to the FIFO.
   /// @param maxSpinCount The maximum number of spin counts before asserting. Never assert if negative.
   MSCCLPP_DEVICE_INLINE void flush(int64_t maxSpinCount = 1000000) {
+    if (backend_ == PortChannelBackend::GpuNetIo) {
+      validateGpuNetIo();
+      gin_->flush(ginPeer_);
```
mahdiehghazim: `flush(maxSpinCount)` drops the caller's finite spin budget for the GPUNetIO backend and enters an unbounded CQ poll. A stalled QP can therefore hang the kernel indefinitely instead of triggering the documented diagnostic. Please use `tryFlush` for nonnegative budgets and handle timeout/CQ errors consistently, retaining the blocking loop only for the negative “never assert” case if that behavior is intended.

src/gpunetio/src/doca_gpunetio.cpp
```diff
@@ -0,0 +1,1211 @@
...
+  pthread_rwlock_unlock(&service_->service_lock);
+
+  return DOCA_SUCCESS;
+}
+
+doca_error_t doca_gpu_verbs_destroy_service(doca_gpu_verbs_service_t service) {
+  struct doca_gpu_verbs_service *service_ = (struct doca_gpu_verbs_service *)service;
+  if (service == nullptr) return DOCA_ERROR_INVALID_VALUE;
+
+  service_->running = false;
```
mahdiehghazim: `running` is read continuously by the service thread and written here by another thread without atomic operations or the service lock. This is a C++ data race (undefined behavior), and an optimized worker may never observe shutdown, leaving `pthread_join` blocked. Make the flag atomic with acquire/release operations or synchronize both accesses under the lock.

include/mscclpp/internal/port_channel_gpunetio_device_impl.hpp
```diff
@@ -0,0 +1,145 @@
...
+
+MSCCLPP_DEVICE_INLINE void GpuNetIoDeviceContext::putWithSignal(int peer, uint64_t dstOffset, uint64_t srcOffset,
+                                                                uint64_t size, uint64_t signalOffset,
+                                                                uint64_t signalValue, int qpIndex) {
+  const uint32_t remoteKey = detail::ginRemoteKey(*this, peer, qpIndex);
+  const __be32 localKey = detail::ginHtobe32(detail::ginLocalKey(*this, qpIndex));
+  doca_gpu_dev_verbs_addr raddr{peerBase[peer] + dstOffset, remoteKey};
+  doca_gpu_dev_verbs_addr laddr{localBase + srcOffset, localKey};
+  doca_gpu_dev_verbs_addr sigR{peerBase[peer] + signalOffset, remoteKey};
+  doca_gpu_dev_verbs_addr sigL{localBase, localKey};
```
mahdiehghazim: **Blocking:** an RDMA fetch-add writes the fetched 8-byte value to `sig_laddr`; the atomic WQE's data segment uses this address as its local destination. Pointing `sigL` at `localBase` therefore overwrites caller-owned symmetric-buffer bytes `[0, 8)` on every `putWithSignal` (and `atomicAdd` repeats the same problem below), with all QPs also racing on that location. The added P2P test uses offset 0 for payload, so this is not a reserved prefix. Please allocate registered backend-owned result scratch, preferably per flat QP, and publish its address/key in the device context.

src/gpunetio/include/device/doca_gpunetio_dev_verbs_onesided.cuh
```diff
@@ -0,0 +1,797 @@
...
+}
+
+template <
+    typename T,
+    enum doca_gpu_dev_verbs_resource_sharing_mode resource_sharing_mode = DOCA_GPUNETIO_VERBS_RESOURCE_SHARING_MODE_GPU,
+    enum doca_gpu_dev_verbs_nic_handler nic_handler = DOCA_GPUNETIO_VERBS_NIC_HANDLER_AUTO>
+__device__ static __forceinline__ void doca_gpu_dev_verbs_p_warp(
+    struct doca_gpu_dev_verbs_qp *qp, struct doca_gpu_dev_verbs_addr raddr, T value,
+    doca_gpu_dev_verbs_ticket_t *out_ticket, uint32_t code_opt = DOCA_GPUNETIO_VERBS_GPU_CODE_OPT_DEFAULT) {
+  *out_ticket = 0;
```
mahdiehghazim: Selecting `EXEC_SCOPE_WARP` for this documented API silently performs no operation and reports ticket 0. The warp variants of put-with-signal and signal below are identical no-ops, so callers can observe apparent completion without any WQE being posted. Please implement these paths or reject warp scope at compile time rather than silently succeeding.

python/mscclpp/ep/high_throughput.py
```diff
@@ -0,0 +1,530 @@
...
+previous routing matrices and receive count.
+"""
+
+from __future__ import annotations
+
+from typing import Any, List, Optional
+
+import torch
+
+from ._cpp import Config, DispatchLayout, MoEMode, _cpp
```
mahdiehghazim: This added module cannot be imported: `_cpp.py` exports no `Config`, and the module also imports `HighThroughputCombineContext` / `HighThroughputDispatchHandle`, which do not exist in `types.py`. It additionally uses the obsolete `MoEMode.HIGH_THROUGHPUT`. Please either port this implementation to the current `Context`/`Runtime` API and type names or remove the stale duplicate module.

src/gpunetio/include/host/doca_gpunetio.h
```diff
@@ -0,0 +1,484 @@
...
+
+/**
+ * @file doca_gpunetio.h
+ * @brief A header file for the doca_gpunetio APIs
+ */
+
+#ifndef DOCA_GPUNETIO_H
+#define DOCA_GPUNETIO_H
+
+#include "../common/doca_gpunetio_verbs_def.h"
```
mahdiehghazim, line 39: This is a C-facing header and uses `bool` in public structures/signatures, but neither it nor its include chain includes `<stdbool.h>`. A C translation unit including this header fails to compile; the same applies to the other public GPUNetIO C headers. Please include `<stdbool.h>` in the common public header chain.

python/mscclpp/ep/low_latency.py
```diff
@@ -0,0 +1,585 @@
...
+"""Low-latency backend for the high-level MoE communicator."""
+
+from __future__ import annotations
+
+from typing import Any, Optional
+
+import torch
+
+from ._cpp import CombineMode, DispatchDataType, DispatchLayout, MoEMode, create_moe_runtime
+from .types import (
```
mahdiehghazim: This added module cannot be imported because all of `ExpertMajorDispatchHandle`, `ExpertMajorCombineContext`, `RankMajorCombineContext`, and `RankMajorDispatchHandle` are absent from `types.py` (the current implementation uses private combine contexts behind `DispatchHandle`). The file also reads removed config fields such as `low_latency_num_blocks`. Please port it to the current API or remove this stale duplicate of `latency.py`.

src/gpunetio/src/doca_verbs_qp.cpp
```diff
@@ -0,0 +1,2745 @@
...
+doca_verbs_qp::~doca_verbs_qp() { static_cast<void>(destroy()); }
+
+uint32_t doca_verbs_qp::get_qpn() const noexcept { return m_qp_num; }
+
+void *doca_verbs_qp::get_dbr_addr() const noexcept { return (void *)m_db_buffer; }
+
+void *doca_verbs_qp::get_uar_addr() const noexcept { return (void *)m_uar_db_reg; }
+
+enum doca_verbs_uar_allocation_type doca_verbs_qp::get_uar_mtype() const noexcept {
+  return m_init_attr.external_uar->get_uar_mtype();
```
mahdiehghazim: QP creation explicitly supports an internal UAR when `external_uar == nullptr`, but this accessor unconditionally dereferences `external_uar`. `doca_gpu_verbs_export_qp()` calls it, so exporting any QP that used the supported internal-UAR path crashes. Store the internal UAR allocation type (including the BF-to-NC fallback result) and return it here when no external UAR exists.

src/gpunetio/src/doca_verbs_mlx5dv_wrapper.cpp
```diff
@@ -0,0 +1,268 @@
...
+#include "doca_gpunetio_log.hpp"
+#include "doca_verbs_net_wrapper.h"
+#include "host/doca_error.h"
+
+/* *********** dlopen Function Pointers *********** */
+
+static void *mlx5dv_handle = NULL;
+
+/* Function pointer types */
+typedef int (*mlx5dv_init_obj_func_t)(struct mlx5dv_obj *obj, enum mlx5dv_obj_type obj_type);
```
mahdiehghazim: The actual `mlx5dv_init_obj` ABI takes a `uint64_t obj_type`, while this `dlsym` function pointer declares the second parameter as `enum mlx5dv_obj_type`. Calling a dynamically resolved function through an incompatible function type is undefined behavior and is rejected by type-based CFI. Declare the resolved function pointer with the exact `uint64_t` signature; the wrapper may still accept the enum and convert at the call boundary.

src/gpunetio/src/doca_gpunetio.cpp
```diff
@@ -0,0 +1,1211 @@
...
+        goto error;
+      }
+    } else {
+      DOCA_LOG(LOG_WARNING,
+               "GDRCopy not enabled, can't allocate memory type DOCA_GPU_MEM_TYPE_GPU_CPU. "
+               "Using DOCA_GPU_MEM_TYPE_CPU_GPU mode instead");
+
+      mentry->size_orig = mentry->size;
+
+      memptr_cpu_ = (uint8_t *)calloc(alignment, mentry->size_orig);
```
mahdiehghazim: `calloc(alignment, size_orig)` allocates `alignment * size_orig` bytes and does not guarantee `alignment`-byte alignment. With the default 4096-byte alignment, a 1 MiB request attempts 4 GiB, while the returned pointer may still be unsuitable for the requested registration alignment. The `CPU_GPU` branch at line 345 has the same issue. Please use an aligned allocation API and zero exactly the requested allocation size.

src/gpunetio/src/doca_gpunetio_high_level.cpp
```diff
@@ -0,0 +1,933 @@
...
+
+  return DOCA_SUCCESS;
+}
+
+doca_error_t doca_gpu_verbs_create_qp_hl(struct doca_gpu_verbs_qp_init_attr_hl *qp_init_attr,
+                                         struct doca_gpu_verbs_qp_hl **qp) {
+  doca_error_t status = DOCA_SUCCESS;
+
+  if (qp_init_attr == nullptr || qp == nullptr) {
+    DOCA_LOG(LOG_ERR, "Invalid input value: qp_init_attr %p qp %p", (void *)qp_init_attr, (void *)*qp);
```
mahdiehghazim: When `qp == nullptr`, the error-reporting expression `(void *)*qp` dereferences that null output pointer, so invalid input crashes instead of returning `DOCA_ERROR_INVALID_VALUE`. Log `qp` itself; the group constructor repeats this with `*qpg` at line 756.

test/python/ep/test_low_latency_multirank.py
```diff
@@ -0,0 +1,991 @@
...
+        topk_idx[random.randint(0, num_tokens - 1), random.randint(0, num_topk - 1)] = -1
+
+    moe_comm = ep.MoECommunicator(
+        comm=ep_group,
+        num_experts=num_experts,
+        num_local_experts=num_local_experts,
+        hidden_size=hidden,
+        topk=num_topk,
+        max_tokens_per_rank=num_tokens,
+        mode=ep.MoEMode.LOW_LATENCY,
```
mahdiehghazim: This test cannot construct the communicator against the API in this PR: `MoEMode` exposes `LATENCY`, not `LOW_LATENCY`, and `MoECommunicatorConfig` uses `num_blocks` / `combine_mode` rather than `low_latency_num_blocks` / `low_latency_combine_mode` on the following lines. As written, the new test fails before exercising dispatch. Please update it to the current public configuration names.

src/gpunetio/include/device/doca_gpunetio_dev_verbs_qp.cuh
```diff
@@ -0,0 +1,1061 @@
...
+                                                        sizeof(struct doca_gpunetio_ib_mlx5_wqe_raddr_seg));
+  struct doca_gpunetio_ib_mlx5_wqe_inl_data_seg data_seg;
+  uint32_t bytes = sizeof(T);
+
+  data_seg.byte_count = doca_gpu_dev_verbs_bswap32(bytes | DOCA_GPUNETIO_IB_MLX5_INLINE_SEG);
+  *(uint32_t *)data_seg_ptr = data_seg.byte_count;
+  if (bytes <= sizeof(uint32_t)) {
+    T *dst = (T *)((uintptr_t)data_seg_ptr + sizeof(data_seg));
+    *dst = data;
+  } else {
```
mahdiehghazim, line 732: For any `T` larger than 8 bytes, the header advertises `sizeof(T)` bytes but this branch copies only the first two 32-bit words. The NIC then transmits stale/uninitialized WQE bytes for the remainder. Either copy all `sizeof(T)` bytes supported by the inline segment or add a compile-time `sizeof(T) <= 8` constraint so larger values cannot silently corrupt remote data.

src/gpunetio/include/device/doca_gpunetio_dev_verbs_qp.cuh
```diff
@@ -0,0 +1,1061 @@
...
+ * @param wqe - WQE to be ringed. This buffer must be in shared memory.
+ */
+template <enum doca_gpu_dev_verbs_sync_scope sync_scope = DOCA_GPUNETIO_VERBS_SYNC_SCOPE_GPU>
+__device__ static __forceinline__ void doca_gpu_dev_verbs_ring_bf(struct doca_gpu_dev_verbs_qp *qp,
+                                                                  struct doca_gpu_dev_verbs_wqe *wqe_ptr) {
+  void *bf_ptr = (void *)__ldg((uintptr_t *)&qp->sq_db);
+  uint64_t *wqe = (uint64_t *)wqe_ptr;
+
+  doca_gpu_dev_verbs_fence_release<sync_scope>();
+  asm volatile("cp.async.bulk.global.shared::cta.bulk_group [%0], [%1], 64;" : : "l"(bf_ptr), "l"(wqe));
```
mahdiehghazim: This bulk shared-to-global copy is asynchronous, but the function neither commits the bulk group nor waits until the source shared memory is no longer in use. The caller can return and reuse/overwrite the WQE while the BlueFlame copy is still reading it, so the NIC may receive a torn WQE. Commit the group and perform the required `cp.async.bulk.wait_group.read 0` (or use the existing bulk-store helpers with equivalent completion semantics) before allowing source reuse.

src/gpunetio/src/doca_gpunetio.cpp
seagater: Fixed in [3d796fc](https://github.com/microsoft/mscclpp/commit/3d796fcf3d0c305bccf3850a2312ae65095aefb1). The GPU_CPU fallback now records CPU_GPU as its effective backing type, so destruction uses cudaHostUnregister plus free. Allocation and injected-failure cleanup tests pass.

src/gpunetio/host/gpu_net_io_service.cpp
seagater: Fixed in [3d796fc](https://github.com/microsoft/mscclpp/commit/3d796fcf3d0c305bccf3850a2312ae65095aefb1). Setup and teardown now use CudaDeviceGuard with the stored device ordinal. QP/MR teardown and scratch frees occur before that guard restores the caller's device.

src/gpunetio/src/doca_gpunetio_high_level.cpp
seagater: Fixed in [3d796fc](https://github.com/microsoft/mscclpp/commit/3d796fcf3d0c305bccf3850a2312ae65095aefb1). All three export failures now go through exit_error. Group cleanup destroys the companion before the shared-UAR owner; injected export-failure tests cover these paths.

include/mscclpp/port_channel_device.hpp
seagater: Fixed in [3d796fc](https://github.com/microsoft/mscclpp/commit/3d796fcf3d0c305bccf3850a2312ae65095aefb1). Nonnegative flush budgets now go through tryFlush, with an assertion on timeout or CQ error. Negative budgets retain the blocking path; regression coverage checks both behaviors.

src/gpunetio/src/doca_gpunetio.cpp
seagater: Fixed in [3d796fc](https://github.com/microsoft/mscclpp/commit/3d796fcf3d0c305bccf3850a2312ae65095aefb1). The running flag is now a properly constructed atomic with acquire/release access. Shutdown is also checked inside the progress loop; 100 start/stop cycles pass with a continuously progressing mock QP.

include/mscclpp/internal/port_channel_gpunetio_device_impl.hpp
seagater: Fixed in [3d796fc](https://github.com/microsoft/mscclpp/commit/3d796fcf3d0c305bccf3850a2312ae65095aefb1). Added backend-owned, registered uint64_t result scratch per flat QP, with per-HCA local keys in the device context. Both atomic paths use this scratch instead of payload offset zero; tests verify payload preservation and QP/HCA addressing.

src/gpunetio/include/device/doca_gpunetio_dev_verbs_onesided.cuh
seagater: Fixed in [3d796fc](https://github.com/microsoft/mscclpp/commit/3d796fcf3d0c305bccf3850a2312ae65095aefb1). The unimplemented scalar-put, put-with-signal, and signal warp variants now fail at compile time. Dispatch uses if constexpr so supported thread-scope calls remain valid; negative compilation tests cover all three variants.

python/mscclpp/ep/high_throughput.py
seagater: Fixed in [3d796fc](https://github.com/microsoft/mscclpp/commit/3d796fcf3d0c305bccf3850a2312ae65095aefb1). Removed the stale, unreferenced high_throughput.py module. The communicator continues to use the current throughput.py Context/Runtime implementation.

src/gpunetio/include/host/doca_gpunetio.h
seagater: Fixed in [3d796fc](https://github.com/microsoft/mscclpp/commit/3d796fcf3d0c305bccf3850a2312ae65095aefb1). Added stdbool.h to the common public header chain, along with the missing integer/size includes and opaque forward declarations. Each public host header passes an independent C11 compile with warnings treated as errors.

python/mscclpp/ep/low_latency.py
seagater: Fixed in [3d796fc](https://github.com/microsoft/mscclpp/commit/3d796fcf3d0c305bccf3850a2312ae65095aefb1). Removed the stale, unreferenced low_latency.py module. The current latency.py implementation remains the active runtime using DispatchHandle and private combine contexts.

src/gpunetio/src/doca_verbs_qp.cpp
seagater: Fixed in [3d796fc](https://github.com/microsoft/mscclpp/commit/3d796fcf3d0c305bccf3850a2312ae65095aefb1). QP state now records the internal UAR allocation type, including the BF-to-NC fallback. The accessor uses that state when external_uar is null; tests cover BF, NC fallback, allocation failure, and external UARs.

src/gpunetio/src/doca_verbs_mlx5dv_wrapper.cpp
seagater: Fixed in [3d796fc](https://github.com/microsoft/mscclpp/commit/3d796fcf3d0c305bccf3850a2312ae65095aefb1). Changed the resolved mlx5dv_init_obj function-pointer parameter to uint64_t. A static type check against the installed mlx5dv header passes on the compile host.

src/gpunetio/src/doca_gpunetio.cpp
seagater: Fixed in [3d796fc](https://github.com/microsoft/mscclpp/commit/3d796fcf3d0c305bccf3850a2312ae65095aefb1). Host-backed allocations now use posix_memalign and zero exactly the requested size. The GPU_CPU fallback shares this CPU_GPU path; regression tests check alignment, size, initialization, and cleanup.

src/gpunetio/src/doca_gpunetio_high_level.cpp
seagater: Fixed in [3d796fc](https://github.com/microsoft/mscclpp/commit/3d796fcf3d0c305bccf3850a2312ae65095aefb1). Both constructors now log the output pointer itself rather than dereferencing it. They also clear valid output pointers before allocation; null-output regression cases pass.

test/python/ep/test_low_latency_multirank.py
seagater: Fixed in [3d796fc](https://github.com/microsoft/mscclpp/commit/3d796fcf3d0c305bccf3850a2312ae65095aefb1). Updated the test to LATENCY, num_blocks, and combine_mode. Also migrated its handle access and removed get_expert_output_buffer calls in favor of each dispatch result's combine_input_buffer.

src/gpunetio/include/device/doca_gpunetio_dev_verbs_qp.cuh
seagater: Fixed in [3d796fc](https://github.com/microsoft/mscclpp/commit/3d796fcf3d0c305bccf3850a2312ae65095aefb1). Added a compile-time sizeof(T) <= 8 constraint. Tests accept 1/2/4/8-byte values and require a clear compilation failure for a 16-byte value.

src/gpunetio/include/device/doca_gpunetio_dev_verbs_qp.cuh
seagater: Fixed in [3d796fc](https://github.com/microsoft/mscclpp/commit/3d796fcf3d0c305bccf3850a2312ae65095aefb1). The BlueFlame helper now performs the async shared-memory proxy fence, commits the bulk group, and waits with cp.async.bulk.wait_group.read 0 before returning. The shared address uses the proper conversion/constraint; the actual helper compiles for SM90 and SM100a.

